PR #44, opened 2015-03-17
Interactions system.

shalabhc:
Replace the readline and inputs with a redesigned command_line and
interactions system.


lib/howl/bindings.moon
```diff
@@ -1,8 +1,8 @@
--- Copyright 2012-2014 Nils Nordman <nino at nordman.org>
--- License: MIT (see LICENSE.md)
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 _G = _G
-import table, coroutine, pairs from _G
+import table, coroutine, pairs, io from _G
```
nilnor: Unused require 'io' added here

lib/howl/command.moon
```diff
@@ -1,13 +1,15 @@
...
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
-import inputs, config from howl
+import dispatch, interact, config from howl
 import Matcher from howl.util
+import style, markup, StyledText from howl.ui
 
 append = table.insert
 
+style.define_default 'commandname', 'keyword'
```
nilnor: Snake case should be used for style names.

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+    binding = bindings[name] or ''
+    append items, { name, binding, desc }
+
+  return items
+
+class CommandRunner
+  run: (@finish, opts={}) =>
+    @command_line = howl.app.window.command_line
+    cmd = nil
+    cmdname = ''
```
nilnor: Snake case here as well, `cmd_name`.

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+    @command_line.prompt = ':'
+    @command_line.title = 'Command'
+    if opts.directory
+      @command_line.directory = opts.directory
+    if opts.cmd_string
+      @run_command opts.cmd_string
+
+  run_command: (cmd_string) =>
+    cmd, cmdname, text = parse_cmd cmd_string
```
nilnor: And snake case here as well. Generally speaking, snake case all around :)

kirbyfan64: @nilnor But don't classes in Howl use CamelCase?

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+      @command_line.text = cmd_string
+      log.error 'No such command'
+      return
+
+    @command_line\clear!
+    @command_line.prompt = markup.howl "<prompt>:</><commandname>#{cmdname}</> "
+
+    local results
+
+    ok, error = pcall ->
```
nilnor: This seems like it would clobber the global `error`. I'm surprised and intrigued it hasn't shown up in use.

Also, it can be simplified:

```moonscript
ok, results = pcall -> 
  table.pack '...'
```

nilnor: I see now that the logic is slightly more involved. It doesn't seem right though. If we get an error we log and still try to call finish, and should that actually return `true` we'll try to unpack `nil` which will raise an error.

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+      results = table.pack howl.app.window.command_line\run cmd
+
+    if not ok
+      log.error error
+
+    self.finish results and unpack results
+
+  on_update: (text) =>
+    if text\find ' '
```
nilnor: I interested in this, since it seems like it would mean that we'll never be able to specify arguments with spaces in them? Anyway, I experimented a little bit and encountered an issue. In my case I opened the command prompt, typed 'ctrl' and entered space. This caused the current command in the completion list to be run (`view-close` in my case), which does not seem desirable.

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+    command_line\run_auto_submit command_runner, cmd_string: cmd_string, directory: opts.directory
+  else
+    command_line\run command_runner, cmd_string: cmd_string, directory: opts.directory
+
+howl.interact.register
+  name: 'select_command'
+  description: 'Selection list for all commands'
+  evade_history: true
+  handler: (opts={}) ->
+    with opts
```
nilnor: `opts` should be copied here to avoid mutating the passed in parameter.

lib/howl/command.moon
```diff
@@ -182,48 +83,147 @@ command_completer = ->
...
+  howl.app.window.command_line.command_history
+
+interact.register
+  name: 'select_historical_command'
+  description: 'Selection list for previously run commands'
+  evade_history: true
+  handler: ->
+    line_items = {}
+    for idx, item in ipairs get_command_history!
+      table.insert line_items, {idx, item, command: tostring(item)}
```
nilnor: `table.insert` is aliased as `append` in this file, so that could be used here as well. Also, some subjective style nit-picking; this function seems very compact to me. I personally would prefer empty lines before `result = [..]` and `if result`.

lib/howl/command.moon
nilnor: @kirbyfan64 Absolutely. I belive the Github diff might give the impression that this refers to the class name, but it does not, it refers to `cmdname`. By "snake case all around" I'm referring to variable names and the like.

lib/howl/command.moon
shalabhc: It compiles to `local ok, error = ..` so it seems the global is shadowed, not clobbered. It's still not right and I'll rename the variable.

The last line compiles to `self.finish(results and unpack results)` - which is what I want. I'll add parenthesis to make this clearer.

lib/howl/command.moon
shalabhc: This line itself is invoked only when you are in the CommandRunner interaction - IOW, when the completion list is not open. E.g. if you just type 'open file' without first hitting `tab`, this line will cause the open command to run as soon as you type the space. The arguments after that are handled by the invoked command and may have spaces (depending on what the command allows).

The undesirable behavior you describe is from using submit_on_space when calling `interact.select` below. The selection list interaction has no knowledge of commands, it just matches items in the list and submits when you hit `space`. That behavior can be tweaked, it would require rolling the completion list into the CommandRunner rather than running it as a nested interaction, since we want more control that what `interact.select` provides. 

However consider what behavior you want if you type space with the list showing - should it search the list for two words or should it run a command? E.g. in the current version of howl, with the command completion list showing, if I type 'new buffer', I see that it matches the description of a command. However if I type 'close buffer', it immediately runs the close command rather than show me matching commands.

nilnor: Re the current version, I don't see that behaviour in master. If I type 'close buffer' with the command completion list open, it just closes the completion list and leaves me with the text in the readline as I would expect. Anything other than that I would consider a bug. Generally speaking, typing a space should never submit and run a command, that's what `enter` is for.

As for what should happen when typing a space with the completion list open, I think the best option is probably to continue with the completions. Personally, I wouldn't want a space to change what I've written. That might possibly be a non-default option, but even so it shouldn't submit and run anything.

shalabhc: Ah you are right. I should have said typing 'open ' - but even that doesn't really run the command, rather, just runs the input reader. In terms of the new system, it actually 'runs' the
`open` command, which the interacts with the user.

Changing the completion list to not submit on space is easy. However in the new system, even if you type 'close' followed by space when the completion list is not open, it runs the close command. The CommandRunner doesn't know if a command will or will not accept further input via some interaction, so typing `space` always runs the command. For e.g. in 'open ' case this works fine because the command does accept further input.

I'll think about a better way to not submit argument-less commands on space. An simple way would be for the command definition to specify if it accepts arguments, but then that's one additional thing you have to specify when registering commands.

lib/howl/bindings.moon
shalabhc: Fixed

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/command.moon
shalabhc: Fixed.

lib/howl/commands/app_commands.moon
```diff
@@ -131,25 +132,11 @@ command.register
...
-
-command.register
   name: 'describe-signal',
   description: 'Describes a given signal'
-  input: 'signal'
-  handler: (name) ->
+  interactive: true
+  handler: ->
+    name = interact.select_signal!
     def = signal.all[name]
```
nilnor: Need to check whether a signal was choosen here, otherwise we get the error "Unknown signal 'nil'".

lib/howl/commands/file_commands.moon
```diff
@@ -50,14 +53,19 @@ command.register
 command.register
   name: 'save',
   description: 'Saves the current buffer to file'
+  evade_history: true
   handler: ->
     buffer = app.editor.buffer
-    return command.run 'save-as' unless buffer.file
+    if not buffer.file
+      app.window.command_line\run_after_finish ->
+        command.run 'save-as' unless buffer.file
```
nilnor: The `unless buffer.file` seems redundant given that it's already being checked above, or is it because that could change before the invocation?

lib/howl/commands/file_commands.moon
```diff
@@ -70,14 +78,19 @@ command.alias 'save', 'w'
...
+    file = interact.select_file allow_new: true
+    return unless file
     if file.exists
-      input = inputs.yes_or_no false
-      unless inputs.read input, prompt: "File '#{file}' already exists, overwrite? "
+      unless interact.yes_or_no prompt: "File '#{file}' already exists, overwrite? "
         log.info "Not overwriting; buffer not saved"
         return
 
+    unless file.parent.exists
```
nilnor: This is nice, but perhaps we should ask first? At any rate, to be consistent the same behaviour should be added for the ordinary `save` command.

lib/howl/interact.moon
```diff
@@ -0,0 +1,29 @@
...
+    error 'Missing field for command: "' .. field .. '"' if not spec[field]
+
+  if not (spec.factory or spec.handler) or (spec.factory and spec.handler)
+    error 'One of "factory" or "handler" required'
+
+  interactions[spec.name] = moon.copy spec
+
+unregister = (name) ->
+  interaction = interactions[name]
+  return if not interaction
```
nilnor: The extra check doesn't seem to have a purpose here, we could just assign `nil` directly it seems.

lib/howl/commands/file_commands.moon
shalabhc: This shows the new path with [New] in the list, giving a good indication of what would happen. I'd suggest you try it out and if you still think it's necessary, I'll inject a confirmation question.

Agree about consistency for `save`. That probably should ask for confirmation for new paths.

nilnor: I'm mostly concerned about the consistency vs `save`, it seems like if one of these asks for confirmation than both should do so since the operation performed is the same.

lib/howl/commands/app_commands.moon
```diff
@@ -206,7 +202,8 @@ command.register
     for buffer in *app.buffers
       bundle_name = buffer.file and bundle.from_file(buffer.file) or nil
       if bundle_name
-        command.run "bundle-reload #{bundle_name}"
+        howl.app.window.command_line\run_after_finish ->
+          command.run "bundle-reload #{bundle_name}"
```
nilnor: Need a `submit: true` option here.

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+  parts[#parts]
+
+get_cwd = ->
+  buffer = app.editor and app.editor.buffer
+  directory = buffer and (buffer.file and buffer.file.parent or buffer.directory)
+  directory or File glib.get_current_dir!
+
+looks_like_path = (text) ->
+  return unless text
+  for p in *{
```
nilnor: The dots here are not escaped, matching any character instead of the intended `.`.

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+    @auto_show_list = true
+    local directory
+    if @command_line.directory and @opts.path
+      directory = @command_line.directory / @opts.path
+    elseif @command_line.directory
+      directory = @command_line.directory
+    elseif @opts.path
+      directory = File @opts.path
+    if not (directory and directory.exists and directory.is_directory)
+      directory = get_cwd!
```
nilnor: If a directory or path was actually specified, I think it makes sense to raise an error if the resulting directory does not exist, rather than silently ignoring it? Also styling, the `if not` stanza is not part of the preceding if/else chain so a blank line would improve readability.

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+    @directory = directory
+    trailing = directory.path == File.separator and '' or File.separator
+    @command_line.prompt = markup.howl "\n<operator>[</><directory>#{@directory.short_path}#{trailing}</><operator>] $</> "
+
+  _updir: =>
+    if @directory.parent
+      @_chdir @directory.parent
+
+  _initialize_list_widget: =>
+    @list_widget = ListWidget nil,
```
nilnor: Styling: one line works better I think when there isn't more than one option.

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+
+    path, unmatched = parse_path @directory.path .. File.separator .. last_part
+    @list_widget.matcher = file_matcher self.directory_reader(path), path
+    @list_widget\update unmatched
+    @list_widget_path = path
+    @list_widget_unmatched = unmatched
+
+  _auto_complete_command: (text) =>
+    if not @commands
+      @commands = available_commands!
```
nilnor: These two lines could be:

```moonscript
@commands or= available_commands!
```

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+          path = @directory / dir
+          if path.exists and path.is_directory
+            @_chdir path
+            @list_widget\hide!
+            @command_line.text = ''
+          return
+        @_select_completion!
+        return
+
+      if @list_widget and @list_widget.showing
```
nilnor: This almost duplicates the `if` above it. Perhaps get that out the way first at the start of the method?

```moonscript
unless @list_widget and @list_widget.showing
  @_submit!
  return
```

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+    -- remove prompt so correct history is catpured
+    @command_line.prompt = ''
+    self.finish @directory.path, @command_line.text
+
+  keymap:
+    enter: =>
+      text = @command_line.text
+      cd_cmd = text\umatch '^%s*cd%s+'
+      if cd_cmd and @list_widget and @list_widget.showing
+        if (@list_widget.selection and @list_widget.selection.name == './') or not @list_widget.selection
```
nilnor: Reads better as: 

```moonscript
if not @list_widget.selection or @list_widget.selection.name == './'
```

lib/howl/interactions/external_command.moon
```diff
@@ -0,0 +1,207 @@
...
+
+  _select_completion: =>
+    @list_widget\hide!
+    return unless @list_widget.selection
+
+    filename = @list_widget.selection.name
+    new_path = @list_widget_path / filename
+    @command_line.text = @command_line.text\sub(1, -#@list_widget_unmatched - 1)
+    @command_line\write filename
+    if not new_path.is_directory
```
nilnor: You typically use `if not` in lieu of `unless`, which is fine by me as as I chalk it down to a personal preference, and personally find `unless` to add confusion in some cases. In this case though, `unless new_path.is_directory` read a lot better :)

lib/howl/interactions/file_selection.moon
```diff
@@ -0,0 +1,302 @@
...
+  default: {'a', 'bc', 'git', 'hg', 'o', 'pyc', 'so'}
+
+home_dir = -> File glib.get_home_dir!
+
+root_dir = (file) ->
+  if not file
+    return root_dir home_dir!
+  while file.parent
+    file = file.parent
+  return file
```
nilnor: General style comment: In the rest of the code implicit returns are used, except for early returns.

lib/howl/interactions/file_selection.moon
```diff
@@ -0,0 +1,302 @@
...
+  rel_path = file\relative_to_parent(base_directory)
+  return is_directory and rel_path .. separator or rel_path
+
+parse_path = (path) ->
+  if not path or path.is_blank or not File.is_absolute path
+    local directory
+    if app.editor and app.editor.buffer
+      file = app.editor.buffer.file
+      directory = file.parent if file
+    if not directory
```
nilnor: General style comment 2; blank line before an `if` statement.

lib/howl/interactions/file_selection.moon
```diff
@@ -0,0 +1,302 @@
...
+    selected_path = interact.select
+      title: opts.title or project.root.path .. separator
+      prompt: opts.prompt or ''
+      :matcher
+      columns: { {style: 'filename'} }
+
+    if selected_path
+      return project.root\join selected_path.selection
+
+return {
```
nilnor: I'm not a big fan of this code reuse here. I realize it's convenient since the external command needs similar things, but the things that are really generic should find another home. `home_dir` and `root_dir` might possible be moved into `io.File` with adjustments. `parse_path` and `file_matcher` could be moved into a separate utils file or something similar.

lib/howl/interactions/file_selection.moon
```diff
@@ -0,0 +1,302 @@
...
+
+  unmatched = ''
+  local closest_dir
+  while not path.is_empty
+    pos = path\rfind(separator)
+    if pos
+      unmatched = path\sub(pos + 1) .. unmatched
+      path = path\sub 1, pos
+    closest_dir = File path
+    if closest_dir.exists and closest_dir.is_directory
```
nilnor: The `if .. else` can be removed. A `break if closest_dir.is_directory` line with the other code following suffices. The `if not pos\nbreak` below could be also be more readable in my opinion as a one-liner; `break unless pos`.

lib/howl/interactions/file_selection.moon
```diff
@@ -0,0 +1,302 @@
...
+    append matches, {
+      text,
+      markup.howl '<keyword>[New]</>'
+      name: text
+      is_new: true
+    }
+
+    return matches
+
+separator_count = (s) ->
```
nilnor: I think this could be remove entirely in lieu of `s\count(separator)`?

lib/howl/interactions/grep_search.moon
```diff
@@ -0,0 +1,85 @@
...
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import interact from howl
+import highlight from howl.ui
+import Matcher from howl.util
+
+interact.register
+  name: 'select_match'
+  description: ''
+  handler: (opts) ->
```
nilnor: This function ends up being a little bit too long; it seems like there should be helper functions that could be extracted.

lib/howl/interactions/replacement.moon
```diff
@@ -0,0 +1,385 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import Matcher, TaskRunner from howl.util
+import app, command, config, interact, timer from howl
+import highlight, ListWidget, NotificationWidget from howl.ui
+
+append = table.insert
+
+highlight.define_default 'replace_strikeout', {
```
nilnor: This should be defined in the bundled themes, or at least in the light theme as the default is not that visible there.

lib/howl/interactions/replacement.moon
```diff
@@ -0,0 +1,385 @@
...
+    -- @buffer always holds original text
+    @buffer = howl.Buffer!
+    @buffer.text = @text
+
+    @start_pos = opts.editor.active_chunk.start_pos or 1
+    @end_pos = opts.editor.active_chunk.end_pos or @buffer.length
+
+    @runner = TaskRunner!
+
+    self.find = opts.find or (buffer, target, init) ->
```
nilnor: Good candidate for extracting into it's own helper function.

lib/howl/interactions/replacement.moon
```diff
@@ -0,0 +1,385 @@
...
+    true
+
+  on_update: (text) =>
+    return if not text or text.is_empty
+
+    marker = text[1]
+    target, replacement = parse_replacement text\usub(2), marker
+
+    local loader
+    if target != @target
```
nilnor: I just had to test this and find the case that doesn't work :)

I entered a target and replacement, and then when back and updated the target. This reloads the matches, which removes any replacements, but since the replacement wasn't changed these aren't reloaded.

lib/howl/commands/file_commands.moon
shalabhc: Not sure why I added it as I can't imagine how it would change - removed.

lib/howl/commands/file_commands.moon
shalabhc: Done - both save and save-as now ask for confirmation.

lib/howl/interactions/file_selection.moon
shalabhc: Refactored as suggested. Moved some functions into io.File and others into a new util/paths module.

lib/howl/interactions/grep_search.moon
shalabhc: Just removing the selection change handler makes this significantly shorter.

lib/howl/interactions/replacement.moon
shalabhc: Added. Maybe the light theme should be considered the default and all defaults should be set for that?

lib/howl/interactions/replacement.moon
shalabhc: Ah, right. I knew about this, it just slipped off my radar. Fixed.

lib/howl/interactions/replacement.moon
```diff
@@ -0,0 +1,390 @@
...
+        if @selected_idx == 1
+          @selected_idx = @num_matches
+        else
+          @selected_idx -= 1
+        @_preview_highlights!
+
+      ['editor-scroll-up']: -> app.editor\scroll_up!
+      ['editor-scroll-down']: -> app.editor\scroll_down!
+
+      ['buffer-replace']: => @_switch_to 'buffer-replace'
```
nilnor: As a general note, the ability to switch between the different replacement modes is nice, but we should extend it in the future. As an example, it would be really useful to be able to switch to replacement from any of the search commands, e.g. `buffer-search-forward`, then the key stroke for replacement to enter into replacement mode scoped to the previous matches.

lib/howl/log.moon
```diff
@@ -45,5 +51,6 @@ error = (message) -> last_error = dispatch 'error', message
 clear = ->
   entries = {}
   last_error = nil
+debug = (msg, ...) -> io.stderr\write '\n'..msg, serpent.block {...}
```
nilnor: The output of this isn't very nice, e.g. `log.debug "foo"` generates this:

foo{} --[[table: 0x41dd76d0]]

If we want to to keep it it should be improved.

lib/howl/scintilla.lua
```diff
@@ -1531,7 +1534,14 @@ end
 
 -- Insert string at a position.
 function sci:insert_text(pos, text)
-  self:send(2003, pos, string_ptr(text))
+  ok, result = pcall(function()
```
nilnor: Old debug  code left here.

lib/howl/ui/editor.moon
```diff
@@ -306,11 +306,12 @@ class Editor extends PropertyObject
     @buffer\as_one_undo -> f lines
 
   with_position_restored: (f) =>
-    line, column, indentation = @cursor.line, @cursor.column, @current_line.indentation
+    line, column, indentation, top_line = @cursor.line, @cursor.column, @current_line.indentation, @line_at_top
     status, ret = pcall f, self
     @cursor.line = line
     delta = @current_line.indentation - indentation
     @cursor.column = column + delta
+    @line_at_top = top_line
```
nilnor: This should be set before `@cursor.line` to avoid scrolling twice.

lib/howl/ui/editor.moon
```diff
@@ -465,6 +466,21 @@ class Editor extends PropertyObject
   scroll_up: => @sci\line_scroll_up!
   scroll_down: => @sci\line_scroll_down!
 
+  is_visible: (start_pos, end_pos) =>
```
nilnor: Better named as `range_is_visible` or something similar.

lib/howl/ui/highlight.moon
```diff
@@ -1,9 +1,10 @@
--- Copyright 2012-2013 Nils Nordman <nino at nordman.org>
--- License: MIT (see LICENSE.md)
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 import Scintilla from howl
 import print, tostring, error, pairs from _G
 bit = require 'bit'
+io = require 'io'
```
nilnor: Spurios io `require`.

lib/howl/ui/list_widget.moon
```diff
@@ -0,0 +1,287 @@
...
+  alpha: 50
+  outline_alpha: 100
+}
+
+reversed = (list) -> [item for item in *list[#list, 1, -1]]
+
+class ListWidget extends PropertyObject
+  new: (@matcher, opts={}) =>
+    super!
+    @swidget = TextWidget!
```
nilnor: Should be `@text_widget`.

lib/howl/ui/notification_widget.moon
```diff
@@ -0,0 +1,30 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import TextWidget from howl.ui
+
+class NotificationWidget
+  new: (...) => @swidget = TextWidget ...
```
nilnor: `@text_widget` here as well.

lib/howl/ui/style.moon
```diff
@@ -234,6 +234,19 @@ set_for_theme = (theme) ->
   for name in pairs styles
     define_extended name if is_extended name
 
+ -- set status styles
```
nilnor: What's this? Why is there some special case here for these styles?

lib/howl/interactions/replacement.moon
shalabhc: Absolutely. Somewhat related, I found myself using the replace command to find text as well (if you don't type the separator '/', it can function like a find - you type just the target, use `up`/`down` to jump around and `enter` to re-position cursor). Merging the commands (e.g. `buffer-search-replace`) crossed my mind too but I'm not quite sure it's a good idea.

lib/howl/log.moon
shalabhc: Right. I only use it during debugging so the output doesn't bother me. But I'll remove it for now.

lib/howl/scintilla.lua
shalabhc: Gah! Sorry about that :(

lib/howl/ui/style.moon
shalabhc: This is an attempt to remove the duplicated definition for window.status.(info|warning|error) and editor.styles.(info|warning|error). It just copies the former into editor.status.status-<level>.

Should I remove this?

lib/howl/interactions/replacement.moon
nilnor: I don't think merging the two is a good idea, but being able to switch from a generic search command to the replace mode seems like it would be great. Consider being able to switch from `buffer-search-word-forward` directly to the replacement, or some command that will highlight all occurences in scope. Voila, a simple but effective refactoring tool :)

lib/howl/ui/style.moon
nilnor: Yes, please do, it just adds a special case for code that was previous pretty generic. Besides, the duplication in the themes is not a big issue in my opinion. In the future we can get rid of the GtkLabel instead and just use a normal editing view which can use all the normal styles and Howl markup instead.

lib/howl/interactions/replacement.moon
shalabhc: Hmm. Maybe I misunderstood what you meant by 'scoped' earlier. There are a few issues with this. Right now there is no way to pass a list of matches to buffer-replace. Even if there was, it would have to be out-of-band, i.e. can't be on the command line itself (or via some existing context like the selection). I don't particularly like having out-of-band data. Another issue is if you then modified the target, you might expect it to still match words, but it will reset to a regular match.

I think the right way to do this would be to implement buffer-replace-word and switch to that from from buffer-search-word-forward.

nilnor: (discussion taken to email)

lib/howl/project.moon
```diff
@@ -1,7 +1,7 @@
--- Copyright 2012-2014 Nils Nordman <nino at nordman.org>
--- License: MIT (see LICENSE)
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
-import VC, inputs from howl
+import VC, app, interact from howl
```
nilnor: Unused import `app`.

lib/howl/ui/command_line.moon
```diff
@@ -0,0 +1,503 @@
...
+
+  @property stack_depth: get: => #@running
+
+  @property command_history: get: => moon.copy @_command_history
+
+  run_auto_submit: (...) =>
+    @auto_submit_once = true
+    @run ...
+
+  run: (activity_spec, ...) =>
```
nilnor: This is an overly long method. After perusing it I can't find any obvious candidates for extraction, but perhaps you can?

lib/howl/ui/command_line.moon
```diff
@@ -0,0 +1,503 @@
...
+  enforce_left_pos: =>
+    -- don't allow cursor to go left into prompt
+    return if not @current
+    left_pos = @_prompt_end or 1
+    if @command_widget.cursor.pos < left_pos
+      @command_widget.cursor.pos = left_pos
+
+  handle_keypress: (event) =>
+    -- keymaps checked in order:
+    --   @preemptive_keymap - keys that cannot be remapped by activitys
```
nilnor: Spelling: `activitys`

lib/howl/ui/command_line.moon
```diff
@@ -0,0 +1,503 @@
...
+    widget = @_widgets[name]
+    @box\remove widget\to_gobject! if widget
+    @_widgets[name] = nil
+
+  get_widget: (name) => @_widgets[name]
+
+  show: =>
+    return if @showing
+    @last_focused = @window.focus if not @last_focused
+    @window\_remember_focus!
```
nilnor: `_remember_focus` is a private method, not intended for outside calls (as is any method prefixed with '_').  If it's usable from outside it should be exposed and documented properly.

lib/howl/ui/list_widget.moon
```diff
@@ -0,0 +1,287 @@
...
+style.define_default 'list_highlight', color: '#ffffff', underline: true
+
+highlight.define_default 'list_selection', {
+  style: highlight.ROUNDBOX,
+  color: '#888888'
+  alpha: 50
+  outline_alpha: 100
+}
+
+reversed = (list) -> [item for item in *list[#list, 1, -1]]
```
nilnor: (Interesting use of slicing, I hadn't considered that before)

lib/howl/ui/list_widget.moon
```diff
@@ -0,0 +1,287 @@
...
+class ListWidget extends PropertyObject
+  new: (@matcher, opts={}) =>
+    super!
+    @text_widget = TextWidget!
+
+    @opts = moon.copy opts
+    with @opts
+      .filler_text or= '~'
+
+    @_max_height = 10 * @text_widget.row_height * 10
```
nilnor: The formatting of this and the following line makes it clear that the unnecessary multiplications holds some special significance, but in my befuddled state I can't figure out what it is?

nilnor:
`ui.List` is replace with `ListWidget` if I understand correctly? Neither it nor it's spec is removed though.

lib/howl/ui/list_widget.moon
```diff
@@ -0,0 +1,287 @@
...
+  @property is_multi_column:
+    get: => return #@_columns > 1
+
+  @property has_status:
+    get: => #@_items == 0 or #@_items > @text_widget.height_rows - (@has_header and 1 or 0)
+
+  @property has_items:
+    get: => #@_items > 0
+
+  @property num_columns:
```
nilnor: Naming convention: `nr_*` should be used in lieu of `num_*`.

lib/howl/ui/list_widget.moon
```diff
@@ -0,0 +1,287 @@
...
+
+    @text_widget.buffer\append status, 'comment'
+
+  _select: (idx) =>
+    if not @has_items
+      @selected_idx = nil
+      @_highlight nil
+      return
+
+    if idx < 1
```
nilnor: `min`/`max`, or at least an `if`/`elseif` here as the second `if` will never hold true if the first one does.

lib/howl/ui/notification_widget.moon
```diff
@@ -0,0 +1,30 @@
...
+  info: (text) => @notify 'info', text
+
+  warning: (text) => @notify 'warning', text
+
+  error: (text) => @notify 'error', text
+
+  clear: => @text_widget.text = ''
+
+  notify: (level, text) =>
+    @\clear!
```
nilnor: No need for the `\` here.

lib/howl/ui/list_widget.moon
shalabhc: In Python, `lst[::-1]` was often used to reverse in place, until they added a `reverse()` method.

shalabhc:
ui.List is still used in menu_popup :(

lib/howl/ui/command_line.moon
shalabhc: I only copied what was in readline.moon :p. But I will rename.

lib/howl/ui/editor.moon
shalabhc: Done but had to revert - didn't work as expected - i.e. doing that make the view 'jump' to another a slightly shifted view. Will this actually scroll twice, because the drawing will only happen once after it returns from Lua?

nilnor: Ok, leave it in for now. Yes, the drawing will be done once but display calculations to move the page could happen twice (and will definitely aullar).

lib/howl/ui/command_line.moon
nilnor: Yes, I was going to add a comment to the effect that I realized this but it didn't happen :) Anyway, it would be great if we could change it now at any rate.

bundles/vi/init.moon
```diff
@@ -50,12 +53,12 @@ vi_commands = {
   {
     name: 'vi-toggle',
     description: 'Toggles VI mode'
-    handler: -> if state.active then command.vi_off! else command.vi_on!
+    handler: -> if state.active then command.run('vi-off') else command.run('vi-on')
```
nilnor: These changes could be removed now I believe?

lib/howl/commands/app_commands.moon
```diff
@@ -302,36 +313,54 @@ command.register
...
-    read_stdout: true,
-    read_stderr: true,
-    working_directory: working_directory,
-  }
-
-  buffer = ProcessBuffer p
-  editor = app\add_buffer buffer
-  editor.cursor\eof!
-  buffer\pump!
+  dispatch.launch ->
```
nilnor: Why do we need  to use`dispatch.launch` here now, something to do with the command line flow I guess?

lib/howl/commands/edit_commands.moon
```diff
@@ -1,73 +1,112 @@
...
+      app.editor.searcher\cancel!
 
 command.register
   name: 'buffer-repeat-search',
   description: 'Repeats the last search'
   handler: -> app.editor.searcher\repeat_last!
 
-command.register
-  name: 'buffer-replace',
+howl.command.register
```
nilnor: Unnecessary `howl` prefix slunk in here.

lib/howl/interactions/replacement.moon
nilnor: I think all themes should explicitly define all the basic styles it needs to define, including the light theme, even though it's currently the default.

lib/howl/ui/styled_table.moon
```diff
@@ -0,0 +1,73 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import style, StyledText from howl.ui
+serpent = require 'serpent'
```
nilnor: Unused require.

lib/howl/util/paths.moon
```diff
@@ -0,0 +1,154 @@
...
+
+  for f in *files
+    is_directory = f.is_directory
+    if is_directory continue
+    append paths, display_name(f, is_directory, root)
+
+  sort_paths paths
+  return Matcher paths, reverse: true
+
+return { :file_matcher, :get_cwd, :parse_path, :subtree_matcher }
```
nilnor: It's good that this has been extracted, but should it really be in `howl.util`? These do not seem like generic utility function, so if these for instance are of interest only to interactions then let's move them to `howl.interactions.util` instead. And also, let's not define the config option inside this file, having it in a util file somewhere does not seem good. If no other relevant option pops up, let's move that to `core_variables.moon` instead.

spec/support/spec_helper.moon
```diff
@@ -102,3 +102,21 @@ export pump_mainloop = ->
...
+    @orig_show!
+    on_show!
+
+  ok, error = dispatch.launch activity_function
+  if not ok
+    print error
+
+  command_line.show = command_line.orig_show
+
+export get_list_items = (column=1, widget_name='completion_list') ->
```
nilnor: Too generic a name I think, better to have a descriptive name here, even if it's long. The specs can alias it themselves for better readability if required.

spec/ui/command_line_spec.moon
```diff
@@ -0,0 +1,276 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import app, dispatch, interact from howl
+import Window from howl.ui
+
+describe 'commandline', ->
```
nilnor: `describe 'CommandLine'`

spec/ui/command_line_spec.moon
```diff
@@ -0,0 +1,276 @@
...
+      print result
+    within_activity = nil
+    command_line = nil
+    app.window = nil
+
+  describe 'command_line', ->
+    describe '\\run(activity_spec)', ->
+      it 'errors if handler or factory field not in spec', ->
+        f = -> command_line\run {}
+        assert.has_error f, 'activity_spec requires "name" and one of "handler" or "factory" fields'
```
nilnor: `assert.has_error` only takes one argument. You probably wanted the custom matcher `assert.raises` (not that the error message argument comes first for that one).

spec/ui/command_line_spec.moon
```diff
@@ -0,0 +1,276 @@
...
+        command_line\run
+          name: 'run-activity'
+          handler: ->
+            command_line\run_after_finish nextf
+        assert.spy(nextf).was_called 1
+
+    describe '.text', ->
+      it 'cannot be set when no running activity', ->
+        f = -> command_line.text = 'hello'
+        assert.has_error f, 'Cannot set text - no running activity'
```
nilnor: `assert.raises` here as well, and for all following usages.

spec/ui/command_line_spec.moon
```diff
@@ -0,0 +1,276 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import app, dispatch, interact from howl
+import Window from howl.ui
+
+describe 'commandline', ->
+  local command_line, within_activity
```
nilnor: The `within_activity` shadows the global helper here, and does something else it seems. Rename to something else?

spec/ui/command_line_spec.moon
```diff
@@ -0,0 +1,276 @@
...
+
+      it 'returns the prompt previously set', ->
+        within_activity ->
+          command_line.prompt = 'set'
+          assert.equal 'set', command_line.prompt
+
+    describe 'title', ->
+      it 'is hidden by default', ->
+        within_activity ->
+          assert.equal false, command_line.header\to_gobject!.visible
```
nilnor: For boolean assertions  `assert.is_false` and `assert.is_true` can be used.

spec/ui/list_widget_spec.moon
```diff
@@ -0,0 +1,344 @@
...
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import Buffer from howl
+import ListWidget, style, highlight from howl.ui
+import Matcher from howl.util
+s = require 'serpent'
+
+describe 'ListWidget', ->
+  local list, buf
+  throwaway = ListWidget -> { 'H' }
```
nilnor: Snake case for variables even in specs please :)

nilnor:
Good news - I've now completed the full pass over the diff :)

So there are some comments from the review left to discuss/address, but once that's done I think the code is ready. As we said before we'll do style adjustments in master later, so apart from the remaining PR comments there's "only" the documentation that needs updating before merging.

spec/ui/command_line_spec.moon
nilnor: I see that the standard error assert takes two arguments now, but I think that's in upcoming 2 release of busted?

shalabhc:
> Good news - I've now completed the full pass over the diff :)

Excellent. I'll get to the pending items in the next few days. None of them seem particularly major.

spec/ui/command_line_spec.moon
shalabhc: Changing the second parameter to a string that's not a substring of the error message causes the test to fail. So I'm guessing it uses the second parameter. Not sure how I discovered that, but I'll change it to `raises` anyway.

nilnor: I'm guessing it's been added to the luassert rock, and newer installs of the stable busted version picks it up now as well. You can keep it if you prefer it since it works.

lib/howl/commands/app_commands.moon
shalabhc: Right. The handler is called in the same coroutine that called  `command.run` - so running a long running process doesn't work well because the command stays active and the command line doesn't close. (Maybe you remember the bug you reported..).

nilnor: Perhaps all handlers should be invoked using dispatch instead? It's slightly more inefficient to use coroutines for all handlers, on the other hand it means that handlers needn't worry about it. So there's long running commands now, but in the future it will also include any handler that fetches something from the network, etc. Or would that cause other problems?

lib/howl/util/paths.moon
shalabhc: Matcher is in util, so I thought path based matchers are similar in some respect. But I don't mind moving this except `get_cwd` is now used  by `app_commands`, so at least one function is used outside interactions. Since it depends on `howl.editor`, moving it to `io.File` doesn't seem right either. Open to other suggestions.

nilnor: All right, so perhaps this is as good a place as any for it. I'm thinking my biggest gripe about it right now is probably the naming of `parse_path`, which could possibly be better named to reflect what it returns. E.g. `get_dir_and_path`, `separate_path`, `directory_and_path`.

spec/ui/command_line_spec.moon
shalabhc: I had to shorten the expected error string in a couple of cases once I renamed this - not sure why the original string didn't work.

lib/howl/commands/app_commands.moon
shalabhc: The behavior (as designed) is that the handler runs in the 'foreground'. In this case we want the command to exit but to run something else in the background. I think this should be explicit with dispatch.launch.

To elaborate - there's the concept of a running command being 'active' for the duration it is running and you can have a stack of running commands (or interactions). When exiting, the stack unwinds until it's clear and the command line gets hidden. In this case we want the command to exit and the command line to be hidden, but another task to run. Hope that made some sense but I can elaborate further if needed.

nilnor: I believe I see the reason for doing it like this, with the stack and all and the need for a return to properly deactivate things. I'm not sure it sounds it's a good idea though to require all command handlers to be aware of this. For one, I imagine this is really the most common mode for most commands; they might perhaps read some input but after that they're going to do something else, which might or might not end up yielding in some operation. Whether a certain operation will yield or not is not something that will always be readily apparent, and might also change after the command has been written if the implementation of something else changes. The nice thing about having a coroutine-based API, such as for commands now, and for other things in the future, is that you can just write your code and not worry to much about whether this will block or not, etc. This seems a step backwards in that regard. To be on the safe side commands would then always have to explicitly launch stuff in an co-routine. That stuff is not trivial to get right though. Consider for instance the example above: `dispatch.launch` is invoked, but nothing checks the return value, so any errors before the actual yield later is silently ignored.

It would be better if this could be supported automatically. What would the issues be if all handlers were dispatched in their own coroutines? It would be possible to know after a dispatch whether the handler is done or has yielded, so that information could then be used for determining how to handle the command line logic.

shalabhc: Just running each handler in it's own coroutine might be straightforward but I don't think it will solve the issue. The issue stems from having the semantics that the running command 'owns' the command line up until it exits. Only after that are other commands free to take over the command line.

Since the command line state is shared and determined by the stack of running command+interactions, I'm not sure how it would play out if multiple handlers could be active at the same time, each allowed to interact via the command line at will. Consider one command might invoke a yes_or_no question, while another might invoke invoke a read_text in it's own coroutine. Because interactions are on a stack, this would also look weird (':command1_promptcommand2_promptyes_or_no_prompt?read_text_prompt:'). I understand you are not talking about this case, but probably mean that a command that is not going to interact with the command line anymore should not block other commands from running. The issue then is determining that a command is 'not going to interact with the command line anymore'. There is no automatic way to determine this. Yielding is not a good indicator for this because commands may yield while intending to keep ownership of the command line - consider something that reads the file system for a large number of files, or the current replace command that yields to stay responsive while replacing in large buffers.

So the only way I can think of at the moment is for the handler to explicitly declare that it is done with the interactive command line. Other than dispatch.launch (which may ignore errors, as you point out) you can currently use `command_line\run_after_finish` - which could be made to handler errors properly, if it doesn't already. Alternatively, maybe we can have the handler itself return a function to run? E.g.

```moonscript

 handler:
    ...
    return -> launch_cmd working_directory, cmd
```
 
Currently the return value is returned to the caller of `command.run` (but I don't think it's really used anyway) - so we would remove that feature.

shalabhc: Another way (coming full circle further) might be to provide two separate functions in the command definition - the interactive and non-interactive parts. This would subsume the `interactive` boolean. The non interactive part would run in its own coroutine. This is not that different from my earlier suggesting of returning a function from the handler (because you split your code into two functions either way), but it is better in other ways because two distinct keys in the command definition make the meaning clearer I think.

To develop this idea further:

```moonscript
command.register:
   name: 'command'
   interactor: -- (optional) either a function or a factory table, blocks the command line while running
   handler: -> -- function that runs in a separate coroutine
```

The output of interactor would be fed into handler (details tbd).

nilnor: So, I think I like the last proposed solution the best. Since we've moved back to allowing direct dispatch anyway, we currently need to check whether arguments are provided or not anyway in the handlers. This way, the `handler` can always assume arguments are provided properly, so we would be moving that conditional stuff clearly into a separate get-the-input function. 

I'm not a big fan of any solution that would require all commands to invoke some special dispatch function - the above seems much better to me.

shalabhc: Ok, then. There are still some implementation details that need to be worked out. For e.g. the command line currently runs both interactions and commands. (Perhaps it is really designed for interactions - i.e. it blocks - which is why it requires commands to use dispatch.launch). 

With commands specifying both - the `handler` and an optional `interactor` - maybe it's best if the command module itself runs the command handlers, and only invokes the command_line to run the interactor if provided. Also, the `command_line` only runs interactions (and not 'activities' as it does currently - which is used to mean both commands and interactions). This removes one extra term from the terminology (activity) and decouples the commands from the tightly coupled interactions+command_line. I feel this will be a really good change, assuming it's doable without running into other issues.

nilnor:
All doc updates so far looks good (I added some minor comments/suggestions).

spec/ui/list_widget_spec.moon
shalabhc: Not 100% which variable you meant but assuming 'throwaway'? Renamed.

shalabhc:
Addressed all doc suggestions and added the pending docs.

PR #45, opened 2015-03-19
Lexer fixes

kirbyfan64:
- Reorder stuff in the C/C++ lexer to correctly lex things like `static_assert`.
- Remove Python decorator support. It disabled lexing for the rest of the line, which was ugly, and it was kind of broken.


nilnor:
Thanks!

PR #48, opened 2015-03-25
Python decorator styling

shalabhc:
@kirbyfan64 - Do you mind trying this out? This fixes the issue with the entire decorator line being unstyled. Instead, the '@' and just the following word is styled using the 'preproc' style.

### DIFF
--- a/bundles/basic_modes/lexers/python.lua
+++ b/bundles/basic_modes/lexers/python.lua
@@ -103,6 +103,9 @@ local identifier = token(l.IDENTIFIER, l.word)
 -- Operators.
 local operator = token(l.OPERATOR, S('!%^&*()[]{}-=+/|:;.,?<>~`'))
 
+local decorator = token('decorator',
+                        #P('@') * l.starts_line('@' * l.word^0))
+
 M._rules = {
   {'whitespace', ws},
   {'keyword', keyword},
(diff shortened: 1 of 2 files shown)


nilnor:
Slightly related to this, it would be nice to create a "real" Python mode with a new lexer and everything. I could take a look at that later, but what would really help in that case is if someone could provide a good example file containing all kinds of Python syntax (such as the [example file](https://github.com/nilnor/howl/blob/master/bundles/ruby/misc/example.rb) for Ruby).

kirbyfan64:
@nilnor You might want to try out the [pygments examplefiles](https://bitbucket.org/birkenfeld/pygments-main/src/tip/tests/examplefiles/). Only issue is that they have every language except Python (but that's probably more because the Python docs use pygments).

kirbyfan64:
So...this looks like it *kind of* works. Only other issue is that decorators don't need to be at the beginning of a line. Consider:

```
def outer():
    @abc
    def inner(): pass
```

When they aren't at the beginning, they aren't colored.

nilnor:
@kirbyfan64 Thanks for the tip. That seems to be more of collection of different files each containing some subset of the syntax though, still requiring manual work to compile into one example file, and not neccesarily complete either I'm guessing.

shalabhc:
A proper Python mode is a better, but more time consuming, alternative.
For now, I've updated this PR to better handle indented decorators as well as dotted names.

kirbyfan64:
This looks like it works perfectly! What are the chances of getting it merged in?

nilnor:
Very high now that you've given feedback :)

+1 from me, merge at will if you're satisfied @shalabhc .

PR #49, opened 2015-04-07
Fix character lexing in the C lexer

kirbyfan64:
This fixes lexing of the following:

```c
'\''
'\n'
'\a'
'\b'
... // more \x characters
'\"' // right now, this will treat the rest of the file as a string; this PR fixes that, too
```


nilnor:
Good catch!

Two things that I'd like to see addressed before merging:

- The lexer rule could be written using a set instead of `any` I think, e.g. S'abfn..', listing all the possible one char combinations
- The `misc/example.c` file should be updated to add examples for these cases to make it more complete

kirbyfan64:
@nilnor Done!

nilnor:
@kirbyfan64 excellent, thanks!

PR #54, opened 2015-06-08
Implement window\save_screenshot

shalabhc:


nilnor:
A general note first, I'll add some details inline after that:

The `save_screenshot` should not be implemented in `Gtk.Window`. ljglibs should only map the existing API:s (even though it also takes some liberties with introducing "missing" properties for getters). So no `save_window` in the Gtk API means no method here either. The implementation should be completely in `howl.Window` I think.

lib/ljglibs/gtk/window.moon
```diff
@@ -83,4 +83,14 @@ core.define 'GtkWindow < GtkBin', {
   set_default_icon_from_file: (filename) ->
     catch_error(C.gtk_window_set_default_icon_from_file, filename) != 0
 
+  save_screenshot: (filename, format='png') =>
+    gdkwindow = ffi.C.gtk_widget_get_window ffi.cast 'GtkWidget*', @
```
nilnor: The Gdk window is available via the `.window` property.

lib/ljglibs/gtk/window.moon
```diff
@@ -83,4 +83,14 @@ core.define 'GtkWindow < GtkBin', {
...
+    shot = ffi.C.gdk_pixbuf_get_from_window gdkwindow, 0, 0, @allocated_width, @allocated_height
```
nilnor: Instead of going directly through the FFI we should create a proper wrapper for pixbufs under `ljglibs.Gdk.Pixbuf`. That would then have the `get_from_window` function. This leaks a reference (missing `g_object_unref` handling), and it does not check its return for `NULL`.

lib/ljglibs/gtk/window.moon
```diff
@@ -83,4 +83,14 @@ core.define 'GtkWindow < GtkBin', {
...
+    err = ffi.new 'GError *[1]'
+    ffi.C.gdk_pixbuf_save shot, filename, format, err, nil
```
nilnor: In accordance with the above, this would the `pixbuf\save(shot, filename, format)`. Also note that the error handling is missing a `g_error_free`.

shalabhc:
Ah I see - your comments make a lot of sense. I've updated the patch to address all of them - PTAL.

lib/ljglibs/gdk/pixbuf.moon
```diff
@@ -0,0 +1,20 @@
...
+
+core.define 'GdkPixbuf', {
+  get_from_window: (window, x, y, width, height) ->
+    pixbuf = C.gdk_pixbuf_get_from_window(window, x, y, width, height)
+    error 'Failed to get pixbuf' unless pixbuf
+    ffi.gc(pixbuf, C.g_object_unref)
+    pixbuf
+
+  save: (filename, type) =>
+    glib.catch_error(C.gdk_pixbuf_savev, @, filename, type, nil, nil)
```
nilnor: The documentation for savev states that the last two arguments here are NULL-terminated arrays. It does not explicitly allow for `NULL`, so there seem to be no guarantee that this would always work.

lib/ljglibs/gdk/pixbuf.moon
```diff
@@ -0,0 +1,20 @@
...
+C = ffi.C
+
+core.define 'GdkPixbuf', {
+  get_from_window: (window, x, y, width, height) ->
+    pixbuf = C.gdk_pixbuf_get_from_window(window, x, y, width, height)
+    error 'Failed to get pixbuf' unless pixbuf
+    ffi.gc(pixbuf, C.g_object_unref)
+    pixbuf
+
+  save: (filename, type) =>
```
nilnor: In this case I think it perhaps doesn't make sense to adhere slavishly to the API with regards to the optional parameters for `save` and `savev`, but it would be nice to make this complete in the sense that it would accept as its last parameter a Lua table with optional parameters that we'd pass into `savev`. There's a wrapper helper in the glib module, `char_p_arr`, that takes a table and produces a NULL terminated array suitable for this. So, given a last parameter `opts = {}`, something like this:

```moonscript
option_keys = glib.char_p_arr [k for k in pairs opts]
option_values = glib.char_p_arr [v for k, v in pairs opts]
```

And pass those to `savev`.

shalabhc: Done!

Btw, I'm not sure but it seems that `char_p_arr` does not anchor the result of `strdup` so those objects might get GCed?

nilnor: Right you are, good catch! I'll look at this.

PR #57, opened 2015-06-27
Various lexer fixes/improvements

kirbyfan64:
- `CMakeLists.txt` should use the CMake lexer.
- Same thing for `*.gmk` and `makefile`: make lexer.
- I added some common Python file names from build systems (Waf, SCons, and Bazel), as well as the Cython extension `.pyx`.
- In Ruby, `case` should add an `end`.
- Also in Ruby, `class` can be followed by [whatever the heck you want](http://stackoverflow.com/a/474888/2097780).
- [Multi-character constants are valid but undefined in C and C++](http://stackoverflow.com/questions/7755202/multi-character-constant-warnings); [`boost::mpl::string` uses them](http://www.boost.org/doc/libs/1_58_0/libs/mpl/doc/refmanual/string.html).

The other extension changes aren't really necessary. [Hy](http://docs.hylang.org) uses the Clojure lexer, and [Crystal](http://crystal-lang.org) uses the Ruby lexer. I can take them out if wanted, but it seemed useful just for completeness.

### DIFF
--- a/bundles/basic_modes/mode_definitions.moon
+++ b/bundles/basic_modes/mode_definitions.moon
@@ -294,7 +294,7 @@ common_auto_pairs = {
 
   python:
     extensions: { 'sc', 'py', 'pyw', 'pyx' }
-    patterns: { 'wscript$', 'SConstruct', 'SConscript', 'BUILD' }
+    patterns: { 'wscript$', 'SConstruct$', 'SConscript$' }
     shebangs: '[/ ]python.*$'
     comment_syntax: '#'
     indentation: {
--- a/bundles/basic_modes/mode_definitions.moon
+++ b/bundles/basic_modes/mode_definitions.moon
@@ -61,7 +61,7 @@ common_auto_pairs = {
 
   cmake:
     extensions: { 'cmake', 'ctest' }
-    patterns: { '.cmake.in$', '.ctest.in$' }
+    patterns: { '.cmake.in$', '.ctest.in$', 'CMakeLists.txt$' }
     comment_syntax: '#'
     auto_pairs: {
       '(': ')'
@@ -229,9 +229,9 @@ common_auto_pairs = {
     }
 
   makefile:
-    extensions: { 'iface', 'mak' }
+    extensions: { 'iface', 'mak', 'mk', 'gmk' }
     aliases: 'make'
-    patterns: { 'GNUmakefile$', 'Makefile$' }
+    patterns: { 'GNUmakefile$', 'Makefile$', 'makefile$' }
     comment_syntax: '#'
     auto_pairs: common_auto_pairs
 
@@ -293,7 +293,8 @@ common_auto_pairs = {
     }
 
   python:
-    extensions: { 'sc', 'py', 'pyw' }
+    extensions: { 'sc', 'py', 'pyw', 'pyx' }
+    patterns: { 'wscript$', 'SConstruct', 'SConscript', 'BUILD' }
     shebangs: '[/ ]python.*$'
     comment_syntax: '#'
     indentation: {
--- a/bundles/c/c_lexer.moon
+++ b/bundles/c/c_lexer.moon
@@ -55,6 +55,7 @@ howl.aux.lpeg_lexer ->
 
   include_stmt = sequence {
     c('preproc', '#include'),
+    ws^0,
     c('operator', '<'),
     c('string', complement('>')^1),
     c('operator', '>'),
--- a/bundles/c/c_lexer.moon
+++ b/bundles/c/c_lexer.moon
@@ -33,14 +33,7 @@ howl.aux.lpeg_lexer ->
     span '/*', '*/'
   }
 
-  char_constant = P"'" * any({
-    "\\" * any {
-      R('07') * R('07') * R('07'),
-      'x' * xdigit * xdigit,
-      S'abfnrtv\\\'"?'
-    },
-    1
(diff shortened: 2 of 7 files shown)


bundles/c/c_lexer.moon
```diff
@@ -62,7 +55,6 @@ howl.aux.lpeg_lexer ->
 
   include_stmt = sequence {
     c('preproc', '#include'),
-    ws^1,
```
nilnor: Did this slink in by mistake, or were you trying to fix something?

bundles/c/c_lexer.moon
```diff
@@ -33,14 +33,7 @@ howl.aux.lpeg_lexer ->
...
 
-  char_constant = P"'" * any({
-    "\\" * any {
-      R('07') * R('07') * R('07'),
-      'x' * xdigit * xdigit,
-      S'abfnrtv\\\'"?'
-    },
-    1
-  }) * "'"
+  char_constant = span("'", "'", '\\')
```
nilnor: Could you add an example of the multi-character literal syntax to `bundles/c/misc/example.c`?

bundles/c/c_lexer.moon
kirbyfan64: @nilnor It's intentional. It's technically valid and makes reading APL-style C code a little easier. :)

nilnor: I see. The problem is now only include statements without white space are handled correctly. `ws^0` is probably what you want in this case. Similar to my other comment, could you add an additional example to (and verify against) the example.c file?

bundles/basic_modes/mode_definitions.moon
```diff
@@ -293,7 +293,8 @@ common_auto_pairs = {
     }
 
   python:
-    extensions: { 'sc', 'py', 'pyw' }
+    extensions: { 'sc', 'py', 'pyw', 'pyx' }
+    patterns: { 'wscript$', 'SConstruct', 'SConscript', 'BUILD' }
```
nilnor: These are too generic. First off, apart from the first one they're not anchored, meaning these will match files such as `RUBYBUILD`, `FILESConstruct`, etc. Even if they were anchored (using `^`), `BUILD` is so generic I wouldn't any mode to claim that.

bundles/clojure/init.moon
```diff
@@ -9,7 +9,7 @@ parser = bundle_load 'clojure_parser'
 register_mode = ->
   mode_reg =
     name: 'clojure'
-    extensions: {'clj', 'cljs', 'edn'}
+    extensions: {'clj', 'cljs', 'edn', 'hy'}
```
nilnor: This can be added to the lisp mode instead, which has generic Lisp support.

bundles/ruby/init.moon
```diff
@@ -9,7 +9,8 @@ mode_reg =
     'rb',
     'rbw',
     'gemspec',
-    'builder'
+    'builder',
+    'cr'
```
nilnor: I'm going to have to say no to this one. While it seems to have a very similar syntax to Ruby, it's not Ruby (and while the syntax is similar, it's not the same). For now, unless you want to write a proper mode for it, I suggest something like this in the init file:

```moonscript
howl.mode.register {
  name: 'crystal',
  extensions: 'cr',
  parent: 'ruby',
  create: -> {}
}
``` 

This will create an empty Crystal mode that is derived from the Ruby mode, meaning lexing, etc., will work the same way.

bundles/ruby/ruby_mode.moon
```diff
@@ -69,8 +69,8 @@ continuation_indent = (line, indent_level) ->
     multiline: {
       { r'\\s+do(?:\\s*\\|[^|]+\\|)?\\s*$', '^%s*end', 'end' },
       { r'^\\s*def\\s+\\w[\\w.\\d]+(?:\\s*\\([^)]*\\))?\\s*$', '^%s*end', 'end' },
-      { r'^\\s*(class|module)\\s+\\p{Lu}[\\w\\d]*\\s*$', '^%s*end', 'end' },
-      { r'^\\s*(if|unless)\\s+', '^%s*end', 'end' },
+      { r'^\\s*(class|module)\\s+\\p{Lu}.*$', '^%s*end', 'end' },
```
nilnor: The problem with this is that it will now insert an additional end for one-line declarations, e.g.:

```ruby
module PlaceHolder; end
```

nilnor:
All right, all looked over :)

Needs some changes and some removals. For the ruby completion stuff I think some other changes would be needed in the formatting code to support the one-line declarations properly (e.g. it could detect that the start line also contains the end pattern already), but until that happens I think it's better to leave out the `module`/`class` completion fixes.

bundles/basic_modes/mode_definitions.moon
kirbyfan64: @nilnor 

1. I forgot the anchors...
2. I'll remove `BUILD`, though I don't know of any build systems that use it except for Bazel (uses a subset of Python 2) and Pants (uses Python, 3 I believe).

bundles/ruby/init.moon
kirbyfan64: @nilnor I'll just remove it and upload it separately somewhere else.

kirbyfan64:
@nilnor I think I got everything!

nilnor:
All right, looks good!

PR #58, opened 2015-06-27
Implement select_line|location interactions, replace select_match

shalabhc:
Also add previews for project-open and switch-buffer.

'open' does not include previews yet because it is different enough and doesn't use interact.select. Previews for 'open' can be added as a separate feature later.


lib/howl/interactions/line_selection.moon
```diff
@@ -0,0 +1,97 @@
...
+          if positions
+            for hl_pos in *positions
+              highlight.apply 'search_secondary', buffer, start_pos + hl_pos - 1, 1
+
+interact.register
+  name: 'select_line'
+  description: 'Selection for buffer lines'
+  handler: (opts) ->
+    lines = opts.lines
+    opts.lines = nil
```
nilnor: Why is this done?

nilnor:
It would be great if we could add these new interaction to the documentation as well ("Built-in interactions").

lib/howl/interactions/location_selection.moon
```diff
@@ -0,0 +1,46 @@
...
+        buffer = selection.buffer or get_buffer_for_file selection.file, local_buffers
+        editor.buffer = buffer
+        if selection.line_nr
+          editor.line_at_center = selection.line_nr
+
+      if on_selection_change
+        on_selection_change selection, text, items
+
+    result = interact.select opts
+    return result if result
```
nilnor: This will leave the temporary preview buffer in the editor? That doesn't seem like it's a good idea. I haven't seen this be a problem currently when I'm testing, but I think that's because the commands now using the interaction will replace it immediately anyway. But consider a command that is just going to use the result for something else. We'll end up with a preview buffer that will have to be manually closed.

lib/howl/interactions/location_selection.moon
```diff
@@ -0,0 +1,46 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import app, interact, mode, Buffer from howl
+import highlight from howl.ui
+import Matcher from howl.util
+
+get_buffer_for_file = (file, local_buffers) ->
```
nilnor: Just a note: We could possibly have `local_buffers` be a table keyed by file path. That would cut down on the iteration required for revisited files.

lib/howl/interactions/line_selection.moon
```diff
@@ -0,0 +1,97 @@
...
+    opts.on_selection_change = line_match_highlighter(editor)
+
+    result = interact.select_location opts
+
+    if result
+      line = result.selection.line
+      return {
+        :line
+        text: result.text
+        positions: matcher.explain result.text, line.text
```
nilnor: I'm thinking we could skip returning `positions` here, and just return `column` instead. From the code using this it's unclear how positions would be useful, since they're typically related to the interactive search. Once it's done it's done. The only possible case I could think of would be to keep the highlights, but that doesn't seem useful anyway. What do you say?

nilnor:
All done!

Very nice :) I was interested in seeing how the performance for this would be, but I'm pleased to see it's very snappy when I test it.

lib/howl/interactions/line_selection.moon
shalabhc: There's no harm in leaving this in but I didn't want to pass down unused, unnecessary options to select_location. I also didn't want anyone to use opts.lines within select_location, just because it happens to be available.

lib/howl/interactions/location_selection.moon
shalabhc: Ah, good point. What do you think is better - registering the selected buffer or reverting to the original buffer before returning? I'm leaning towards the former because In almost all cases the selected buffer will end up being opened anyway.

shalabhc:
Added the docs as well!

lib/howl/interactions/line_selection.moon
nilnor: I noticed later that you do a copy of `opts` further down. The clearing should be done after the copying to avoid clobbering the input argument.

lib/howl/interactions/location_selection.moon
shalabhc: Done!

lib/howl/interactions/location_selection.moon
shalabhc: On further thought, I decided to go with the latter option. I think interactions should not affect editor state as far as possible - that's the job of the handler.

lib/howl/interactions/line_selection.moon
shalabhc: Fixed!

lib/howl/interactions/location_selection.moon
nilnor: I agree.

lib/howl/interactions/line_selection.moon
shalabhc: Done! They highlights are still kept because the buffer stays the same.

nilnor:
Well, I think that's a wrap :)

PR #60, opened 2015-07-01
Fix #59

### LINKED ISSUE #59: Auto pairs don't always work
Sometimes, entering the first character in an auto pair correctly adds the other one afterwards. Other times, it simply doesn't work.

I know that's not much info, but I don't know what other info would be wanted, so I decided to post this first.

kirbyfan64:
YESSS!!!!!!!!! ;)

### DIFF
--- a/lib/howl/bindings.moon
+++ b/lib/howl/bindings.moon
@@ -194,9 +194,6 @@ export cancel_capture = ->
   capture_handler = nil
   is_capturing = false
 
-export is_capturing = ->
-  capture_handler != nil
-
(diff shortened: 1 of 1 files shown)


nilnor:
Good find! However, as you might have noticed the build fails now. That's because the API is actually that `is_capturing` should be a boolean property and not a function. The fix seems like it should be just removing the erroneous `is_capturing` function instead. Could you try that instead, and presuming that it works update the PR?

kirbyfan64:
@nilnor Fixed and rebased!

nilnor:
Excellent, thanks!

PR #61, opened 2015-07-08
Various file preview fixes

shalabhc:
- Don't reset last_shown on buffers during preview
- Add config.preview_files
- Don't preview binary files

Also, now the *-buffer-switch events are not fired during previewing.

### DIFF
--- a/lib/howl/variables/core_variables.moon
+++ b/lib/howl/variables/core_variables.moon
@@ -28,7 +28,7 @@ config.define
 
 config.define
   name: 'preview_files'
-  description: 'Whether to automatically preview files when switching between them'
+  description: 'Whether to automatically preview the selected file or buffer'
   default: true
   type_of: 'boolean'
   scope: 'global'
--- a/lib/howl/variables/core_variables.moon
+++ b/lib/howl/variables/core_variables.moon
@@ -25,3 +25,10 @@ config.define
   description: 'Whether to automatically format code when possible'
   default: true
   type_of: 'boolean'
+
+config.define
+  name: 'preview_files'
+  description: 'Whether to automatically preview files when switching between them'
+  default: true
+  type_of: 'boolean'
+  scope: 'global'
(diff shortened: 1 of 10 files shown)


lib/howl/variables/core_variables.moon
```diff
@@ -25,3 +25,10 @@ config.define
   description: 'Whether to automatically format code when possible'
   default: true
   type_of: 'boolean'
+
+config.define
+  name: 'preview_files'
+  description: 'Whether to automatically preview files when switching between them'
```
nilnor: The description should be a little more generic, it's not necessary that a switch will be the action taken after the interaction.

spec/ui/editor_spec.moon
```diff
@@ -281,6 +281,25 @@ describe 'Editor', ->
...
+      time = os.time
+      now = time!
+      os.time = -> now
+      pcall ->
+        editor.buffer = Buffer {}
+      os.time = time
+      assert.same now, buffer.last_shown
+
+  context 'previewing', ->
+    it 'does not update last_shown for previewed or switched out buffer', ->
```
nilnor: Why would we not want to update `last_shown` for an ordinary buffer that is switched out as the result of a preview being shown?

nilnor:
All right, two minor comments, and otherwise looks good!

spec/ui/editor_spec.moon
shalabhc: The idea was just to not affect existing buffer state at all while previewing.

lib/howl/variables/core_variables.moon
shalabhc: Updated.

shalabhc:
Addressed your comments - PTAL when convenient.

spec/ui/editor_spec.moon
nilnor: Yes, that makes sense. This isn't currently a problem with the interaction, since that always restores the original buffer, but I'm thinking that there is a general case where a preview is shown in an editor and the original buffer isn't restored. In that case we fail to properly update `last_shown` for the switched out buffer.

shalabhc: Valid point. Fixed and pushed - PTAL when you get a chance.

nilnor:
Looks good!

PR #62, opened 2015-08-05
More Python and Ruby lexer fixes

kirbyfan64:
1. The Python lexer currently won't indent after `else:`.
2. Ruby function names can have the `!` and `?` characters.

### DIFF
--- a/bundles/basic_modes/mode_definitions.moon
+++ b/bundles/basic_modes/mode_definitions.moon
@@ -299,9 +299,9 @@ common_auto_pairs = {
     comment_syntax: '#'
     indentation: {
       more_after: {
-        { ':%s*$', 'else:%s*$' }
+        { ':%s*$' }
       }
-      less_for: { 'else:%s*$'}
+      less_for: { 'else%s*:%s*$' }
     }
     auto_pairs: common_auto_pairs
     structure: (editor) =>
--- a/bundles/ruby/ruby_mode.moon
+++ b/bundles/ruby/ruby_mode.moon
@@ -68,7 +68,7 @@ continuation_indent = (line, indent_level) ->
   code_blocks:
     multiline: {
       { r'\\s+do(?:\\s*\\|[^|]+\\|)?\\s*$', '^%s*end', 'end' },
-      { r'^\\s*def\\s+\\w[\\w.\\d]+(?:\\s*\\([^)]*\\))?\\s*$', '^%s*end', 'end' },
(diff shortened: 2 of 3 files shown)


bundles/basic_modes/mode_definitions.moon
```diff
@@ -299,9 +299,8 @@ common_auto_pairs = {
     comment_syntax: '#'
     indentation: {
       more_after: {
-        { ':%s*$', 'else:%s*$' }
+        { ':%s*$' }
       }
-      less_for: { 'else:%s*$'}
```
nilnor: Why remove the `less_for`?

kirbyfan64: @nilnor Does it make a difference? For me, it doesn't...

nilnor: `less_for` controls what statements should be indented less  compared to the previous line. This change removes this for `else:`.

bundles/ruby/ruby_mode.moon
```diff
@@ -68,7 +68,7 @@ continuation_indent = (line, indent_level) ->
   code_blocks:
     multiline: {
       { r'\\s+do(?:\\s*\\|[^|]+\\|)?\\s*$', '^%s*end', 'end' },
-      { r'^\\s*def\\s+\\w[\\w.\\d]+(?:\\s*\\([^)]*\\))?\\s*$', '^%s*end', 'end' },
+      { r'^\\s*def\\s+\\w[\\w.\\d?!]+(?:\\s*\\([^)]*\\))?\\s*$', '^%s*end', 'end' },
```
nilnor: The ruby bundle has a spec for the indentation support. This should be updated to cover these cases (you can run the specs with `./bin/howl-spec bundles/ruby/spec/`).

kirbyfan64: @nilnor I added specs. However, I can't test them right now, since my Lua installation is seriously messed up, so... \*crosses fingers*

bundles/basic_modes/mode_definitions.moon
```diff
@@ -299,9 +299,9 @@ common_auto_pairs = {
     comment_syntax: '#'
     indentation: {
       more_after: {
-        { ':%s*$', 'else:%s*$' }
+        { ':%s*$', 'else%s*:%s*$' }
```
nilnor: So with this latest change it seems we're almost back to the way it used to be, except we're allowing some space before the colon. Is this what you intended?

kirbyfan64: @nilnor Nope. Fixed. :)

nilnor:
All right, all good!

PR #63, opened 2015-08-07
Add build-exec

kirbyfan64:
This is a command I've had in my `init.moon` for a while now since it's quite handy, so I decided to send it in to see if you guys like it. This is *really* useful with projects written in compiled languages (e.g. C++, Crystal, etc.) and also works quite well with my [project extension](https://github.com/kirbyfan64/howl-proj).

### DIFF
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -318,20 +318,33 @@ launch_cmd = (working_directory, cmd) ->
   editor.cursor\eof!
   buffer\pump!
 
+get_project_root = ->
+  buffer = app.editor and app.editor.buffer
+  file = buffer.file or buffer.directory
+  error "No file associated with the current view" unless file
+  project = Project.get_for_file file
+  error "No project associated with #{file}" unless project
+  return project.root
+
 command.register
   name: 'project-exec',
   description: 'Runs an external command from within the project directory'
-  input: ->
-    buffer = app.editor and app.editor.buffer
-    file = buffer.file or buffer.directory
-    error "No file associated with the current view" unless file
-    project = Project.get_for_file file
-    error "No project associated with #{file}" unless project
-    return interact.get_external_command path: project.root
+  input: -> interact.get_external_command path: get_project_root!.root
   handler: launch_cmd
+
+command.register
+  name: 'build-exec'
+  description: 'Runs the command in config.build_command from within the project directory'
+  handler: -> launch_cmd get_project_root!.root, config.build_command
 
 command.register
   name: 'exec',
   description: 'Runs an external command'
   input: (path=nil) -> interact.get_external_command :path
   handler: launch_cmd
+
+config.define
+  name: 'build_command'
+  description: 'The command to execute when build-exec is run'
+  default: 'make'
+  type_of: 'string'
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -329,13 +329,13 @@ get_project_root = ->
 command.register
   name: 'project-exec',
   description: 'Runs an external command from within the project directory'
-  input: -> interact.get_external_command path: get_project_root!.root
+  input: -> interact.get_external_command path: get_project_root!
   handler: launch_cmd
 
 command.register
   name: 'project-build'
   description: 'Runs the command in config.project_build_command from within the project directory'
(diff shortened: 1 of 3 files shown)


nilnor:
Seems useful to me. @shalabhc?

shalabhc:
Seems reasonable to me too. However I think the command name (and perhaps the config name) should have the 'project' prefix to make it clearer they operate on the project level (e.g. 'project-build').

kirbyfan64:
@shalabhc Updated.

kirbyfan64:
Bump?

shalabhc:
Is app_commands.moon a good place for this config definition, @nilnor? Otherwise this looks great. 

For the future I envision better support for other project level commands (multiple build commands, run command etc.) and also better support for external commands that are run repeatedly (e.g. one buffer per command rather than per invocation, displaying the status in the status line or buffer list, ability to list just the running commands, ability to rerun from the command output buffer, or the list etc.). @kirbyfan64  do you have any feedback or other ideas around this?

kirbyfan64:
@shalabhc That all sounds very nice, but what I *really* find myself constantly wishing for is command history...

I'm not sure if a run command is really needed, since many build systems will automatically run tests anyway...

I kind of think that

kirbyfan64:
(stupid phone)

that's beyond the scope of this PR. I'm happy with `ctrl_shift_b` for now. :)

shalabhc:
@kirbyfan64 
> what I really find myself constantly wishing for is command history

I'm guessing you haven't tried `alt_x` and then `up`?

> that's beyond the scope of this PR.

Agreed - this PR is good for now. If we implement any of the ideas I suggested, we could easily integrate this command with those.

nilnor:
All right, looks good - and thanks for reviewing @shalabhc! The other project command are currently in `app_commands`, so it seems like a good place for this as well. @kirbyfan64: As @shalabhc pointed command history now exists :) It's new since 0.2 though so you might have missed it fly by along with the big interactions rework.

shalabhc:
@nilnor I was referring to a good place for the `config.define` not the `command.register`.

nilnor:
Ah, I see. Well I considered that as well, but couldn't find any more fitting place at the moment, and there is precedent for defining the options close to the relevant code so it works for me for now. Perhaps there'll be reasons for consolidating options later, but so far there's only core variable and that didn't seem like a tight fit either.

kirbyfan64:
@shalabhc

> what I really find myself constantly wishing for is command history

I was referring to shell commands, not Howl commands. :)

nilnor:
@kirbyfan64 
> I was referring to shell commands, not Howl commands. :)

What do you mean, a scoped history of Howl commands that run external commands or something else?

kirbyfan64:
@nilnor I was just referring to shell command history like Bash has, but local to that session.

nilnor:
@kirbyfan64 So like a terminal emulator thing in Howl? I'm not sure I understand exactly what you're envisioning.

kirbyfan64:
@nilnor Not really.

Basically, I want to be able to use the Up and Down arrows to navigate between the commands I've run before in `*-exec`.

nilnor:
So that opening say `project-exec` you'll be able to navigate between previously run commands from there, by utilizing the arrow keys from the exec prompt then? Should be doable, and possible in a generic fashion for all commands.

kirbyfan64:
Yup.

shalabhc:
@kirbyfan64 - I understand this is not exactly what you want for now you could invoke the regular history and then type 'exec' to show only matching commands, and then you use use up-arrow.

@nilnor - I'm not sure how exactly such a generic history would work. Are you suggesting that using `up` *after* invoking a command would be handled not by the invoked command but by some generic history handler that lets you select and then *replaces* the running command with the one that the user selected? Something like that may be possible, yes. Initially I assumed each command would handle `up` independently (and possibly differently) but a generic mechanism for this might be ok too.

shalabhc:
Consider an invoked command may already be displaying a list and handle `up` to change the selected item. Not sure how this command would switch to 'history-mode'.

nilnor:
@shalabhc hence the "possibly" (at least that's the spelling I aimed for) :)

You're the expert there, I was mostly speculating about the possibility. Still, an interaction should of course then still be able to override this, which might make for inconsistencies. The best solution would likely be to do this specifically for the external command interaction by handling keys there and keeping an internal history.

PR #64, opened 2015-08-10
A proper Python bundle

shalabhc:


bundles/python/python_mode.moon
```diff
@@ -0,0 +1,34 @@
...
+  lexer: bundle_load('python_lexer')
+
+  comment_syntax: '#'
+
+  default_config:
+    word_pattern: r'\\b[\\w_][\\w\\d_]+\\b'
+
+  indentation: {
+    more_after: {
+      r':\\s*$',
```
nilnor: Nitpick - no need for a regex here.

bundles/python/python_mode.moon
```diff
@@ -0,0 +1,34 @@
+-- Copyright 2012-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+{
+  lexer: bundle_load('python_lexer')
+
+  comment_syntax: '#'
+
+  default_config:
+    word_pattern: r'\\b[\\w_][\\w\\d_]+\\b'
```
nilnor: This is not really sufficient for Python 3 I believe, since unicode is supported in variable and function names. This could I believe be supported by using the `\pL` unicode character class (used currently in places where unicode letters should be matched). See [this](https://developer.gnome.org/glib/stable/glib-regex-syntax.html#id-1.5.25.4.11) for reference. Apparently the same holds true for the Ruby mode it seems.

bundles/python/misc/examples.py
```diff
@@ -0,0 +1,252 @@
...
+double "quotes" ok too!
+'''
+
+triple_dq_strings = """can span
+multiple lines easily
+single 'quotes' ok - no escaping needed
+double "quotes" ok too!
+"""
+
+adjascent_strings_auto_concat = "hello " "world"
```
nilnor: Speling.

bundles/python/python_mode.moon
shalabhc: Ah, good catch. This is also a big problem for the lexer then? `alpha` wont match all unicode letters. It also seems to depend on the current locale which would make the behavior inconsistent.

bundles/python/misc/examples.py
```diff
@@ -0,0 +1,252 @@
...
+FloatingPointError, FutureWarning, GeneratorExit, IOError,
+ImportError, ImportWarning, IndentationError, IndexError, KeyError,
+KeyboardInterrupt, LookupError, MemoryError, NameError, None,
+NotImplemented, NotImplementedError, OSError, OverflowError,
+PendingDeprecationWarning, ReferenceError, RuntimeError, RuntimeWarning,
+StandardError, StopIteration, SyntaxError, SyntaxWarning, SystemError,
+SystemExit, TabError, True, TypeError, UnboundLocalError,
+UnicodeDecodeError, UnicodeEncodeError, UnicodeError,
+UnicodeTranslateError, UnicodeWarning, UserWarning, ValueError,
+Warning, ZeroDivisionError, __debug__, __doc__, __import__,
```
nilnor: When trying this all the `__foo__` tokens are lexed as `identifier`, intentional? There's a `special` styles that might be more appropriate for this.

bundles/python/python_lexer.moon
```diff
@@ -0,0 +1,99 @@
...
+  binary = P'0' * S'bB' * R'01'^1
+
+  digit_run = digit^1
+  float = digit_run * '.' * digit_run
+  integer = digit_run
+
+  basic_number = c 'number', any { hexadecimal, octal, binary, float, integer }
+
+  long_integer = c('number', digit_run) * c('operator', S'lL')
+  exponent_float = c('number', any { float, integer }) * c('operator', S'eE') * c('number', S'+-'^-1 * integer)
```
nilnor: I think this looks good, but is slightly weirder from a conceptual perspective. A number containing an exponent is a number followed by an operator followed by a number, where the operator isn't really an operator?

bundles/python/misc/examples.py
```diff
@@ -0,0 +1,252 @@
...
+single 'quotes' ok - no escaping needed
+double "quotes" ok too!
+"""
+
+adjascent_strings_auto_concat = "hello " "world"
+same_as_above = "hello""world"
+
+if False:
+  # Dictionaries
+  d = {a: b, c: d, "s": 1234 }
```
nilnor: Completely up to you, but it would be nice to lex the keys in the dictionary as `key`, and the same for kwargs. It's more complex though, and I'm perfectly fine with it not being done.

bundles/python/misc/examples.py
```diff
@@ -0,0 +1,252 @@
...
+
+  # Sets
+  s = {a, b, c}
+
+  # String conversions
```
nilnor: These aren't actually lexed in any special way?

bundles/python/python_mode.moon
nilnor: Yes, that would also show up in the lexer for the alpha matcher. It's hard to say what the best way of handling this would be, perhaps a final case for non-matched tokens that would classify them as identifiers (e.g. if not a number, string, etc..). Correctly checking unicode character classes seems like it would be to much work and have performance implications, but perhaps that's doable as well.

bundles/python/python_lexer.moon
```diff
@@ -0,0 +1,99 @@
...
+    'SyntaxWarning', 'SystemError', 'SystemExit', 'TabError', 'True', 'TypeError',
+    'UnboundLocalError', 'UnicodeDecodeError', 'UnicodeEncodeError',
+    'UnicodeError', 'UnicodeTranslateError', 'UnicodeWarning', 'UserWarning',
+    'ValueError', 'Warning', 'ZeroDivisionError'
+  }
+
+  comment = c 'comment', P'#' * scan_until(eol)
+  operator = c 'operator', S'+-*/%~&^=<>;:,.(){}[]|'
+
+  name = (alpha + '_')^1 * (alpha + digit + S'_')^0
```
nilnor: The `S` here should be a `P` by the way.

bundles/python/misc/examples.py
kirbyfan64: @nilnor That might look weird for dictionary keys, since it's not uncommon to get code like `{f()[1][0]: 1}`.

nilnor: For those cases the key expressions wouldn't be lexed as keys, but rather as identifiers, operators, etc. I was thinking plain identifiers, and perhaps string keys. So `{yes: 1, 'perhaps': 1, no(1)[0]: 0}`

nilnor:
All in all it looks solid, and it's definitely a big improvement :) I'll let you decide what if any of the points you'd like to address. I'm not a heavy Python user myself, but between you and @kirbyfan64 it seems this would get some good testing :+1:

bundles/python/misc/examples.py
shalabhc: I thought of the `key` style as mainly for auto-'stringified' keys such as in Lua, Moonscript and JS. So doing it for identifiers and strings, but not really for other expressions or numbers, seems a bit odd.

I'm not sure what you mean by kwargs. Is it at the call site such as in`f(a=1, b='hello')`? Again these don't end up in a dictionary unless you explicitly capture them using `**kwargs`. E.g. you could just have a function `def f(b, a)` in which case these would just end up in regular parameters.

bundles/python/misc/examples.py
shalabhc: Yes that was intentional since I wasn't sure if styling these is useful. But using `special` for all (built-in as well as user defined) double underscore identifiers seems like a good idea.

bundles/python/misc/examples.py
nilnor: Yes, for kwargs I was referring to the call site. If you don't think it makes sense than I'll happily defer to your judgement. The `key` style is not strictly defined (an as an aside it would probably be a good idea to go through all default styles at some point and perhaps restructure these). But note that in both Lua and Moonscript it's possible to use an expression as a table key, so by that logic the lexing for those languages is also odd.

bundles/python/python_lexer.moon
shalabhc: Agree. I like the way this looks - it makes it harder to miss the important information. I also do this for the prefix in raw/encoded strings and the long number suffix(e.g. `r"abc"`, `123L`).

Do you think `special` makes more sense here?

LuaJIT's `ULL` is similar and both Rust and Nim allow suffixes like `123i32`  or `123'i32` to integers. Would be nice to style these literal modifiers differently.

bundles/python/misc/examples.py
shalabhc: There is a distinction in that bare words used as keys in Moonscript/Lua are converted to string keys, but no such thing happens in Python - it is treated as just an expression. So I think leaving it styled as any other expression makes sense. However using the `key` style in Moonscript signals to me that there is something special about them. Also, when expressions do get used as keys in Moonscript, they don't get styled as `key` - that would make it odd.

bundles/python/misc/examples.py
shalabhc: Yeah I left this out because backticks are extremely rare and have also been removed from Python 3.0. Perhaps just adding the backtick to operator is sufficient to style this better.

shalabhc:
Thanks for reviewing! @kirbyfan64 - I'm happy to hear any feedback or comments you may have about this as well.

kirbyfan64:
Looks great! (I just tried to test this, but accidentally cloned the `master` branch and was confused by the fact that there was no difference. :) The highlighting of function names seems weird at first, but it *is* more consistent.

Some slight nitpicks:

- Entering `pass`, `continue`, and `return` should un-indent. Putting code after any of these (especially `pass`) is extremely uncommon (I don't think I've ever actually seen someone do that).
- [PEP 492](https://www.python.org/dev/peps/pep-0492/) adds two new keywords, `async` and `await`, that will be included in Python 3.5. Maybe these should be highlighted as keywords, too?

kirbyfan64:
Also, in this example:

```python
class X(metaclass=123):
    pass
```

should `metaclass` be highlighted? GitHub's Linguist seems to highlight it as a keyword (as you can see above :), but `metaclass` is still a valid identifier, so it would have to be highlighted only when followed by an equals sign in a class definition, which might be slightly tricky...

kirbyfan64:
(And I think you forgot to highlight `assert` as a keyword)

shalabhc:
Thanks for the feedback @kirbyfan64!

> Entering pass, continue, and return should un-indent

I think Howl doesn't support this currently so that would have to be implemented first - something like a 'less_after' table for modes. @nilnor?

> PEP 492 adds two new keywords, async and await

Good point - I'll add these to the keywords list.

> should metaclass be highlighted

I think it should because it's more special than an arbitrary keyword argument. Perhaps names for all keyword arguments in any class definition should be highlighted the same way. I'm not sure which style should be used though.

> (And I think you forgot to highlight assert as a keyword)

Hmm, I'm guessing this got shadowed by the `as` keyword and needs to be moved to the front of the list.

nilnor:
@shalabhc: Yes, a `less_after` table would do the trick. The reason it's not there yet is simply because so far there hasn't been any users of it. @kirbyfan64 has a good point though regarding `return` et al, and that could be applied to more languages to improve indentation support.

bundles/python/misc/examples.py
nilnor: A good point - I overlooked that Python has no special syntax for dictionary keys and that it's all expressions. It certainly doesn't make sense then.

bundles/python/python_mode.moon
shalabhc: Done.

bundles/python/python_mode.moon
shalabhc: I've fixed the word_pattern but the lexer has to wait for proper support in the grammar rules.

bundles/python/misc/examples.py
shalabhc: Fixed.

bundles/python/misc/examples.py
shalabhc: Now all __foo__ identifiers are styled as 'special'.

bundles/python/python_lexer.moon
shalabhc: Changed style of these to 'special' to avoid conceptual confusion.

bundles/python/misc/examples.py
shalabhc: Added backticks to operators - I think this makes sense because they always surround a proper expression which should be styled as such.

bundles/python/python_lexer.moon
shalabhc: Fixed.

shalabhc:
@kirbyfan64, @nilnor - I think I've addressed all comments.

- less_after will have to be a separate commit that adds proper Howl support for that.
- I left out `metaclass=` styling because it was too much work to correctly style it in all cases. 
- Added 'async', 'await' and fixed 'assert' keyword.

nilnor:
+1, rebased and merged locally.

PR #65, opened 2015-08-10
A rudimentary Nim bundle

shalabhc:


kirbyfan64:
YEAH!!!

kirbyfan64:
(I was going to do this, but you beat me to it!)

kirbyfan64:
I don't think this properly supports Nim's case insensitivity. Technically, these are all equivalent:

```nim
proc
p_ROC
p_R_O_C
P__________________________________r_O_c
```

shalabhc:
Oh.. I didn't realize this applies to keywords too. Will fix.

shalabhc:
@kirbyfan64 - the case insensitivity for keywords is now fixed. It appears the first character is case sensitive and more than one contiguous underscore is not allowed. I've also fixed and simplified the number parsing. Please take another look when you get a chance!

kirbyfan64:
I haven't actually tried this, but it looks great!!

One more thing: the Nim style guide (and 99% of Nim programmers) capitalize the first letter of a type. Should it be highlighted in a special way like the MoonScript bundle?

An interesting extension (likely for a later time) would be to use [nimsuggest](http://nim-lang.org/docs/nimsuggest.html) to implement [`show-doc-at-cursor`](http://howl.io/doc/manual/editing.html#documentation-popup) support. I think this is good enough for now, though.

If (when?) this gets merged in, I'll add Howl to the list of editors with Nim support.

bundles/nim/README.md
```diff
@@ -0,0 +1,11 @@
+# Nim bundle
+
+## What is it?
+
+This is a Nim bundle for the Howl editor, providing Nim specific lexing
+and a Nim mode.
+
+## License
+
+Copyright 2012-2015 The Howl Developers
```
nilnor: Should probably be just 'Copyright 2015' to avoid mental time warps.

shalabhc:
@kirbyfan64 - great suggestions.

> the Nim style guide (and 99% of Nim programmers) capitalize the first letter of a type. Should it be highlighted in a special way like the MoonScript bundle?

Yes, this sounds reasonable. Perhaps with nimsuggest it may even be possible to accurately determine if an identifier is a type, but integrating that with lexing doesn't seem simple.

> An interesting extension (likely for a later time) would be to use nimsuggest to implement show-doc-at-cursor support.

Yes this sounds like a great idea (as a separate enhancement in the future). In fact we could also nimsuggest for completions. :)

bundles/nim/init.moon
```diff
@@ -0,0 +1,33 @@
...
+    unless howl.io.File(exe).exists
+      error "Nonexistent path '#{exe}' - invalid config.nim_executable."
+    unless howl.app.editor.buffer.file
+      error "No file associated with current buffer."
+    howl.command.run "exec #{exe} c -r " .. howl.app.editor.buffer.file.basename
+
+unload = -> howl.mode.unregister 'nim'
+
+howl.config.define
+  name: 'nim_executable',
```
nilnor: Considering the suggestion by @kirbyfan64 to incorporate `nimsuggest` at a later stage, I think it's a good idea to contemplate having a `nim_bin_dir` variable or something similar instead. This could be used to find the `nim` executable as well as other related executables. Should default to `/usr/bin` as that's where distribution package would typically place the files, which would make it "just work" at a later stage.

nilnor:
(will continue reviewing later)

bundles/nim/init.moon
kirbyfan64: @nilnor But `nimsuggest` is usually installed somewhere into `~/.nimble`...

kirbyfan64:
After actually trying this, I noticed none of the built-in types are highlighted (e.g. `string`, `int`, etc.), nor are any of the "fake types" (like `varargs`). Is this intentional?

bundles/nim/init.moon
nilnor: @kirbyfan64 
> But nimsuggest is usually installed somewhere into ~/.nimble...

Oh? The page for nimsuggest you linked to says it doesn't install properly via nimble and suggests compiling it and putting it in the same bin path as nim, which made me believe that was the standard way of doing it.

@shalabhc Even if we keep this variable it could default to just `nim`, right? Provided that `nim` is in the path somewhere already it would still just work then I think.

nilnor: Or as an alternative we could auto-find it given the current $PATH if we want the full path of the binary.

shalabhc:
@kirbyfan64 
> none of the built-in types are highlighted

Ah yes this should be fixed. The primary reason is  laziness on my part, plus not having an exhaustive list of these builtins. Any idea how I can get one?

bundles/nim/init.moon
shalabhc: Using 'nim' as the default is a good idea - it should just work if it's in the path. We don't need to know the full path.

Re nimsuggest, currently I've copied it in the directory containing nim, but I believe when fixed it will install automatically via nimble and may not be in the same directory. So that might need its own variable - not sure what we'd call that - 'nimble_executable' or 'nim_nimble_executable' (in case we want to namespace everything by the bundle name.

kirbyfan64:
http://nim-lang.org/docs/manual.html#types isn't quite exhaustive, but it's close.

bundles/nim/nim_lexer.moon
```diff
@@ -0,0 +1,94 @@
+
```
nilnor: Missing copyright notices at the top.

bundles/nim/nim_mode.moon
```diff
@@ -0,0 +1,61 @@
...
+class NimMode
+  new: =>
+    @lexer = bundle_load('nim_lexer')
+
+  comment_syntax: '#'
+
+  indentation: {
+    more_after: {
+      '[([{:=]%s*$'
+      r'^\\s*\\b(const|enum|let|type|proc|func|iterator|macro|template|method|object|tuple|var)\\s*$',
```
nilnor: The `\\b` seems excessive since the pattern is already anchored.

nilnor:
What I really miss for this bundle is a syntax example file.

bundles/nim/nim_lexer.moon
```diff
@@ -0,0 +1,94 @@
...
+   hexadecimal_number
+   binary_number
+   simple_number
+  }
+
+  number = c('number', simple_number) * exponent_suffix * (float_size_suffix^-1)
+  number += c('number', number_with_point) * (exponent_suffix^-1) * (float_size_suffix^-1)
+  number += (integer * integer_size_suffix^-1)
+  number *= #-(alpha + digit + S'_') -- no alphanum should be attached to the number
+  -- number = integer
```
nilnor: Old code left in?

nilnor:
Finished looking it over, looks good as far as I can tell :) 

Like I commented above, a syntax example file containing all the lexed constructs is what I really miss, doing any kind of larger changes to the lexer later without breaking stuff will be unnecessarily hard later. Such a file also gives a very nice overview over exactly what is lexed and how it appears.

shalabhc:
I believe I've addressed all issues. In particular, anything that looks like a type (starts with uppercase) is styled as a class and all builtin system types are also styled as classes. I've also added an example file and made some bug fixes.

bundles/nim/init.moon
```diff
@@ -0,0 +1,36 @@
...
+
+howl.command.register
+  name: 'nim-run'
+  description: 'Compile and run the current Nim file'
+  handler: ->
+    unless howl.app.editor.buffer.file
+      error "No file associated with current buffer."
+
+    exe = howl.app.editor.buffer.config.nim_executable or 'nim'
+    unless howl.io.File(exe).exists
```
nilnor: This won't work when exe is "nim". Either forego the check completely or add a path search for finding the full path. It's a shame to lose the informational warning/error pointing to the correct configuration variable though if we just go with skipping the check.

shalabhc: Good catch. Now fixed - please look when you get a chance.

nilnor:
All right, looks good!

PR #73, opened 2015-08-29
Fix null error message

kirbyfan64:
Without this, for every other character I enter, I get an error message about indexing `nil` field `cur_style`.


nilnor:
Thanks! The API has changed to returning `nil` for unstyled text.

PR #80, opened 2015-09-03
Basic support for XDG-complaint config directory.

eoranged:
If there are `$XDG_CONFIG_HOME/howl` or `~/.config/howl` directory, use one to load config files.  
Requires user to manually change config directory location with `mv ~/.howl ~/.config/howl` otherwise will work as before.

See [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) for more info.

### DIFF
--- a/lib/howl/settings.moon
+++ b/lib/howl/settings.moon
@@ -8,7 +8,7 @@ default_dir = ->
   return File(howl_dir) if howl_dir
   home = os.getenv('HOME')
   xdg_config_home = os.getenv('XDG_CONFIG_HOME')
-  -- if none of this are set, we wwon't be able to find config
+  -- if none of this are set, we won't be able to find config
   return nil unless home or xdg_config_home
   xdg_conf_dir = nil
   if xdg_config_home
--- a/lib/howl/settings.moon
+++ b/lib/howl/settings.moon
@@ -10,17 +10,20 @@ default_dir = ->
   xdg_config_home = os.getenv('XDG_CONFIG_HOME')
   -- if none of this are set, we wwon't be able to find config
   return nil unless home or xdg_config_home
-  -- trying ~/.howl first
-  dotdir = nil
-  if home
-    dotdir = File(home)\join('.howl')
-    return dotdir if dotdir.is_directory
-  -- trying xdg-complaint ~/.config/howl
   xdg_conf_dir = nil
   if xdg_config_home
     xdg_conf_dir = File(xdg_config_home)\join('howl')
   elseif home
     xdg_conf_dir = File(home)\join('.config')\join('howl')
+  -- trying ~/.howl first
+  dotdir = nil
+  if home
+    dotdir = File(home)\join('.howl')
+    if dotdir.is_directory
+      if xdg_conf_dir and xdg_conf_dir.is_directory
+        howl.log.warn("Ignoring #{xdg_conf_dir.path} in favor of #{dotdir.path}")
+      return dotdir
+  -- trying xdg-complaint ~/.config/howl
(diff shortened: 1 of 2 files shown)


eoranged:
By the way I would make It default, but I don't want to change default behavior without consulting with you guys first.

nilnor:
I'm fine with checking `${XDG_CONFIG_HOME}/.howl` and loading from that if it exists (and `~/.howl` doesn't), as it seems that would clearly indicate some preference on part of the user. I wouldn't want it as the default however. For starters the directory can contain not only configuration but also custom user code (e.g. personal bundles) so that makes the use of a clearly marked configuration dubious in my eyes. It would also make the default location of the Howl dir non-deterministic which makes for confusion.

We might be able to tighten up the code a little bit later, but for now what I really miss are the specs :) The current behaviour has tests [here](https://github.com/howl-editor/howl/blob/master/spec/settings_spec.moon), so this should be updated for the new order of things. You can run that test using the `bin/howl-spec` script, after installing the needed busted version using luarocks (if you don't have any of test run the script and it will tell you). Let me know if you need any help.

shalabhc:
I want to propose that warning be printed if the ``.howl`` directory exists in both locations, one of which gets ignored.

eoranged:
@nilnor okay, I've got your point.
As for `bin/howl-spec` it fails with strange

    ** (howl:32199): CRITICAL **: Failed to run script: /home/eoranged/dev/stuff/howl/lib/howl/init.lua:144: /home/eoranged/dev/stuff/howl/spec/support/spec_helper.moon:34: attempt to index global 'assert' (a function value)

error.

@shalabhc nice catch. I'm trying to figure out the way of displaying such warning.

shalabhc:
@eoranged - you can use howl.log.warn('message')

Also, your error might be because you are using a different version of some module. Here are the versions that I have currently:

ansicolors
   1.0.2-3 (installed) - /usr/local/lib/luarocks/rocks

busted
   1.11.1-2 (installed) - /usr/local/lib/luarocks/rocks

dkjson
   2.5-2 (installed) - /usr/local/lib/luarocks/rocks

lua_cliargs
   2.5-1 (installed) - /usr/local/lib/luarocks/rocks

luafilesystem
   1.6.3-1 (installed) - /usr/local/lib/luarocks/rocks

luassert
   1.7.6-0 (installed) - /usr/local/lib/luarocks/rocks

penlight
   1.3.2-2 (installed) - /usr/local/lib/luarocks/rocks

say
   1.3-1 (installed) - /usr/local/lib/luarocks/rocks

eoranged:
Thanks, @shalabhc. It was caused by installing busted-2.0 (Nils fixed the warning in howl-spec just an hour ago).

nilnor:
@eoranged sorry about that, I should have thought to give you a heads up!

eoranged:
@nilnor no problem. I'll try to take time to finish the spec this week.

eoranged:
Okay, guys. I've added some tests and a warning.  
Warning is also printed during the test. Is it expected behaviour?

nilnor:
And it's in, thanks! 

The warning being printed during the test was because the determination of whether we're in test mode or not is based on the environment, and since the specs stubbed out `os.getenv` there was a unfortunate side effect. I rewrote the spec to use `howl.sys.env`, which is a newer addition that also lets you set the environment, so that became cleaner. I also fixed another (pre-existing) case where we now raise an error if neither HOME or XDG_CONFIG_DIR are present, rather than failing with a `nil` value later. Merged manually.

eoranged:
Great, thanks!

PR #83, opened 2015-09-07
Add a Kotlin bundle

kirbyfan64:
Note that I'm still learning Kotlin, so there may be a few issues here!

Also note that *this does not implement Kotlin's string interpolation.* Basically, I'm still trying to figure out how the heck the Ruby lexer does that...

### DIFF
--- a/bundles/kotlin/init.moon
+++ b/bundles/kotlin/init.moon
@@ -1,0 +1,16 @@
+howl.mode.register
+  name: 'kotlin'
+  extensions: {'kt', 'kts'}
+  aliases: {'kt', 'kts'}
+  create: -> bundle_load 'kotlin_mode'
+  parent: 'curly_mode'
+
+unload = -> howl.mode.unregister 'kotlin'
+
+{
+  info:
+    author: 'The Howl Developers'
+    description: 'A Kotlin bundle'
+    license: 'MIT'
+  :unload
+}
--- a/bundles/kotlin/kotlin_lexer.moon
+++ b/bundles/kotlin/kotlin_lexer.moon
@@ -23,8 +23,8 @@ howl.aux.lpeg_lexer ->
   special = c 'special', word { 'true', 'false', 'null' }
 
   string = c 'string', any {
-    span '"', '"', '\\'
     span '"""', '"""', '\\'
+    span '"', '"', '\\'
   }
 
   char = span "'", "'", '\\'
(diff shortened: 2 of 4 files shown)


nilnor:
I don't see an example kotlin syntax file there :P 

Seriously, I won't accept any new language bundle without a corresponding comprehensive example file. It also helps tremendously while developing a lexer, so I very much suggest creating one as you go along with the lexer. Use the `bundle-reload-current` with the file open and you get instant feedback on lexer changes and how/if they work.

String interpolation is trickier than other more basic lexing, as you need to use Lpeg grammars for them. The Ruby lexer is not a good source for inspiration, as the Ruby syntax is very complex which makes the lexer very complex. I think the moonscript lexer is a better simpler example that should works as for Kotlin. Provided you got a good example file going I could help you with this if you get stuck.

Finally, there's a bug in the current lexer. Create an example file and paste in this example code (from the [Kotlin ref](http://kotlinlang.org/docs/reference/basic-types.html#string-literals):

```kotlin
val text = """
  for (c in "foo")
    print(c)
"""
```

kirbyfan64:
> I don't see an example kotlin syntax file there :P

Ah, crap. I had one, but I forgot to `git add` it. :/

> I think the moonscript lexer is a better simpler example

That's a great example; I think I get how to implement it now.

> Finally, there's a bug in the current lexer. 

Seems I forgot that the order of patterns in an `any` block matters.

kirbyfan64:
@nilnor Uhh...well...

I pretty much copied-and-pasted the interpolation code from MoonScript. Yet, I keep getting this:

```
*error in 'signal open' callback handler: ../lib/aullar/styling.moon:137: attempt to perform arithmetic on local 'styling_end' (a string value)
```

nilnor:
Well, you "pretty much" copied and pasted the code :)

As it is, the error you're seeing is because of the incorrect `string` definition, i.e.:

```moonscript
string: c 'string', any {
  c('string', '"') * V'q_string_chunk'
```
With this you're telling the lexer capture all results of the `any` block as a string, with predictably disastrous results. Consider the following, slimmed down working code:

```moonscript
    string: any {
      c('string', '"') * V'q_string_chunk'
    }
    q_interpolation: #P'$' * (-P'}' * (V'all' + 1))^1 * c('operator', '}') * V'q_string_chunk'
    q_string_chunk: c('string', scan_to(P'"' + #P'${', P'\\')) * V('q_interpolation')^0
```

So removing the first incorrect capture should get you going.

Some other points I thought of while looking at this: 

- I don't know if you're done with the example file, but it needs to be much more comprehensive.
- Have a look at what Kotlin actually allows within string interpolations to make sure you don't allow more than you should
- Should triple-string actually be interpolated at all? They are described as raw strings with no escaping allowed in the docs (but I just had a quick look)

kirbyfan64:
@nilnor All the issues are now fixed.

PR #84, opened 2015-09-07
Implemented history for exec and project-exec

shalabhc:
The `up` key in 'exec' and 'project-exec' commands now pulls up previously run exec commands. History is not persisted across sessions yet.

@kirbyfan64 - did you get a chance to try this out?


nilnor:
+1

PR #85, opened 2015-09-07
OSX support

kirbyfan64:
...of course, the `alt` key still has that weirdness and all, but this is a start!!

### DIFF
--- a/lib/howl/ui/theme.moon
+++ b/lib/howl/ui/theme.moon
@@ -226,7 +226,7 @@ with config
   .define
     name: 'font'
     description: 'The main font used within the application'
-    default: 'Liberation Mono'
+    default: if ffi.os == 'OSX' then 'Monaco' else 'Liberation Mono'
     type_of: 'string'
     scope: 'global'
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,5 +1,7 @@
 PREFIX ?= /usr/local
 
+UNAME_S := $(shell uname -s)
+
 GTK = gtk+-3.0
 GTK_CFLAGS = $(shell pkg-config --cflags $(GTK))
 GTK_LIBS = $(shell pkg-config --libs $(GTK) gmodule-2.0 gio-unix-2.0)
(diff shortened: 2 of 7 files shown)


lib/howl/ui/theme.moon
```diff
@@ -226,7 +226,7 @@ with config
   .define
     name: 'font'
     description: 'The main font used within the application'
-    default: 'Liberation Mono'
+    default: if ffi.os == 'OSX' then 'Monaco' else 'Liberation Mono'
```
nilnor: This could be a comma separated list instead, e.g. `Liberation Mono,Monaco` without any OS checking. Also, should be `Monaco` or `Menlo` be preferred? @bunster?

src/Makefile
```diff
@@ -32,10 +34,20 @@ SCI_GEN_TARGET = ../lib/howl/scintilla.lua
...
-LD_FLAGS = -Wl,-E
+ifeq ($(UNAME_S),Darwin)
+	LD_FLAGS = -Wl,-export_dynamic -pagezero_size 10000 -image_base 100000000
+else
+	LD_FLAGS = -Wl,-E
+endif
 OBJECTS = main.o process_helpers.o
 DEP_OBJECTS = $(LPEG_OBJECT)
 
+ifeq ($(UNAME_S),Darwin)
```
nilnor: Since `md5` isn't used in the Makefile itself, I think it would be better to have this conditional directly in the `download` script.

src/process_helpers.c
```diff
@@ -5,6 +5,15 @@
 #include <sys/wait.h>
 #include <signal.h>
 
+#ifndef __linux
```
nilnor: Rather than mucking around with these signals  we should just remove them instead. I added a rather full list of signals when I added the process class, but these I shouldn't have added at all so let's just remove them instead from process.

nilnor:
So I have some change I'd like to see before this goes in. If you want to, you can close this and open a PR from a branch within the main howl repository instead and make a PR from that. That way me and @shalabhc can help out with adding commits directly.

PR #86, opened 2015-09-08
Implement preview for 'open' command

shalabhc:
Adds preview functionality to `interact.select_file`

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Added previews for the `open` command.
+
 - Added a new command, `project-build` that executes a pre-configured command
 from the projects root directory (using the command configured in the new
 `project_build_command` variable).
--- a/lib/howl/interactions/file_selection.moon
+++ b/lib/howl/interactions/file_selection.moon
@@ -3,8 +3,9 @@
 
 glib = require 'ljglibs.glib'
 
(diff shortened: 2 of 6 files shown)


nilnor:
All right, looks good!

Here are some suggestions that could be added (none of which are required apart from the Changelog one):

- Add it to the Changelog :)
- The previews for directories could be improved if one wants to, e.g.
  * The path to the directory can be shorted (using `~`)
  * The preview could contain a fancy styled directory listing
  * If the previous point were implemented, perhaps a `config.preview_directories` var would be in order
- The preview module should be easily testable if so desired

Merge at will, :+1:

nilnor:
Oh, I updated the Changelog just now, so a rebase and edit before the final merge is probably a good idea.

shalabhc:
But a rebase will lose the comments on the PR, no?

I thought about directory preview too but wasn't sure of the performance (i.e. lag while scrolling over large directories, even when you don't intend to visit it). I think I can still do it as a separate PR in a non-blocking way.

nilnor:
Yes, I believe you're right in that a rebase would loose the history, so probably better to just merge it. And there's a valid point about the performance, that's why I suggested the separate config variable. However, it might be that it's simply not worth it, or not being actually useful enough that it would warrant it. Leaving it out sounds good for now at any rate.

PR #87, opened 2015-09-08
OSX support 2

kirbyfan64:


nilnor:
Looked great with the changes. I applied it manually to master, in order to get rid of the merges. Thanks!

The remaining thing then is to have a look at the key bindings - I have started the work a little bit, hopefully I can finish off a first version some soonish.

PR #88, opened 2015-09-14
Stricter auto_module loading

shalabhc:
Added check for the existance of a directory before
creating and auto_module table for an import.

Fixed various invalid imports.


nilnor:
Fantastic :)

PR #90, opened 2015-09-15
JS and CS lexer fixes

kirbyfan64:


nilnor:
Looks good, but please add an example of the expression/statement not being lexed correctly to the coffeescript example file.

kirbyfan64:
@nilnor Done! I also added some support for [ECMAScript 6 generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*).

nilnor:
:+1:

PR #91, opened 2015-09-15
Use a text mouse cursor for the aullar edit area

### LINKED ISSUE #76: Set correct (text entry) cursor for aullar widget
As mentioned in #72, the standard pointer cursor is currently used within the aullar editing widget, where a text editing pointer would be expected.

nilnor:
Fixes #76.


kirbyfan64:
Thanks!

Issue #68, opened 2015-08-24
OS X support


bunster:
I'm running on an OSX and during the make process, this error pops up. I've installed glib via Homebrew and still it calls that error. Anyone know what I can do to fix this?

kirbyfan64:
What's the full error message? I'm pretty sure there has to be *something* from `pkg-config`...

bunster:
[Here](http://pastebin.com/raw.php?i=mFZ9p7HS) is the paste of it from start-to-error.

bunster:
I notice it said I didn't have gtk+-3.0, currently I have gtk+2.24.28. That may be part of the problem.

kirbyfan64:
You need to install Gtk+ 3:

```
$ brew install gtk+3
```

You might also want to install `md5sum` to get rid of the checksum warnings:

```
brew install md5sha1sum
```

I read that OSX's `md5` command has an option `-r` that causes it to work the same way as `md5sum`. Should the Makefile use that on OSX?

bunster:
Aye, I am just installing gtk+3 now :) And I installed md5sha1sum too. I'll try again once this is done and update the posts here!

kirbyfan64:
I put together the following patch:

```diff
diff --git a/src/process_helpers.c b/src/process_helpers.c
index d71066e..efc33b3 100644
--- a/src/process_helpers.c
+++ b/src/process_helpers.c
@@ -25,7 +25,10 @@ int sig_USR2 = SIGUSR2;
 int sig_PIPE = SIGPIPE;
 int sig_ALRM = SIGALRM;
 int sig_TERM = SIGTERM;
+#ifdef __linux
 int sig_STKFLT = SIGSTKFLT;
+int sig_PWR = SIGPWR;
+#endif
 int sig_CHLD = SIGCHLD;
 int sig_CONT = SIGCONT;
 int sig_STOP = SIGSTOP;
@@ -38,6 +41,10 @@ int sig_XFSZ = SIGXFSZ;
 int sig_VTALRM = SIGVTALRM;
 int sig_PROF = SIGPROF;
 int sig_WINCH = SIGWINCH;
+#ifdef _DARWIN_C_SOURCE
+int sig_IOT = SIGIOT;
+int sig_EMT = SIGEMT;
+#else
 int sig_POLL = SIGPOLL;
-int sig_PWR = SIGPWR;
+#endif
 int sig_SYS = SIGSYS;
```

[...]

PR #95, opened 2015-09-17
Uni bindings (OS X)

nilnor:
This is still work in progress, but lays a possible foundation for generic key bindings across *nix and OS X., to make Howl usable from OS X (issue #68). What is does is that it adds a new, synthetic, modifier "action". On non-OS X platforms `ctrl` will map to `action`, in addition to control. On OS X the `meta` modifier will map to `action`. Example from `describe-key` on Linux:

```mooonscript
{
  "ctrl_S",
  "ctrl_shift_s",
  "ctrl_shift_83",
  "action_S",
  "action_shift_s",
  "action_shift_83",
}
```

This also adds a new binding, `action_shift_x` for bringing up the command line.

Remaining:

- Update binding documentation
- Verify bindings
- Determine how to handle the current bindings which use `alt`


kirbyfan64:
@nilnor Mind if I push doc fixes to the `uni-bindings` branch?

nilnor:
@kirbyfan64 On the contrary :)

shalabhc:
While having a synthetic modifier seems reasonable, applying it to all bindings doesn't quite make the shortcut keys consistent with OSX. While something like `meta_c` (editor-copy) will work as expected, something like ``meta_q`` (show-doc-at-cursor) is not quite right because I expect it to quit the current application on OSX. Another one is `meta_up` which should got to the beginning of the document, not scroll the view up. Here is a quick reference: https://support.apple.com/en-us/HT201236

Also, consider that Control+P and not Command+P moves up one line on OSX, so I'd expect that to move the selection up in a menu_popup.

For the top level bindings, maybe a base `keymap.moon` can specify `action_*` keys and an override `keymap-osx.moon` can fix-up the inconsistencies? Fixing the shortcuts within an interaction's keymap is harder. I can't think of anything better than introducing aliases such as:

```
  keymap:
     [keys.up]: ...
     [keys.down]: ...
```

This adds complexity due to another level of indirection. OTOH, it also allows Linux users to fully customize these interactions, if they want.

kirbyfan64:
I don't know...

Having OSX bindings seems nice, but makes documentation harder and is really confusing. Really, Howl already doesn't follow normal conventions (`ctrl_shift_a` is select all in Howl, while `ctrl_a` is used by almost everything else). This would just be kind of weird.

I need to see what Atom does here again. I don't like it a lot, but they've evidently already encountered thia problem.

shalabhc:
Looking at atom, sublime or numerous other cross platform editors may be useful, but we should still discuss what is the ideal behavior for Howl. Of course, we may end up saying we can't shoot for that because of cost or effort needed.

Anyway, if you're concerned about documentation, we could also automate some of that. E.g. if you write ``Press [key:open] to open a file`` - it could be automatically filled in.

kirbyfan64:
Hmmm, maybe, I should ask by brother about this. He uses Macs all the time and has tried a bunch of editors. I've been trying to convince him to try Howl for a long time anyway...

nilnor:
@shalabhc Good points. As you say we could override things for the base keymap, but it's unclear how to handle other bindings such as for interactions. Introducing aliases seems cumbersome and hard to understand to me, I'm not even sure what those aliases would be or how they would differ from the bindings?

Another possible solution would be to add OS specific binding dispatching by allowing keymaps to have explicit OS bindings, e.g. `{for_os: { osx: {..} }`. Not crazy about that, but then I don't see a super clean solution for this.

shalabhc:
@nilnor - yes aliasing could become cumbersome. Developing the idea a bit further, we could have interactions forego the `keymap` and instead define an `actions` table. Something like this:

```moonscript
class Interaction
  run: ...
  on_update: ...
  actions:
    commit: => self.finish @current_item
    cancel:=> self.finish!
    up: => @select_previous!
    down: => @select_next!
    back: => @updir!
```
And then separately map keys to the actions:

```moonscript
interact.register
  name: 'my_interaction'
  factory: MyInteraction
  keymap:
   enter: 'select'
   escape: 'cancel'
   backspace: 'back'
   by_os:
    osx: 
      ctrl_p: 'up'
      ctrl_n: 'down'
    linux:
      up: 'up'
      down: 'down'
```
 
I'm not quite sure what would be the best place for the key -> action mapping. Perhaps it could even use the usual bindings:

```moonscript
bindings.push_for 'my_interaction':
  ctrl_p: 'up'
  ctrl_n: 'down'
```

Advantages of actions:
- Automation may be easier, i.e. if you are driving the interaction programatically.
- User can change any key binding within interactions.

Disadvantages:
- Another level of indirection and associated complexity.

nilnor:
So, going back to this - I'm thinking now just having a `for_os` section in a keymap is the simplest and by extension most understandable approach. Any strong opposition to just doing that?

nilnor:
So, going back to this - I'm thinking now just having a `for_os` section in a keymap is the simplest and by extension most understandable approach. Any strong opposition to just doing that?

shalabhc:
No, lets go ahead with `for_os`. Will 'linux' also mean 'windows' in this table?

nilnor:
@shalabhc No, each OS will be separate, so 'linux' and 'windows' will have separate stanzas.

shalabhc:
Sounds good - lets do this! Excited about proper OSX support :smiley:

kirbyfan64:
Will there be one for both Linux and Windows, since they both share a lot of keybindings?

kirbyfan64:
Also, is there a particular reason for a `for_os` section over Howl just identifying nested tables?

shalabhc:
> Will there be one for both Linux and Windows, since they both share a lot of keybindings?
Or maybe if the 'windows' stanza doesn't exist it just falls back to 'linux'? (and same for 'osx')

> Also, is there a particular reason for a for_os section over Howl just identifying nested tables?
Personally I don't have a preference and either could work, but I'll note that there's already a precedent with the 'binding_for' key (http://howl.io/doc/api/bindings.html#indirect-bindings). Also, nested tables are used to specify a temporary keymaps for sequential keystrokes (e.g. `ctrl_x: ctrl_s: ...`) and though the OS names are very unlikely to conflict with a key name, perhaps its best to group them into one parent key.

kirbyfan64:
@shalabhc Ah, yes...I forgot that was already taken.

nilnor:
The way I'm thinking about is that OS X is really the odd one out, and that the `for_os` mechanism is for overrides where they would be needed. Thus most of the stuff would be shared, as is currently today, in the default key map. When there's need for overriding or adding something specific for an OS that can be done using the `for_os`. So I'm not thinking that everything should be split and make heavy use of this, nor am I'm aware of lots of OS specific things that would need that? Apart from OS X then.

nilnor:
Added support for the `for_os` sections now. The question I had while updating the documentation is whether we still want the synthetic `action`, or whether we just bite the bullet and add specific bindings for OS X everywhere? The pros of going that route now would be to not have to introduce a new concept, and the cons, like I think we've touched upon previously, that all specific key maps would have to consider OS explicitly.

shalabhc:
>  The question I had while updating the documentation is whether we still want the synthetic action, or whether we just bite the bullet and add specific bindings for OS X everywhere?

I'm in favor of removing `action` and always using explicit OS X bindings. Otherwise we would have two ways of doing things and it would get confusing. Also, I think deliberate thought for platform specific key bindings will lead to better usability.

> The way I'm thinking about is that OS X is really the odd one out, and that the for_os mechanism is for overrides where they would be needed

This could mean OS X will end up with unintended bindings (i.e the top level keys in the keymap intended for Linux and Windows). Not sure if this will be an issue in practice but maybe we should have a guideline that all platform specific bindings go entirely inside the `for_os` section?

nilnor:
Let's get rid of `action` then, that feels cleaner - I'll open a new PR for the new approach. And yes, it could potentially mean that OS X ends up with unintentional bindings, but that doesn't seem like a giant issue right now, particularly compared to the complexity of the alternative approach where we immediately end up talking about fallbacks, etc. Should this actually be a problem in practice we can revise it later.

nilnor:
Closing in favour of #119.

PR #96, opened 2015-09-17
Fix not scrolling far enough to the right when searching for text

kirbyfan64:
To reproduce the issue:

1. Open up a new file.
2. On the first line, type some text (I did `abc`).
3. Enter the same thing on the second line, but put a bunch of spaces before the text so that you need to scroll right to see it.
4. Move the cursor to the first line, first column.
5. Open a search for the text and go to the second match.


nilnor:
Eventually we should probably cover this by (e.g.) extending Editor:ensure_visible to take a range instead. But this should do the trick until then, thanks!

PR #103, opened 2015-09-22
C/C++ lexer enhancements

kirbyfan64:
- Make highlighting of `true` and `false`, as well as types, consistent with other lexers.
- Highlight type definitions.


nilnor:
Thanks!

PR #108, opened 2015-09-29
Scroll down if needed after ctrl_d

kirbyfan64:


lib/howl/ui/editor.moon
```diff
@@ -443,7 +443,9 @@ class Editor extends PropertyObject
   duplicate_current: =>
     if @selection.empty
       line = @current_line
-      @buffer.lines\insert line.nr + 1, line.text
+      tline = line.nr + 1
```
nilnor: Looks good, but I can't figure out what `tline` would be short for? Rename that to something more fitting, e.q. `line_below` or `next_line` or inline it. Then merge at will :)

kirbyfan64: @nilnor It was supposed to be `target_line`... :/

PR #110, opened 2015-10-07
Line wrapping

nilnor:
This adds line wrapping support for aullar.  It supports the previously existing `line_wrapping` configuration variable, which supports wrapping on either character or word (or none). It also introduces two new configuration variables related to wrapping:

- `line_wrapping_navigation`

This provides control over how wrapped lines are navigated. "real" means cursor movement will step over the line in an ordinary fashion even if it's wrapped, and "visual" means that cursor movement will walk into a wrapped line. The mode is also reflected in the current line marker which will highlight the entire line for the former, and the visual line for the latter.

- `line_wrapping_symbol`

This is the symbol used for indicating wrapping at the right side of the wrapped visual line. The appearance of the symbol itself is controlled using ordinary styles.

The one (big) thing that is not present is "proper" scrolling of lines when the navigation mode is set to visual. We currently only support scrolling by real lines, meaning scrolling can now appear a bit jagged since scrolling one line might mean actually shifting the screen by two or more visual lines. The plan is not to include that in this PR however, but instead leave that for the  future.

### DIFF
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -176,6 +176,14 @@ class Editor extends PropertyObject
 
       @view.config.view_line_wrap = value
 
+  @property line_wrapping_navigation:
+    get: => @view.config.view_line_wrap_navigation
+    set: (value) =>
+      unless value\umatch(r'^(?:real|visual)$')
+        error "Unknown value for line_wrapping_navigation: #{value}", 2
+
+      @view.config.view_line_wrap_navigation = value
+
   @property cursor_line_highlighted:
     get: => @view.config.view_highlight_current_line
     set: (flag) =>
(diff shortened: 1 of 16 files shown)


lib/howl/ui/editor.moon
```diff
@@ -169,21 +169,20 @@ class Editor extends PropertyObject
...
-    get: =>
-      -- sci_val = @sci\get_wrap_mode!
-      -- switch sci_val
-      --   when Scintilla.SC_WRAP_NONE then 'none'
-      --   when Scintilla.SC_WRAP_WORD then 'word'
-      --   when Scintilla.SC_WRAP_CHAR then 'character'
-      --   else '(unknown)'
+    get: => @view.config.view_line_wrap
+    set: (value) =>
+      unless value\umatch(r'^(?:none|word|character)$')
```
kirbyfan64: Particular reason for the parens here? Seems slightly inconsistent with the rest of the file.

kirbyfan64:
So, while toying with this, I noticed that it's SUPER hard to select the very last character before the line wraps.

Still great though! :)

lib/howl/ui/editor.moon
nilnor: No reason, those just slunk in there (removed now).

nilnor:
@kirbyfan64 using the mouse presumably? I just tried that (for the first time apparently) and that is indeed not working as it should.

PR #111, opened 2015-10-13
Support for custom fonts and custom icon fonts

shalabhc:
Includes Font Awesome and some default fonts for file_selection.

Here's how the default looks:
file-selection-icons


lib/ljglibs/cdefs/fontconfig.moon
```diff
@@ -0,0 +1,9 @@
+-- Copyright 2015 The Howl Developers
```
nilnor: This probably shouldn't be under ljglibs, as it's not part of the g* family as far as I can see. I think `lib/howl/cdefs` would be a better match.

lib/howl/application.moon
```diff
@@ -359,6 +359,7 @@ class Application extends PropertyObject
     require 'howl.commands.ui_commands'
     require 'howl.commands.edit_commands'
     require 'howl.editing'
+    require 'howl.ui.icons.font_awesome'
```
nilnor: Should be moved down into `load_core`.

lib/howl/ui/icon.moon
```diff
@@ -0,0 +1,34 @@
+-- Copyright 2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+import style, StyledText from howl.ui
+
+icons = {}
+
+style_name = (icon_name) -> '_icon_font_'..icon_name
+
+define = (name, definition={}) ->
```
nilnor: Should we have a `define_default` here as well? Would tie in with the other registries.

nilnor:
I'm wondering whether we should have an option for disabling fonts?

@kirbyfan64 what's your opinion on using icons?

kirbyfan64:
@nilnor You mean the icons in the file selection?

shalabhc:
> I'm wondering whether we should have an option for disabling fonts?

Do you mean fonts or icons? I think an option for disabling each is reasonable. Disabling fonts would probably be outside the scope of this CL, though.

@kirbyfan64 - the file selection icons is probably what @nilnor meant. In any case, I'm interested in your opinion on those too. Also, what do you think of icons in general, e.g., to display buffer state (modified-in-buffer or modified-on-disk) and buffer type (file vs. process). This could be displayed in the buffer list or and also in the title bar.

kirbyfan64:
Well, the file icons look really cool. I think they can be used really well, but using them in the title bar for modified files might be kind of overkill...

lib/ljglibs/cdefs/fontconfig.moon
shalabhc: Ah that's where it was first before I decided it belongs with the g family since it's a dependency. :) 

I'll move it back.

shalabhc:
@kirbyfan64 - glad you liked the file icons. What if the buffer icons are in the `switch-buffer` list only? Right now you see a `*` if the buffer has been modified. BTW, I noticed that Atom shows the modified status next in the file title using a subtle blue dot.

lib/howl/application.moon
shalabhc: I don't follow - it is in `_load_core` already.

kirbyfan64:
@shalabhc `switch-buffer` list sounds great. I use it a *lot*, too!

lib/howl/application.moon
nilnor: You're right, not sure what happened there, please disregard.

lib/howl/ui/icon.moon
shalabhc: Done - added `define_default`.

lib/ljglibs/cdefs/fontconfig.moon
shalabhc: Done now - moved to howl/cdefs.

shalabhc:
@nilnor - I believe I've addressed your comments, except the one about having an option to disable icons. I'm not quite sure what to call this and what would be the scope? Would `config.file_icons = true|false` make sense? That would only apply to these icons for the file/directory names.

nilnor:
@shalabhc good question, but not one I have a good answer for. The alternatives would seem to be either a global option that disables all icons, or specific options for various areas. I'm not sure what makes sense. Just starting with your proposal seems like it could be a good idea, and can revise it as necessary.

kirbyfan64:
What about something like `config.show_icons_for` that's a list (table, in Lua lingo I guess :) of strings. The strings determine which icons show. For instance, right now the default could be something like `{'file'}`. If icons are added to show buffer types, then it could be `{'file', 'buffer_types'}`.

Or the opposite: `config.dont_show_icons_for`, which determines for what things there would be NO icons. This one is probably a bit more future-proof.

shalabhc:
@kirbyfan64 - I liked your suggestion. Thinking more about it I feel for something like buffer_types we may not want just true/false but something else ('icon', 'text', 'none' etc.) - which is not clear at the moment. So I've just implemented `config.file_icons` for now - we can adjust this in the future as concrete use cases emerge.

nilnor:
@shalabhc this PR is still waiting for the configuration option. You wrote that you have implemented `config.file_icons`, so I'm mentioning it in case you forgot to push.

shalabhc:
Doh! I did forget.. pushed now.

nilnor:
All right, merging! Could you update the Changelog when you get a chance?

PR #114, opened 2015-10-16
Faster

nilnor:
This add support for collapsing multiple edit operations into a single revision. It's restricted to the case where the size of the region to be changed is known in advance .
   
The is currently a big problem with doing a massive amount of updates for a buffer, is it causes a corresponding amounts of notifications to views. This get's very inefficient very fast. The new `change` method allows for a way around this, by allowing a range to be modified with all changes collected in one revision. In contrast to `as_one_undo`, only one notification will be fired to listeners.

Changed in this PR is buffer replace operations, as well as active lines transformations (e.g comment toggling, indenting, etc.).

### DIFF
--- a/lib/aullar/buffer.moon
+++ b/lib/aullar/buffer.moon
@@ -42,6 +42,32 @@ LineMt = {
       ffi_string line.ptr, line.size
 }
 
+change_sink = (start_offset, count) ->
+  {
+    roof: start_offset + count - 1
+    changes: {}
+    invalidate_offset: nil
+    :start_offset
+
+    add: (type, offset, size, invalidate_offset) =>
+      if offset < start_offset or (offset != start_offset and offset > @roof)
+        error "Out-of-range modification '#{type}' at #{offset} within change (#{start_offset} -> #{@roof})"
+
+      @invalidate_offset or= invalidate_offset
+      @invalidate_offset = min @invalidate_offset, invalidate_offset
+      @changes[#@changes + 1] = :type, :offset, :size
+
+      if type == 'deleted'
+        @roof -= size
+      else
+        @roof += size
+
+    can_reenter: (offset, _count) =>
+      return false if offset < start_offset
+      return false if offset + _count > max(start_offset + count, @roof)
+      true
+   }
+
 Buffer = {
   new: (text = '') =>
     @listeners = {}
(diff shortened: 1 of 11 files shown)


kirbyfan64:
This is really cool! I mean, Howl has always worked well with big buffers, so I can see this being even better.

Quick question: will this still call the complete function at every character entered?

shalabhc:
Very nice! I tried indenting ~1000 lines and it has dropped from about 7 seconds to sub second. :+1:

nilnor:
@kirbyfan64 this will not affect any functionality, and does not affect editing at all. But as an aside Howl is markedly faster for larger unicode buffers after the aullar introduction. 

To be fair, this change is a big improvement, but only because aullar introduced a performance regression for multi-update operations. I think that Scintilla probably optimized this case, so relative to 0.3 the difference is much smaller, to the point of being negligible. It's still faster though :)

kirbyfan64:
@nilnor I think aullar is actually faster for loading files, though. Cut off a second on loading Fbuild, which is about 16k LOC.

nilnor:
@kirbyfan64 Nice to hear! I'm not that surprised, as doing things ourselves means we can cut down on some indirection, but again nice to hear that you've experienced a noticeable speedup :)

kirbyfan64:
Pretty sure this may have caused some issues. After some experimentation with selecting stuff and deleting, sometimes only half the text is deleted, and the rest is deleted after the next few changes.

nilnor:
@kirbyfan64 this is not something I have encountered myself, which is not say that it doesn't happen of course, so it would be great if you could note the operations you did to provoke it so that it could be reproduced.

kirbyfan64:
@nilnor Yeah, I've been trying. All I know right now is that sometimes, if I select the next line and delete it, it's actually deleted, but sometimes it stays on screen. The only way to get it to disappear is reopening Howlm

shalabhc:
I saw some 'funny behavior' recently too. I deleted a line but the content below did not shift up and I saw a duplicate line. However, it seemed to fix itself line-by-line,  s I moved my cursor down IIRC. I haven't been able to reproduce this.

Is it possible to add a 'debug-dump' feature that dumps the undo history, command history, buffer contents and other data that could be used once we find ourselves in a broken state? The dumps might be useful in identifying and reproducing such bugs.

kirbyfan64:
@shalabhc #130

PR #116, opened 2015-10-16
Faster markers

nilnor:
The twin to the previous optimization PR. This makes the adding of aullar markers, and in extension Howl highlights a multi-op affair, meaning we cut down on useless invalidations and refreshes.

Also included in this is changing the ListWidget to use highlights instead of styles for marking search hits (see commit message for a more elaborate discussion of the rationale).

### DIFF
--- a/bundles/solarized-themes/light.moon
+++ b/bundles/solarized-themes/light.moon
@@ -144,6 +144,11 @@ return {
       background: blue
       background_alpha: 0.2
 
+    list_highlight:
+      type: highlight.UNDERLINE
+      text_color: black
+      line_width: 2
+
     cursor:
       type: highlight.RECTANGLE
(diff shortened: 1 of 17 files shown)


shalabhc:
:+1: , however it breaks the rendering a bit. See
image

There seems to be some bleeding of the original text?

nilnor:
@shalabhc good catch! Not actually related to this, but I've adjusted it now in this PR. Note that in the example above you'll still see a slight colour change between the `member`-styled sections and the highlighted sections as the the colours are actually slightly different.

shalabhc:
Thanks - that fixed it.

FWIW, I won't miss the ability to bold the selected item. Highlights seems like a better fit anyway since selection is a transient state.

PR #117, opened 2015-10-21
Add support for PEP 498 string interpolation

kirbyfan64:
This implements [Python PEP 498 string interpolation](https://www.python.org/dev/peps/pep-0498/). The syntax is more complicated than it looks... :O


kirbyfan64:
Wait, is my Git history clobbered? Damn! Hold on a sec...

PR #118, opened 2015-10-21
Implement PEP 498 string interpolation

kirbyfan64:
This implements [Python PEP 498 string interpolation](https://www.python.org/dev/peps/pep-0498/). [Again](https://github.com/howl-editor/howl/pull/117). The syntax is more complicated than it looks!

### DIFF
--- a/bundles/python/python_lexer.moon
+++ b/bundles/python/python_lexer.moon
@@ -1,6 +1,8 @@
 -- Copyright 2012-2015 The Howl Developers
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
+pairs = pairs
+
 howl.aux.lpeg_lexer ->
   c = capture
 
@@ -83,11 +85,49 @@ howl.aux.lpeg_lexer ->
 
   decorator = c 'preproc', P'@' * name * ('.' * name)^0
 
-  P {
+  f_prefix = c 'special', any {
+    S'rR' * S'fF'
+    S'fF' * S'rR'
+    S'fF'
+  }
+
+  format_conv = c('operator', P'!') * c('special', S'sra')
+  format_number = any {
+    c 'number', integer
+    c('operator', P'{') * ((V'all' + space + P 1) - P'}')^0 * c('operator', P'}')
+  }
+  format_part = any {
+    format_number
+    c 'special', (P 1) - S'{}'
+  }
+  format_spec = c('operator', ':') * format_part^0
+
+  string_kinds =
+    sq_string: "'"
+    dq_string: '"'
+    tsq_string: "'''"
+    tdq_string: '"""'
+  interpolations = {}
+  f_strings = {}
+  for kind, quote in pairs string_kinds
+    interpolations["#{kind}_interpolation"] = sequence {
+      ((V'all' + space + P 1) - (S'}:' + format_conv))^0
+      format_conv^-1
+      format_spec^-1
+      c 'operator', '}'
+      V"#{kind}_chunk"
+    }
+    interpolations["#{kind}_chunk"] = sequence {
+      c 'string', scan_to P(quote) + #P'{', P'\\'
+      V"#{kind}_interpolation"^0
+    }
+    f_strings[#f_strings+1] = c('string', quote) * V"#{kind}_chunk"
+
+  rules = {
     'all'
 
     all: any {
-      string,
+      V'string',
       comment,
       number,
       operator,
@@ -100,4 +140,14 @@ howl.aux.lpeg_lexer ->
       identifier,
       decorator
     }
+
+    f_string: any f_strings
+
+    string: any {
+      f_prefix * V'f_string'
+      string
+    }
   }
+
+  rules[k] = v for k, v in pairs interpolations
+  P rules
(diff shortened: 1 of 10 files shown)


nilnor:
@shalabhc is this something you have or could look at, given that you're more of a Python buff than yours truly?

shalabhc:
It would be great to have these format strings styled. I've looked at it briefly and will review more thoroughly. For now here are some issues:

f'{value:#format}' should not style '#format' as a comment.
f'{value:def}' should not style 'def' as keyword.
f'{value:>+#10.2)' is not styled correctly.

This seems to handle the built in format specification well. Handing everything properly is quite a challenge. Also, given that the language allows 'Each formattable type may define how the format specification is to be interpreted' (https://docs.python.org/2/library/string.html#formatspec) - I wonder if it's useful to handle the numeric format specifications at all? Note that the format spec can be absolutely any string (including my examples above). 

I'm thinking that anything after the ':' - i.e. the entire format spec - could be styled as a single style, maybe 'special'. While this would lose the special style for numeric format strings, it will considerably simplify the implementation and be more correct in the general sense. However, this will also lose the style for nested {expression} in the format spec. Maybe that is OK - what do you think @kirbyfan64?

kirbyfan64:
@shalabhc Hmm...you're probably right...I think I may be able to save the formatting of brackets, too. Hold on a sec...

kirbyfan64:
@shalabhc I pushed a new commit...but is it possible with LPeg to match any character? I tried replacing [this line](https://github.com/howl-editor/howl/commit/784f4455cae07d1167d658779608012479d79a9e#diff-851d9b84896270cc7e3bbea3014007a5R97) with:

```moonscript
format_spec = c('operator', ':') * (format_number + c 'special', P 1)^1
```

[like the LPeg docs for `P` say](http://www.inf.puc-rio.br/~roberto/lpeg/#op-p), but it didn't match anything.

shalabhc:
@kirbyfan64 - not sure. Sorry my knowledge of LPeg isn't great but I'll poke around a bit and see if I can figure this out.

shalabhc:
@kirbyfan64 - So I think instead of matching any char, it might be better to match one or more chars that are *not* `{` - what do you think? It might be easier to implement too.

kirbyfan64:
@shalabhc All issues addressed! I also made the formatting a bit more lenient to avoid string formatting getting screwed in the middle of typing.

kirbyfan64:
Bump?

shalabhc:
Sorry for the delay on this. I tried this out but in `f'{value:#format}'` I still get '#format' styled as a comment.

Some other implementation comments inline.

bundles/python/python_lexer.moon
```diff
@@ -100,4 +132,24 @@ howl.aux.lpeg_lexer ->
       identifier,
       decorator
     }
+
+    tsq_interpolation: gen_string_interpolation 'ts'
```
shalabhc: It would be nice to not have all variants of the `f` strings here. Instead just call these, say,  `f_string` as a single rule that is defined separately above using `any`. Also, style wise, it seems strange to use interpolation with 'ts' etc. to refer to other rules - it makes it somewhat hard to follow.  Would be possible to just generate all the variants of the f strings by iterating over the names of the normal strings? Maybe you could add the generated rules to a list and then make an `any` rule out of it?

kirbyfan64:
@shalabhc Are you sure? I just tested it, and it's styled correctly.

bundles/python/python_lexer.moon
kirbyfan64: @shalabhc I just updated it. Is it better now?

bundles/python/python_lexer.moon
```diff
@@ -100,4 +140,14 @@ howl.aux.lpeg_lexer ->
...
+
+    f_string: any f_strings
+
+    string: any {
+      f_prefix * V'f_string'
+      string
+    }
   }
+
+  rules[k] = v for k, v in pairs interpolations
```
shalabhc: I'm not sure why it is necessary to inject all interpolations into the main rule set? Since `f_string` is already in `all`, wouldn't that just work?

kirbyfan64: @shalabhc Because the rules recurse into each other with `V`. Just like in the MoonScript one.

shalabhc:
@kirbyfan64 - ah you're right `f'value:#format'` is fixed. However I found the following issues:

```python

f'{value:abc{f()}}' # not good, string not closed?

f'{val!#comment}' # not good, styled as comment, should be error or unstyled ?

f'{a + b}' # not good, string not closed?

```

kirbyfan64:
@shalabhc *sigh* I'm pretty sure I fixed all the issues this time. If all's OK now, I can squash the commits.

shalabhc:
@kirbyfan64 - seems to work pretty well but I found a special case ;)

```
f'{x!=1}'
```

I think the `!=` operator has special exception in the pep. If fixing this is too complex, I'm ok with leaving it as is.

kirbyfan64:
Crud, my rebase killed the commit history. Hold on...

PR #119, opened 2015-10-21
Os bindings

nilnor:
This adds support for OS specific bindings, by allowing for an optional `for_os` sub table in a key map. Included is a crude, incomplete and most likely incorrect set of base bindings for OS X.

### DIFF
--- a/lib/howl/bindings.moon
+++ b/lib/howl/bindings.moon
@@ -4,7 +4,7 @@
 _G = _G
 import table, coroutine, pairs from _G
 import tostring, pcall, callable, type, print, setmetatable, typeof from _G
-import signal, command from howl
+import signal, command, sys from howl
 append = table.insert
 
 signal.register 'key-press',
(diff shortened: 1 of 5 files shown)


nilnor:
So, no comments on this? I'll merge it soon then. FWIW, someone who actually has an OS X machine around would have to try this and check things out, and ideally work with fleshing out and adapt things. That's not going to be me though for the record.

shalabhc:
Please go ahead. I tried this out on OS X recently and the bindings work as intended. I'll take a shot at adapting them once it's in.

lib/howl/bindings.moon
```diff
@@ -211,12 +219,17 @@ export keystrokes_for = (handler, source = nil) ->
 
   keystrokes
 
-export action_for = (translation, source='editor') ->
+export action_for = (tr, source='editor') ->
+  os = sys.info.os
+  empty = {}
```
kirbyfan64: This use of `empty` feels odd to me. Can't the dict just be left as before like `or {}`?

nilnor: Well, it adds some descriptive value I find, but also I cringed a little bit at creating so many empty tables right after one another, when they needn't be different.

kirbyfan64: @nilnor That makes sense.

PR #120, opened 2015-10-23
Implement subtree mode for file interactions

shalabhc:
To try this out, press `ctrl_s` while running the open command to switch to subtree view. Press `ctrl_s` again to switch back.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Added subtree view to file listings.
+
 - Added custom font support and Font Awesome icons for file listings.
 
 - Added two new configuration variables for line wrapping:
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,8 @@
 
 ## Unreleased (in master)
 
-- Added subtree view to file listings.
+- Added recursive listing feature to file interactions. Pressing `ctrl_s` in the
+`open` command now toggles between recursive and regular list of files.
 
 - Added custom font support and Font Awesome icons for file listings.
 
--- a/lib/howl/interactions/file_selection.moon
+++ b/lib/howl/interactions/file_selection.moon
@@ -62,7 +62,7 @@ class FileSelector
     @directory = directory
     local matcher
     if @show_subtree
-      items, timed_out = self.subtree_reader(directory, timeout: 1)
+      items, timed_out = self.subtree_reader(directory, timeout: 3)
       matcher = subtree_matcher items, directory
       if timed_out
         @command_line.title = (@opts.title or 'File') .. ' (subtree, truncated)'
--- a/lib/howl/interactions/file_selection.moon
+++ b/lib/howl/interactions/file_selection.moon
@@ -62,8 +62,13 @@ class FileSelector
     @directory = directory
     local matcher
     if @show_subtree
-      matcher = subtree_matcher self.subtree_reader(directory), directory
-      @command_line.title = (@opts.title or 'File') .. ' (subtree)'
+      items, timed_out = self.subtree_reader(directory, timeout: 1)
+      matcher = subtree_matcher items, directory
+      if timed_out
+        @command_line.title = (@opts.title or 'File') .. ' (subtree, truncated)'
+        log.warn 'Too many files in subtree - truncated.'
+      else
+        @command_line.title = (@opts.title or 'File') .. ' (subtree)'
     else
       matcher = file_matcher self.directory_reader(directory), directory, @opts.allow_new
       @command_line.title = @opts.title or 'File'
@@ -150,10 +155,12 @@ interact.register
         append dirs, 1, directory
         return dirs
 
-      .subtree_reader = (directory) ->
-        dirs = subtree_reader directory, filter: (file) -> not file.is_directory
+      .subtree_reader = (directory, opts={}) ->
+        opts = moon.copy opts
+        opts.filter = (file) -> not file.is_directory
+        dirs, timed_out = subtree_reader directory, filter: (file) -> not file.is_directory
         append dirs, 1, directory
-        return dirs
+        return dirs, timed_out
 
       .title or= 'Directory'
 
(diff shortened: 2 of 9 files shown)


nilnor:
I will have a more thorough look at the code later, but so far:

- I tried it out, and it works great. The icons are again really nice I find
- "Working with files" is one of the sections of the manual, so hint hint
- Like we just talked about, we need to find a way of making this option discoverable. People won't find it otherwise, and it would be a shame :)

nilnor:
Oh, and another thought I had was how this ties in with `project-open`. They're separate commands, but that and the sub tree open are now similar enough that if feels kind of weird for them to work differently with regards to how we display things (i.e. icons, show directories or not). Not really sure what I think about that yet though, but perhaps you are?

shalabhc:
In terms of icons they're the same as `project-open` now shows icons. However it doesn't show directories - I didn't think too much about it, I just left the current behavior intact. Should I change it to show directories for consistency?

kirbyfan64:
There should be some way of handling huge directories. Maybe some sort of "delayed loading", where the rest of the list will be loaded once you scroll near it?

If you're daring, try going to your home directory and then opening the subtree view. You can then watch as Howl completely freezes and uses more and more memory... :D

shalabhc:
@kirbyfan64 - excellent point and something I've thought about too. Maybe a simple fix is to limit the number of files (or time) during recursive traversal? I doubt this feature is useful for finding a file among 10s of 1000s so we could just limit either time spent or files listed and display a message to that effect. What do you think?

kirbyfan64:
@shalabhc Sounds good!

shalabhc:
@kirbyfan64 - I've commited and pushed the timeout feature. Please let me know how it works out if you give it a shot.

@nilnor 
- Regarding `project-open`, I realized it's not straightforward to show directories. Also while the subtree list displays everything it can find in the subtree, `project-open` may defer to `@vc\files!` and I'm not sure what that returns. IOW, I think leaving `project-open` for files only is OK IMO as it seems in some cases it may be a subset of the files in the subtree.
- I've added a section to the manual for this feature.
- Regarding discovery, yes that should be fixed. I'd like to do that as a separate PR since it's a new *infobar* feature and spans multiple commands.

In summary, I believe I've addressed all reported issues :smiley:

kirbyfan64:
+1

Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Added subtree view to file listings.
```
nilnor: `subtree` feels like an implementation name, rather than a user one (at least it's not readily recognizable to me as generic term for this anyway). I would think of this more as a "recursive" listing, e.g. as used in `ls`. What do you think? We should change this in the UI as well if/when we change it.

lib/howl/interactions/file_selection.moon
```diff
@@ -67,7 +60,18 @@ class FileSelector
 
   _chdir: (directory, text) =>
     @directory = directory
-    matcher = file_matcher self.directory_reader(directory), directory, @opts.allow_new
+    local matcher
+    if @show_subtree
+      items, timed_out = self.subtree_reader(directory, timeout: 1)
```
nilnor: 1 second feels like a tight timeout to me, when switching to a recursive listing I would be willing to wait longer than getting truncated listings. This would be subjective of course, but how about 2 or 3 seconds?

lib/howl/io/file.moon
```diff
@@ -173,7 +173,12 @@ class File extends PropertyObject
     files = {}
     directories = {}
     dir = self
-    while dir
+
+    local deadline
+    if options.timeout
+      deadline = glib.get_monotonic_time! + (1000 * 1000 * options.timeout)
+
+    while dir and not timed_out
```
nilnor: `timed_out` is not defined here.

nilnor:
Re `project-open` we can leave that as is for now. Yes, it's not straight forward as you'll be working on a filtered list, and the directories would have to be deduced. I think we'll have to unify this at some point in the future though, as it's not clear exactly why these type of listings work differently from a user perspective. Why can't I (as a naive user) switch back and forth from recursive and ordinary listing when I use the `project-open` command instead of the `open` command? 

The files returned from a VC instance would typically be filtered from an ignore list (e.g. ".gitignore"). For a recursive listing I think it would be OK to always filter based on VC preferences if you happen to be in a VC root, and it could considered be a nice feature at that. And if we do that, we could make the `project-open` command an ordinary recursive listing command from project's root.

nilnor:
All right, all done! :)

lib/howl/interactions/file_selection.moon
kirbyfan64: Maybe the timeout can be a config option?

Changelog.md
shalabhc: Changed to 'recursive' in the messaging. Also updated the Changelog and added more detail.

lib/howl/interactions/file_selection.moon
shalabhc: Updated to 3s for now. Eventually it would be nice to show something after 1s (e.g. 'please wait..') so the user is not left wondering, but that would have to be implemented across all interactions.

We can make this a config option if needed later.

lib/howl/io/file.moon
shalabhc: Deleted leftover code.

shalabhc:
Yes some unification with `project-open` makes sense down the road.

lib/howl/interactions/file_selection.moon
```diff
@@ -108,21 +112,20 @@ class FileSelector
   keymap:
     enter: =>
       app.editor\cancel_preview!
+      file = @list_widget.selection and @list_widget.selection.file
       name = @list_widget.selection and @list_widget.selection.name
-      if not @allow_new and not name
-        log.error 'Invalid path'
+      if not @opts.allow_new and (not file or not file.exists)
+        log.error 'Invalid path: '..tostring(file)
```
nilnor: Both works, but I thought I'd point out that this could also be the shorter `"Invalid path: #{file}".

PR #124, opened 2015-10-24
Incremental bytecode builds

kirbyfan64:
The actual build time is slower, but now builds are incremental, so you don't have to wait for all 400 source files to compile when you edit one of them! :D

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -29,6 +29,10 @@ else
 endif
 OBJECTS = main.o process_helpers.o
 DEP_OBJECTS = $(LPEG_OBJECT)
+LUASRC := $(shell find ../lib ../bundles -name '*.lua')
+LUABC := $(patsubst %.lua,%.bc,$(LUASRC))
+MOONSRC := $(shell find ../lib ../bundles -name '*.moon')
+MOONBC := $(patsubst %.moon,%.bc,$(MOONSRC))
 
 all: howl bytecode
 
@@ -63,9 +67,15 @@ deps-clean:
 clean:
 	-rm -f ${OBJECTS}
 
-bytecode: howl
-	-@find ../lib ../bundles -name '*.bc' | xargs rm
-	@find ../lib ../bundles -name '*.lua' -o -name '*.moon' | xargs ./howl --compile
+bytecode: howl $(LUABC) $(MOONBC)
+
+$(LUABC): %.bc: %.lua howl
+	rm -f $@
+	./howl --compile $<
+
+$(MOONBC): %.bc: %.moon howl
+	rm -f $@
+	./howl --compile $<
 
 install: all
 	@echo Installing to $(DESTDIR)$(PREFIX)..


nilnor:
I'm not convinced about the need for this. Building it incrementally is generally nice of course, but how often do you really need make the byte code manually? I typically do it when I switch branches or have fetched new stuff, and for those scenarios I don't find it bothersome that all files are recompiled as it's fast anyway. For general development the built-in auto compling works well.

So given that the Makefile becomes more complicated, that the build time for a clean build increases (significantly from what I saw), and that the problem it solves does not seem that significant I'm afraid I'll say no to this. You'll have to provide some good arguments here sway me :P

kirbyfan64:
Well, the logic is developing Howl. For instance, when working on lexers, I'll often modify *one* file, only to wait a few secs for *all* of them to rebuild. This will instead rebuild the one file that actually needs to be. In addition, clean builds are pretty uncommon when compared to incremental ones.

If you want a bit more speed, try `make -j`.

nilnor:
Well, that's what the built in auto compilation handles for you - if you edit the file in Howl and save it Howl will automatically compile the bytecode for you for that file. So there's no need to make bytecode manually after an edit. We should probably add a new section in the documentation for Howl that covers development of Howl itself, as this for instance is not well advertised.

kirbyfan64:
> if you edit the file in Howl and save it Howl will automatically compile the bytecode for you for that file

...so, all this time, I've been running `make` for absolutely NO reason?? And I've been using Howl for over a year?? :O

shalabhc:
You're not alone, @kirbyfan64, that was my reaction as well.

"@nilnor - *now* you tell us!?" :smirk: 

Early on I had at least one incident (quite possibly more than one) where it appeared that a modification did not get picked up on restart, and I've been running make after edits ever since...

nilnor:
Oh my, I'm sorry to hear this, that sounds horrible! I'm afraid this has been something I've always taken for granted. Well, best get on with updating the manual then to include this and prevent similar misfortunes in the future.

@shalabhc the logic had a slight defect at one point, in that it left the old byte code handing around if a new one couldn't be produced (.e.g. due to syntax errors in the new file). This could give the impression that things weren't reloaded, so perhaps that's what you saw.

shalabhc:
@nilnor - Didn't see any smileys in your reply so just wanted to make sure my comment wasn't taken more seriously than intended. While I did run `make` all this time, my giving you a hard time was entirely tongue-in-cheek fun :stuck_out_tongue_winking_eye:

FWIW, it seems I do need to run make for my changes to be picked up. Not sure why but I'll look at it at some point.

nilnor:
@shalabhc No worries, although it actually sounds horrible :P 

Have a look at `Application._on_buffer_saved` for troubleshootng, perhaps you have some other on save handler that interferes?

ozra:
Eh? It doesn't seem to be automatically compiled here. And rebuilding all the bytecode from clean takes forever on my haul-around machine. I just use my `watch-and-do` script to watch the dir and run make on change. Close tested howl instance & re-run `howl &` - since I want it run from terminal to catch possible error messages on stderr. I could as well do `make install && howl &` instead of watch since I'm running from term anyway, but..

How does the inbuilt compilation work? Does it have to be enabled? Does it live-reload too just on a chance it'd work?

kirbyfan64:
Or maybe the other way around (`make incremental`)?

ozra:
Yeah sure - as long as it's available :-)

ozra:
@kirbyfan64 - would you mind reworking it as a "make incremental"? I'd love to have this in the head makefile _soooon_ :-) It's such an annoyance when the build isn't working properly, keeps me from wanna do PR's :-(

nilnor:
So, I still believe that this is not necessary. But I also see that there's a lack of documentation here - it always worked for me so I didn't think about it. Anyway, I finally got around to start writing some development documentation, and also made the byte code updating more flexible for different run scenarios. Have a look here and see whether this helps: [Dev manual](https://howl.io/doc/manual/dev-howl.html)

ozra:
**A couple of thoughts:**

When running `src/howl` instead of `howl` it did recompile to .bc automatically. **Thanks for the new docs! :-)**

But...

- I don't get error messages (other than from the "live checks") _nor a confirmation of success_, or rather: compilation taking place at all. I really prefer having output in a terminal which I can stream through whatever filters I want, trig whatever notification methods I want, etc (I'm a Linux practician to the core, the only OS I've _worked daily with_ other than that was Amiga OS - long time ago)
   - I'm not inherently _against_ building from within an editor, however there is still _a lot of work to do on Howl before I'd be prepared to "go emacs" on workflows with it..._ Until then it's just my _favourite **editor**_.
   - The mere fact of relying on the subject as the build-chain for itself (_see earlier comment_).
   - Importantly: **I don't trust myself!** ;-) - "did I edit it in the session which was started from `repo/src/howl`?" - **uncertainty and doubt** - (_Goes off in terminal doing `ls lib/some/where/file.*` to compare mtime by eye again_)

**Sidenotes on build chain in general:**
- the Makefile really should be in repo _root_ - no? (It doesn't limit its' operations to the src-dir). Also important: _it's a (validly) common expectation_.
- "howl" should be compiled to "bin/" like the other targets, no? Back in the days I also compiled objs and bins to proj-root _and_ kept .c/.h files there - but I really think that's _bad_ twenty-oopmphf-fuckin-years old-school (there's _some good_ old-school idioms applicable today, but personally I spare that for real-time SBC coding). Repo integrity gains from sources and generated derivatives being separated as well as possible.
- Even using the editor-bin as compiler (`howl --compile`) is a bit shady to me ;-) However: since it has the pro of being helpful in ensuring same-byte-code-version, it's fine; _iff_ it could give as clean error output as `moonc`, instead of also outputting an error trace of it's own stack (_the compile failed - not howl_).

_Just my 2 öre/cents/satoshi/mBTC and/or all that._

**A bit terser put:**
- I can't see how an additional incremental target would be a problem to anyone _not using it_.
- On the other hand, it has a clear advantage; to me, and likely others.

**And finally:**

- Long live Howl! :-)

ozra:
Immediate edit: Most of my reactions are based on _my habits_ (duh!), and _the fact that I'm new to the Howl code base_ / workflow. And, that's it: there will be thousands and thousand like me to come. Trust me: this is the shit. Howl's an unpolished diamond. But a diamond it is. And there _will_ be more folks, and there _will_ be more requests for the expected build flow to work as most folks are habituated too. Smooth work flow for us dumb assess masses = more PR's = faster Howl kicks mega-ass :-)

PR #125, opened 2015-10-26
Changed buffer title generation logic

shalabhc:
- removed numeric suffix
- only file names are disambiguated
- project name is used for disambiguation if applicable
- at most one parent directory name used for disambiguation
- titles may not longer be unique (this is ok)
- the title for a buffer may automatically change when a new file
  with the same base name is opened

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -1,7 +1,7 @@
 -- Copyright 2012-2015 The Howl Developers
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
-import BufferContext, BufferLines, BufferMarkers, Chunk, config, signal from howl
+import BufferContext, BufferLines, BufferMarkers, Chunk, Project, config, signal from howl
 import File from howl.io
 import style from howl.ui
 import PropertyObject from howl.aux.moon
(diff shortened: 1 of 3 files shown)


kirbyfan64:
> - at most one parent directory name used for disambiguation

Is there a particular reason for this change? I know that, when working on any modestly-sized Python project, this will completely backfire due to the abundance of `__init__.py` files. For instance, Fbuild has `builders/c/__init__.py` and `config/c/__init__.py`, which both show up as `c/__init__.py`.

shalabhc:
Note that the full path is also visible in the buffer list so these will not be entirely ambiguous. Also, if you had only two files open currently, they would be `__init__.py` and `c/__init__.py` which isn't any better I suppose.

While there is no absolute reason to limiting it to one parent, it followed from the following:
- Titles don't need to be unique because the full path is also shown in the list.
- I wanted titles to be consistent across all buffers with the same basename (e.g. if you have one 'parent1/foo' you will always see 'parent2/foo' and never a bare 'foo').
- It's preferable to limit the length of the title, just for aesthetics.

I assumed that one parent will be sufficient for disambiguation in almost all cases and in the remaining rare cases one can look at the full path. But I'm open to change. Do you think a hard limit of depth=2 would make more sense or do we need an unbounded prefix?

kirbyfan64:
> Note that the full path is also visible in the buffer list so these will not be entirely ambiguous.

In my testing, they were both listed as `c/__init__.py`.

shalabhc:
What I meant is in the second column.

lib/howl/buffer.moon
```diff
@@ -14,22 +14,46 @@ ffi = require 'ffi'
...
-    return title if not file
-    title = file.basename .. File.separator .. title
-
-  title
-
-title_counter = (title) ->
-  title_counters[title] = 1 if not title_counters[title]
-  title_counters[title] += 1
-  title_counters[title]
+file_references = {}
```
nilnor: I don't see how this gets cleaned up. The ref count handling does nothing to remove entries e.g. when a buffer is closed.

spec/buffer_spec.moon
```diff
@@ -551,12 +551,8 @@ describe 'Buffer', ->
...
           assert.equal b2.title, 'sub2' .. File.separator .. 'file.foo'
 
-          sub_sub = sub1\join('sub2')
-          sub_sub\mkdir!
-          f3 = sub_sub\join('file.foo')
-          f3\touch!
-          b3.file = f3
-          assert.equal b3.title, 'sub1' .. File.separator .. b2.title
+      it 'uses the project name suffix as needed', ->
+
```
nilnor: This is missing.

spec/buffer_spec.moon
```diff
@@ -551,12 +551,8 @@ describe 'Buffer', ->
           b2.file = f2
           assert.equal b2.title, 'sub2' .. File.separator .. 'file.foo'
 
-          sub_sub = sub1\join('sub2')
```
nilnor: If you remove this then the `b3` above should be removed as well.

nilnor:
The approach taken here seems very heavy handed to me. Intuitively it feels like what should happens is that when we assign a title, we check whether this would clash with an existing title. If so, we then need to resolve the titles of the affected buffers, which would typically be a very small number of buffers. This PR resolve all the titles all the time, regardless of whether they need to be resolved or not, and it does so in a rather brute force fasion.

In short, this needs a rework. Related to this, I'm wondering whether this logic should perhaps be somewhere else, and be controlled by `Application` instead.

shalabhc:
@nilnor - I think you have valid points. Deleting a buffer isn't handled at all and it may be best to move this logic elsewhere. I'll redo this as a new PR. I might move the logic into a new module called buffer_list so that adding and removing from the list can be tracked. There's still the problem of tracking change to the file associated with a buffer.

Issue #113, opened 2015-10-13
Enhancements to find and replace


kirbyfan64:
So, this is a random idea I've had. Basically:

1. Allow using `ctrl_f` and `ctrl_b` in `buffer-replace` like in `buffer-search-forward`. That way, it's easier to see all matches that will be replaced. Right now, I usually just scroll up and down like crazy and try to find any selected text...which isn't that efficient. :)

2. Add companions to `buffer-replace-regex`: `buffer-search-forward-regex` and `buffer-search-backward-regex` (bound to `ctrl_shift_f` and `ctrl_shift_b`, respectively). Currently, I use `buffer-replace-regex` to search for regexes and again scroll around to find highlighted text. Again not the most efficient thing around...

3. (This one's really debatable.) If any text is selected when doing `buffer-replace*`, replace occurrences found ONLY in the region of selected text. I can see this again being useful for doing things like variable name replacements.

kirbyfan64:
FWIW, I know about `buffer-search-word-*`, but it seems weird for `buffer-replace*` to have regex support but not `buffer-search*`.

shalabhc:
Thanks for your suggestions!

For (1), you can use the `up` and `down` keys. Binding the same behavior also to `ctrl_f` and `ctrl_b` seems reasonable.

You make good points in (2) - I've thought about something like this also, basically to get various types of searches (incremental vs. global) for all types of syntax (regex, exact, boundary, word etc.). OTOH, this will cause an explosion of search commands so I'm divided on this. Btw, an undocumented feature of `buffer-replace` is that you can press `enter` after just typing the match regex (without typing the '/') - this jumps to the highlighted match without making any replacements. This is more useful in combination with the `up` and `down` keys, of course.

(3) is already the current behavior, if I understand you correctly. Replacements are applied only in the selected area. We should probably add highlighting in the preview for this though.

shalabhc:
After pondering this some more, I feel a specialized regex search definitely makes sense. The problem is figuring out the right behavior. Current commands are:

Command|Matching|Region|Direction|List widget
-------|--------|------|---------|----
buffer-grep | exact + boundary | entire file | no | yes
buffer-search-forward / backward | exact | visible region | yes | no
buffer-search-word-forward / backward | exact word | visible region | yes | no

Other things to note:
* Special case: `*search-word*` commands automatically copy the word at cursor.
* There is no search that uses exact matching and displays a list
* The visible region search commands do not show a summary either (e.g. number of matches)
[...]

PR #127, opened 2015-11-05
Added buffer-grep-exact and buffer-grep-regex commands

shalabhc:
Here it is as discussed in #113. As a bonus, I've added `buffer-grep-exact`.

### DIFF
--- a/lib/howl/interactions/line_selection.moon
+++ b/lib/howl/interactions/line_selection.moon
@@ -5,7 +5,7 @@ import interact from howl
 import highlight from howl.ui
 import Matcher from howl.util
 
-line_match_highlighter = (editor) ->
+line_match_highlighter = (editor, explain) ->
   local buffer
   (selection, text, items) ->
     if buffer
@@ -18,7 +18,7 @@ line_match_highlighter = (editor) ->
     line = buffer.lines[selection.line_nr]
     local segments
     if text and not text.is_empty
-      segments = Matcher.explain text, line.text
+      segments = explain text, line.text
 
     if segments
       start_pos = line.start_pos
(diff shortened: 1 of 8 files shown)


kirbyfan64:
Would it be possible to have a key binding for `buffer-grep-regex`, or at least the ability to hit `ctrl_r` in `buffer-grep` to switch to regex mode like the find and replace has?

shalabhc:
@kirbyfan64 - good idea. I'll add a `ctrl_r` binding within `buffer-grep`.

lib/howl/interactions/line_selection.moon
```diff
@@ -49,13 +49,24 @@ line_match_highlighter = (editor) ->
...
-          segments = Matcher.explain text, line.text
+          segments = explain text, line.text
           if segments
             for segment in *segments
               ranges[#ranges + 1] = { start_pos + segment[1] - 1, segment[2] }
 
         highlight.apply 'search_secondary', buffer, ranges
 
+create_matcher = (find) ->
+  class CustomMatcher
```
nilnor: Not a big fan on the in-line class definition here. There seems to be need for it as well? The class could be moved up to top-level, and take both `line_items` and `find` as constructor arguments.

nilnor:
Looks good apart from the one note. I also would like the ability to switch between the modes :)

lib/howl/interactions/line_selection.moon
shalabhc: I don't like this much either but `Matcher.explain` is a class method (not instance method) and so it can't use `self.find`.

nilnor: Ah, I see. In that case I propose a more lightweight solution by returning a table with the `explain`, with a metatable set containing the `__call`.

shalabhc: Done.

shalabhc:
It's ready again - remove inline class and added ability to switch between `buffer-grep*` commands.

nilnor:
All right, looks good! Let's merge this.

I'll add a comment though - I think it would be nice with one generic key binding for switching between different related interactions, such that that binding would cycle through the available modes, regardless of what interaction I'm in. Otherwise having to remember that it's `ctrl_e`, or `ctrl_r`, `ctrl_g` or `ctrl_s` for that other interaction will probably prove to much at least for yours truly. Thoughts?

shalabhc:
In fact I considered 'cycling' behavior and liked it just as much, possibly more. The only reason I went with this is consistency with `buffer-replace` -> `[ctrl_r]` -> `buffer-replace-regex`. But perhaps that should be a cyclic shortcut too?

Should there be a generic binding for switching between command variants or perhaps we just use the main shortcut (`ctrl_h` for replace, `ctrl_g` for grep) for cycling through the respective variants? Your fingers are already in the right position, making repeated invocation quick and easy.

nilnor:
Yes, I think that all interactions should use cyclic switching in that case. I like the idea of just using the main short-cut, that make sense.

PR #128, opened 2015-11-10
Try the new Travis container infrastructure

kirbyfan64:
This should result in faster builds. For more info, see http://docs.travis-ci.com/user/workers/container-based-infrastructure/.


nilnor:
Thanks!

PR #129, opened 2015-11-10
Memory handling

nilnor:
The main focus of this has been to go through all kinds of memory and ref handling to verify that resources are released correctly, and to re-enable the last pending Editor GC spec after the aullar rewrite. Contains fixes for two bigger leaks  - the aforementioned editor spec and also an issue with popups not being released that I found. It also configures the GC to be more aggressive, which prevents RSS from ballooning up.

Also contains a fix for a "bad callback" case, a fix for the view spec which failed on newer Gtk versions, and some various small tweakings.

### DIFF
--- a/.agignore
+++ b/.agignore
@@ -1,3 +1,4 @@
 site/source/versions
 site/**/*.html
 site/**/*.erb
+src/deps/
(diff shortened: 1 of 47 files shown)


nilnor:
I'll add some stuff related to this later, such as a new stand-alone memory stress test I've created and possibly other related changes. That will probably come after this is in master though.

Issue #122, opened 2015-10-23
goto-line command


xcombelle:
I realized this goto-line command. feel free to integrate with current code
If you prefer I realize a proper pull request says so.

```moonscript
-- code licensed under same license than howl (MIT)

import app, command,interact from howl

command.register
  name: 'goto-line'
  description: 'Goto specified line'
  input: () ->
    line_str = interact.read_text prompt:""
    return tonumber(line_str)

  handler: (line_no) ->
    editor = app.editor
    if line_no < 1
      line_no = 1
    if line_no > #editor.buffer.lines
      line_no = #editor.buffer.lines
    editor.cursor.line = line_no
    editor.cursor.column_index = 1
    editor.line_at_top = line_no

howl.bindings.push {
  editor: {
      alt_g: 'goto-line'
  }
}
```

by the way it is very nice software (and not too hard to customize for its pre 1.0 status)

kirbyfan64:
I had suggested this feature in https://github.com/howl-editor/howl/issues/112#issuecomment-148055551, except bound to `ctrl_g`.

nilnor:
It seem there's a pent-up need for a goto line command, so let's add it :) I actually use the same thing quite often, but as I use the VI emulation it's already built in there, so I do see a need for it. 

@kirbyfan64: `ctrl_g` is already bound to `buffer-grep`, so something other than will be needed unless we want to break the current mapping, but at least I am well used to `ctrl_g` meaning grep by now.

@xcombelle if you want to open a PR, then please go ahead! Some initial feedback:

- The error checking for line number is nice, but is already handled deeper down and so not necessary
-  There's a new (not yet documented) method `move_to` method for the cursor now, so the moving could be done as `cursor\move_to line: line_no, column: 1`
- I personally don't think that always scrolling so that the specified line as at the top makes sense - if it's in view then I prefer the cursor to just go there. In the future we might flash the line to indicate where it is. Others will have to chime in here though.

Finally, nice to hear that you had a good experience customizing it. If there's anything in particular you think could make it easier still then let us know :)

shalabhc:
As I mentioned in #112 I use `buffer-grep` as a 'goto line' command (with preview!) and it works quite well.. However, a dedicated command for this makes sense.

@kirbyfan64 - `ctrl_g` for `buffer-grep` is one of my most used shortcuts and I would prefer to not change that binding.

@nilnor:
[...]

PR #131, opened 2015-11-16
Add alt_g

kirbyfan64:
See #122. I just went ahead and implemented it, since there was no response from @xcombelle.

### DIFF
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -134,3 +134,12 @@ command.register
     buffer = app.editor.buffer
     buffer.mode = selected_mode
     log.info "Forced mode '#{selected_mode.name}' for buffer '#{buffer}'"
+
+command.register
+  name: 'goto-line'
+  description: 'Go to the specified line'
+  input: () ->
+    line_str = interact.read_text title: 'Go to line'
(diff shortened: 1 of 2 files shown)


nilnor:
I made some comments for #122 regarding the implementation, which still apply.

kirbyfan64:
@nilnor Fixed.

nilnor:
All right, looks good! One question though, should we a working binding for OS X, since `alt` won't work there?

nilnor:
Well, `ctrl_g` is still taken (`buffer-grep`), so perhaps `ctrl_shift_g`?

kirbyfan64:
@nilnor Wait, shouldn't `buffer-grep` be `command_g` on a Mac, since that's the "primary key"?

nilnor:
Right you are (and it is), so that does free up `ctrl_g` for OS X. I can't say I know whether it's suitable or not, but it works for me. @shalabhc, any preferences?

shalabhc:
My preference agrees - `ctrl_g` for `goto-line` and `command_g` for `buffer-grep` sounds good.

kirbyfan64:
@howl-editor/developers How's this?

nilnor:
That's a wrap :)

PR #132, opened 2015-11-16
Implemented 'open-recent'

shalabhc:
Shows a list of recently closed files. The list persists across restarts.

### DIFF
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -27,6 +27,7 @@ class Application extends PropertyObject
     @windows = {}
     @_editors = {}
     @_buffers = {}
+    @_recently_closed = {}
     bundle.dirs = { @root_dir / 'bundles' }
     @_load_base!
     bindings.push keymap
@@ -38,6 +39,8 @@ class Application extends PropertyObject
     buffers = { table.unpack @_buffers }
     sort_buffers buffers
     buffers
+
+  @property recently_closed: get: => moon.copy @_recently_closed
 
   @property editors: get: => @_editors
 
@@ -116,6 +119,13 @@ class Application extends PropertyObject
 
     @_buffers = [b for b in *@_buffers when b != buffer]
 
+    if buffer.file
+      @_recently_closed = [file_info for file_info in *@_recently_closed when file_info.file != buffer.file]
+      append @_recently_closed, {
+        file: buffer.file
+        last_shown: buffer.last_shown
+      }
+
     if buffer.showing
       for editor in *@editors
         if editor.buffer == buffer
@@ -140,6 +150,7 @@ class Application extends PropertyObject
       log.error "Failed to open #{file}: #{err}"
       nil
     else
+      @_recently_closed = [file_info for file_info in *@_recently_closed when file_info.file != buffer.file]
       signal.emit 'file-opened', :file, :buffer
       buffer, editor
 
@@ -212,13 +223,16 @@ class Application extends PropertyObject
 
   save_session: =>
     session = {
-      version: 1
+      version: 2
       buffers: {}
+      recently_closed: {}
       window: {
         maximized: @window.maximized
         fullscreen: @window.fullscreen
       }
     }
+
+    open_files = {}
 
     for b in *@buffers
       continue unless b.file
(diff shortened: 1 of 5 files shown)


shalabhc:
@nilnor - correct. Limiting the list to some size is probably a good idea. What do you mean by pruning here - manual deletion of history or automatic?

nilnor:
Sorry, I was unclear. I believe I was thinking along the lines that it could either a hard limit, or it could be limited by how long ago the file was opened (i.e. prune entries older than X). Anyhow, I think a simple size limit will be the best.

shalabhc:
OK, added size limit - ready for review!

lib/howl/application.moon
```diff
@@ -4,13 +4,20 @@
...
 import Window, Editor, theme from howl.ui
-import Buffer, Settings, mode, bundle, bindings, keymap, signal, interact, timer, clipboard from howl
+import Buffer, Settings, mode, bundle, bindings, keymap, signal, interact, timer, clipboard, config from howl
 import File, Process from howl.io
 import PropertyObject from howl.aux.moon
 Gtk = require 'ljglibs.gtk'
 
 append = table.insert
 
+config.define
```
nilnor: As a style nit-picking I'd prefer if this was moved down at the end of the file instead, together with the signal definitions.

lib/howl/application.moon
```diff
@@ -116,6 +126,18 @@ class Application extends PropertyObject
...
     @_buffers = [b for b in *@_buffers when b != buffer]
 
+    if buffer.file
+      @_recently_closed = [file_info for file_info in *@_recently_closed when file_info.file != buffer.file]
+      append @_recently_closed, {
+        file: buffer.file
+        last_shown: buffer.last_shown
+      }
+      count = #@_recently_closed
+      limit = howl.config.recently_closed_limit
```
nilnor: `config` is already in scope here, so need for `howl.config`.

lib/howl/application.moon
```diff
@@ -116,6 +126,18 @@ class Application extends PropertyObject
 
     @_buffers = [b for b in *@_buffers when b != buffer]
 
+    if buffer.file
+      @_recently_closed = [file_info for file_info in *@_recently_closed when file_info.file != buffer.file]
```
nilnor: This block could be its own private method instead, as to not clutter `close_buffer` to much.

lib/howl/application.moon
```diff
@@ -311,7 +341,7 @@ class Application extends PropertyObject
   _restore_session: (window, restore_buffers) =>
     session = @settings\load_system 'session'
 
-    if session and session.version == 1
+    if session and session.version >= 1
```
nilnor: I wonder whether we should skip the version check completely for now, as we don't have any breaking incompatible changes yet. We could just check for the existence of `session.recently_closed` below. As it is, the version check seems wrong as it now accepts any future version of the session.

lib/howl/keymap.moon
```diff
@@ -77,6 +77,7 @@
   }
 
   ctrl_o:           'open'
+  ctrl_shift_o:     'open-recent'
```
nilnor: We should have a binding for OS X as well I guess?

lib/howl/commands/file_commands.moon
```diff
@@ -39,6 +39,25 @@ command.register
...
 
+command.register
+  name: 'open-recent',
+  description: 'Open a recently closed file'
+  input: ->
+    recent_files = {}
+    for file_info in *app.recently_closed
+      table.insert recent_files, {
+        file_info.file.basename
+        file_info.file.parent.path
```
nilnor: Use `.short_path` here for contracting the home dir.

nilnor:
All done!

I'll address your idea of changing the semantics to just keep track of all the recently opened files brought up in #134 here - I think it's a good idea. It makes the implementation simpler as well, which is good. The only question I can think of there is what the command should be named, as it's a combo of switch and open.

shalabhc:
I'm not sure it makes the implementation simpler. I would just leave `app.recently_closed` as is but use both `app.buffers` and `app.recently_closed` in the command itself.

As for naming - I don't see a problem with `open-recent` as one could think of it as opening a file rather than a buffer. This is similar to selecting an already open file in the `open` command.

shalabhc:
All done - ready for review again!

nilnor:
Merged!

lib/howl/commands/file_commands.moon
```diff
@@ -39,6 +39,31 @@ command.register
   input: -> interact.select_file allow_new: true
   handler: (file) -> app\open_file file
 
+command.register
```
shalabhc: @nilnor - this branch was merged but the latest change to the command didn't get in? Specifically the buffers are now displayed and the title is not updated.

shalabhc: Hmm, it's not just this - maybe the last few commits didn't make it in, including the OSX binding and implementation cleanup.

PR #133, opened 2015-11-17
Timers

nilnor:
This adds primarily:

- An `.idle` property for Application, returning the time in seconds Howl has been idle
- A new signal `idle`, fired when Howl has been idle for a configurable amount of time (currently defaults to 60s)
- A new signal `every-second`, fired once every second

This in itself is not particularly super exciting perhaps, but it enables the addition of other features that runs upon idle. Things I'm not sure about: 

- Do we want the `every-second` signal? I see that it could be useful for easily running periodical tasks, but that might be forward engineering. It does not cost much though since we have it anyway for the idle timer.
- A good default for the idle timeout. Currently 60s, but that could be considered too short

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,8 +4,7 @@
 
 - Added a new command, `cursor-goto-line` for going to a specified line.
 
-- Added two signals, `idle` and `every-second`, which can be used for performing
-operations upon idle or periodically.
+- Added Timer.on_idle, for performing operations upon idle.
 
 - Added a new property, Application.idle, for determining how long the
 application has been idle.
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@
 
 - Added a new command, `cursor-goto-line` for going to a specified line.
 
+- Added a new property, Application.idle, for determining how long the
+application has been idle.
+
 - Added new configuration variable, `undo_limit`, for controlling the maximum
 number of revisions for each buffer.
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@
 
 - Added a new command, `cursor-goto-line` for going to a specified line.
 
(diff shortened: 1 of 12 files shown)


kirbyfan64:
One potential cause of worry with `every-second` would be the fact that, if the callback is even a little bit slow-ish, it will be brutally obvious...

shalabhc:
Very nice!

Funnily, I thought 60s for idle is too long :) Do you have any specific application for this?

What constitutes idle perhaps depends on context? Another possible API might be something like `timer.when_idle 60, do_something` which can subsume both signals.

nilnor:
@kirbyfan64 Yes, a badly written handler could definitely make things laggy, but then that's already true for the existing timer functions as well, as well as for many of the other signals that are fired. We control what goes into the core, and a user will have to control their own custom code.

nilnor:
@shalabhc That's the API I first thought about. The reason for not going with that was that then I would have to write yet another little custom dispatcher in the timer module, whereas just using a signal means everything is taken care of already. I also toyed with the idea of having `short-idle` and `idle` signals, but that didn't end up feeling good. Anyway, I do find it useful to be able to schedule tasks on shorter idle periods, so I will perhaps revisit this.

e1be9d5073a8576123e7c72bd02275c19c3f3b44 in the janitor branch contains a first example use of this, a new module for doing house-keeping on idle (closing old buffers, releasing memory).

nilnor:
So, I rewrote it to provide an `on_idle` instead. I removed both the previous signals, as I concluded that I didn't like the `every-second` signal after all.

nilnor:
Also, janitor example is now at f29b0f9eea0d2601870de076d316956fa6ff85c0 after the changes.

shalabhc:
Yup I definitely like the new API more.

PR #135, opened 2015-11-19
Implement PEP 498 string interpolation...again...

kirbyfan64:
Supersedes #118, since Git seems to hate me. Fixed all the issues.

@shalabhc I left `{val!#comment}` styled incorrectly, since there isn't really much that can be done there, I doubt that'll ever happen under normal situations, anyway.

### DIFF
--- a/bundles/python/misc/examples.py
+++ b/bundles/python/misc/examples.py
@@ -70,6 +70,17 @@ double "quotes" ok too!
 
 adjacent_strings_auto_concat = "hello " "world"
 same_as_above = "hello""world"
+raw_string = r'abc\d'
+
+simple_f_string = f'abc{de}fgh{ij}klm'
+conv_f_string = f'ab{cd!s}ef'
+complicated_triple_f_string = f'''abc{a+'!':a=+#0{abc},.{prec}b}def'''
+brace_f_string = f'{value:abc{f() + g()}}abc'
+spaced_f_string = f'{a + b}abc'
+f_string_spec = f'{v!r}'
+another_braced_f_string = f'{v:a4c{abc}}abc'
+f_string_with_bang = f'{v!=0}'
+triple_dq_f_string = f"""abc{defg}"""
 
 if False:
   # Dictionaries
--- a/bundles/python/python_lexer.moon
+++ b/bundles/python/python_lexer.moon
@@ -79,7 +79,27 @@ howl.aux.lpeg_lexer ->
   raw_string = c('special', S'rR') * basic_string
   encoded_string = c('special', S'bBuU') * any { raw_string, basic_string }
 
-  string = any { basic_string, raw_string, encoded_string }
+  f_prefix = c 'special', any {
+    S'rR' * S'fF'
+    S'fF' * S'rR'
+    S'fF'
+  }
+
+  format_conv = c('operator', P'!') * c('special', S'sra')
+  format_part = any {
+    c 'number', integer
+    c('operator', P'{') * ((V'all' + space + P 1) - P'}')^0 * c('operator', P'}')
+    c 'special', (P 1) - S'{}'
+  }
+  format_spec = c('operator', ':') * format_part^0
+
+  f_string = sequence {
+    f_prefix
+    c 'string', Cg any({ "'''", '"""', "'", '"' }), 'quote'
+    V'f_string_chunk'
+  }
+
+  string = any { basic_string, raw_string, encoded_string, f_string }
 
   decorator = c 'preproc', P'@' * name * ('.' * name)^0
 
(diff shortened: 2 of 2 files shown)


kirbyfan64:
Bump? I want to merge this soon; Howl will be one of the first editors with PEP 498 support!

shalabhc:
OK - now it works really well and simple attempts can't break it. Pretty exciting!

However I can't speak for the implementation much because I haven't been able to fully grasp cyclic references and LPEG is making my head spin a bit. I *feel* it should be possible to do this without injecting all of these in the main table, but maybe I wrong.

@nilnor sorry to bounce this back but any comment on the implementation?

Since the PEP is huge, here's a high level description of the syntax - this is Python's new feature equivalent to moon's `"#{expr} "`

The basic syntax is simple: `f'{expr:format}'` where the `:format` part is optional.

The `f` prefix is required and what makes the interpolation automatic, otherwise it's just a regular string.

`expr` is any python expression.

`format` is a format specification which is just a free form literal. The meaning of format is determined by the object that `expr` evaluates to. An the object can choose to interpret the format any way it wants . E.g. a float object might use the format `5.2` to means a total width of 5 and a precision of 2 digits after the point.  

Three features add complications.
1. Python has 4 string literal delimiters - all support interpolation with `f` prefix.
2. The format itself can nest another `{expr|format}` and that will be evaluated before the outer one e.g. `f"{myfloat:{width}:{precision}}"`
3. An optional `!s` or `!r` just before the colon is possible for some automatic transforms.

bundles/python/python_lexer.moon
```diff
@@ -83,11 +85,49 @@ howl.aux.lpeg_lexer ->
...
+  format_spec = c('operator', ':') * format_part^0
+
+  string_kinds =
+    sq_string: "'"
+    dq_string: '"'
+    tsq_string: "'''"
+    tdq_string: '"""'
+  interpolations = {}
+  f_strings = {}
+  for kind, quote in pairs string_kinds
```
nilnor: This is not right - the order of keys from `pairs` is not specified (and even if it was, and tables had a set order the above definition would still not work right). The triple quotes need to be matched first, otherwise the single variants will also match with predictably bad results. It so happens on my machine that the order ends up being `'''`, `"`, `"""`, `'`. So by luck the triple example in the example file actually works. The example file does by chance not contain an example of triple double quoted strings, but as is expected they do not work.

bundles/python/python_lexer.moon
```diff
@@ -100,4 +140,14 @@ howl.aux.lpeg_lexer ->
...
+
+    f_string: any f_strings
+
+    string: any {
+      f_prefix * V'f_string'
+      string
+    }
   }
+
+  rules[k] = v for k, v in pairs interpolations
```
nilnor: One way around having to explode all of these as separate rules is to capture the actual opening quote in a named group capture, and then match the correct ending quote using `match_back`. This used for instance in the Ruby lexer, where you can specify your own quote character e.g. for a double quoted string, and in the Lua lexer for where block quotes can have an arbitrary number of '='s. This would allow for a single body of rules for all the fstring cases, which would probably be easier to follow than having to expand the variants in one's head. I'll leave it up to you whether to pursue this at this time or not, but you might be pleasantly surprised (would take of the ordering issue at the same time as the opening quote list would be ordered).

nilnor:
All right, I've had a look now!

bundles/python/python_lexer.moon
kirbyfan64: @nilnor Oh my gosh. *I can't believe I didn't see that!*

kirbyfan64:
@nilnor @shalabhc All done!

bundles/python/python_lexer.moon
```diff
@@ -79,7 +79,30 @@ howl.aux.lpeg_lexer ->
...
+  f_prefix = c 'special', any {
+    S'rR' * S'fF'
+    S'fF' * S'rR'
+    S'fF'
+  }
+
+  format_conv = c('operator', P'!') * c('special', S'sra')
+  format_number = any {
+    c 'number', integer
+    c('operator', P'{') * ((V'all' + space + P 1) - P'}')^0 * c('operator', P'}')
```
shalabhc: The inner interpolation may not always evaluate to a number - could be any string. It might make sense to move this line to the `format_part` rule below? I don't think it makes any difference whatsoever to the parsing.

bundles/python/misc/examples.py
```diff
@@ -70,6 +70,16 @@
 
 adjacent_strings_auto_concat = "hello " "world"
 same_as_above = "hello""world"
+raw_string = r'abc\d'
+
+simple_f_string = f'abc{de}fgh{ij}klm'
+conv_f_string = f'ab{cd!s}ef'
+complicated_triple_f_string = f'''abc{a+'!':a=+#0{abc},.{prec}b}def'''
```
nilnor: I'm missing the example of the triple-double-quoted string here.

kirbyfan64:
@nilnor @shalabhc Fixed.

nilnor:
All right, in it goes :)

In the future, could we skip the squashing of commits for a PR? It makes it harder to following along with the comment and the changes done.

PR #137, opened 2015-11-19
Even *more* C/C++ lexer fixes

kirbyfan64:
This should allow pretty much all C++ types to be properly highlighted.

### DIFF
--- a/bundles/c/c_lexer.moon
+++ b/bundles/c/c_lexer.moon
@@ -32,19 +32,19 @@ howl.aux.lpeg_lexer ->
 
   classdef = any {
     sequence {
-      c 'keyword', word { 'enum', 'union' }
+      c('keyword', word { 'enum', 'union' })
       ws^1
-      c 'type_def', ident
+      c('type_def', ident)
       ws^0
       c('operator', '{')
     }
     sequence {
-      c 'keyword', word { 'class', 'struct' }
+      c('keyword', word { 'class', 'struct' })
       ws^1
-      c 'type_def', ident
+      c('type_def', ident)
       ws^0
       (((V'all' + P 1) - S':{}=;') + ws)^0
-      c 'operator', S':{'
+      c('operator', S':{')
     }
   }
 
--- a/bundles/c/c_lexer.moon
+++ b/bundles/c/c_lexer.moon
@@ -4,7 +4,7 @@
 howl.aux.lpeg_lexer ->
   c = capture
   ident = (alpha + '_')^1 * (alpha + digit + '_')^0
-  ws = c 'whitespace', blank
+  ws = c 'whitespace', space
 
   identifer = c 'identifer', ident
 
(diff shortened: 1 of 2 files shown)


bundles/c/c_lexer.moon
```diff
@@ -32,22 +32,23 @@ howl.aux.lpeg_lexer ->
 
   classdef = any {
     sequence {
-      c('keyword', word { 'enum', 'union' })
+      c 'keyword', word { 'enum', 'union' }
       ws^1
-      c('type_def', ident)
+      c 'type_def', ident
```
nilnor: Please avoid these kind of cosmetic changes, when it's not code that is actually touched by any modifications. It just makes the change unnecessarily verbose, and for instance for this block it makes things less consistent, as one occurrence of the original style is left. For these table listings I actually prefer the parentheses to be there.

nilnor:
As always, I'm opposed to merging lexer changes without accompanying example updates. What are the things that are now lexed correctly compare to before? Examples of these should be added to the `example.c` file.

kirbyfan64:
@nilnor Sorry about that! I undid the style changes and updated the example file.

The changes were:

- Properly highlight type names in C++ template specializations.
- Highlight the operators `,`, `?`, and `:`.

nilnor:
Thanks for the updates :)

Looking at the lexer changes, it now really allows *anything* before the colon or brace, where it previously was very strict. My C++ is rusty by now, but that seems excessive. The example file contains one rather specific example, which does not seem to warrant this. Can the specialization start with anything other than a `<` for example? In short, I think the allowed syntax should be limited to what there is rather than everything.

kirbyfan64:
@nilnor Because `struct X<A, B> y;` is not a type, but `struct X<A, B> virtual : Z` is?

Just take a look at the new example file. And all the additions. It's pretty crazy stuff. :D

bundles/c/c_lexer.moon
```diff
@@ -41,13 +48,15 @@ howl.aux.lpeg_lexer ->
     sequence {
       c('keyword', word { 'class', 'struct' })
       ws^1
-      c('type_def', ident)
+      attribute_spec^0
+      c('type_def', ident) * (c('operator', P'::') + c('type_def', ident))^0
       ws^0
+      (((V'all' + P 1) - S':{}=;') + ws)^0
```
nilnor: If this is going to be referencing grammar rules I'd rather have it in the grammar itself. That leaves anything outside of the grammar as "simple" helpers.

nilnor:
Yes, that's a surprising variety of allowed constructs :) Still, I don't think include statements, preprocessor statements and other class definitions can follow? So I think it would make sense to just have an explicit grouping with the allowed grammar elements, for clarity if nothing else.

kirbyfan64:
@nilnor The problem is kinda that almost anything is allowed. If the class is followed by a `<`, like here:

```c++
class X<Y>
```

this still may or may not be a class definition, depending on what comes after the `>`. Even worse, literally anything is allowed between the `<` and `>`, like:

```c++
struct X<A, 1>>2, 3<<4>
```

I mean, it's *possible* to enumerate all the possibilities, but C++ is crazy enough that I think it's out of the scope of this lexer.

nilnor:
All right then :)

PR #140, opened 2015-11-23
More extensive change()

nilnor:
This is a performance/efficiency improvement. The major theme is to extend the `change()` operation introduced a while ago. The original `change` collapsed buffer modifications notification into one single notification, with great results for larger operations such indenting large blocks of code.

This takes it further by also collecting and collapsing styling and marker changes, which makes it possible to optimize batch changes for text based interface buffers. This PR does just that for the ListWidget, where the write_page now writes everything in one change. 

Since I was primarily interested in reducing memory pressure I measured that, and the re-allocation of display lines is cut by approximately a third e.g. when interacting with `project-open`. From a performance perspective this also cuts down a lot on unnecessary display refreshes, which makes things snappier, although it was pretty snappy already.

This also optimizes marker invalidations to not re-allocate display lines but instead just refresh flairs  (which in extension means snappier refreshes of Howl highlights).

### DIFF
--- a/lib/aullar/buffer.moon
+++ b/lib/aullar/buffer.moon
@@ -541,6 +541,12 @@ Buffer = {
     start_line: start_line.nr, end_line: end_line.nr, :invalidated
 
   _on_modification: (type, offset, text, prev_text, size, invalidate_offset, extra) =>
+    lines_changed = text\find('[\n\r]') != nil
+    if not lines_changed and prev_text
+      lines_changed = prev_text\find('[\n\r]') != nil
+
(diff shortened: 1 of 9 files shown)


shalabhc:
Nice! This makes a visible difference in heavy use of markers. E.g.

- the response time for highlighting every character with `buffer-regex-replace .` and also subsequently with pressing the down arrow.
- speed of scrolling through the list in `buffer-grep-regex .+`

nilnor:
FWIW I explained it wrong when talking about the gains, the re-allocations was cut down *to* a third of the original.

PR #142, opened 2015-11-25
Change early-exit behaviour in signal emission to be explicit

nilnor:
The `signal` module is one of the earliest modules written, and it was written from day one with an early-exit mechanism based on boolean values from signal handlers. This might have worked out fine if Lua had been used, but since we're using moonscript with its implicit returns it ended up a giant mistake, which makes it all to easy for signals to lost due to some implicit return somewhere.

The early-exit option is retained, but is now rather explicit due to the use of a required string return value instead.

### DIFF
--- a/lib/howl/bindings.moon
+++ b/lib/howl/bindings.moon
@@ -186,7 +186,8 @@ export process = (event, source, extra_keymaps = {},  ...) ->
   event = substitute_keyname event
   translations = translate_key event
   return true if process_capture event, source, translations, ...
-  return true if signal.emit 'key-press', :event, :source, :translations, parameters: {...}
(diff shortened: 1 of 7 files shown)


kirbyfan64:
I like this idea, but I feel like using a string constant doesn't seem quite right.

Tables are compared by reference, right? Would it be possible to have something like:

```moonscript
abort = {}
...
if the_return_value == abort
```

nilnor:
I agree with you that using special strings can feel ugly. The latest commit switches to a special sentinel value instead, so have a look at that instead.

kirbyfan64:
:+1

I think the tests failing might just be a Travis thing. Closing-and-reopening this PR should re-run the build.

nilnor:
Yes, the test's a Travis glitch.

Issue #145, opened 2015-12-02
Errors with nested styles


kirbyfan64:
Just try creating a Markdown file with the following:

    ```lua
    ffi.cdef [[
    
    ]]
    ```

and then type in between the brackets.

kirbyfan64:
Still doesn't work. Create a Markdown file containing:

    ```lua
    ffi.cdef [[{}
    ]]
    ```

Then put the cursor after the first `{` and hit backspace.

![howl_sub](https://cloud.githubusercontent.com/assets/1690697/12212348/dfd52536-b62e-11e5-9943-5f94e47d11cf.png)

nilnor:
Ok, so this should be fixed now, but I'll leave it open until you have had a chance to try something else wrong :)

kirbyfan64:
Looks good!

PR #146, opened 2015-12-02
Implement sub modes

### LINKED ISSUE #138: Better support for embedded languages
Case in point: open up an HTML file, put:

```
function() {
```

and hit enter after the '{'. It won't indent like the JavaScript mode does.

kirbyfan64:
Closes #138. I threw every case I could think of at it, save multiple nested modes, since those don't work yet anyway (see #145).


kirbyfan64:
BTW, my other idea as to how to implement was a lot simpler, but `Styling.apply` seems to occasionally be off-by-one with the offsets, so I had to go the more complicated route.

nilnor:
Well if so it sounds like a better idea to narrow down and fix the issue with Styling.apply and then proceed with the simpler solution.

Back to the PR for some quick feedback: 

- There are no tests for anything of this. This is a typical thing that really needs massive test coverage, covering all possible angles.
- Nothing is done for `invalidate_from`, meaning things will definitely become incorrect
- There is an existing module, `aullar.markers` which does pretty much exactly this handling of ranges

kirbyfan64:
@nilnor 

- Didn't know about `aullar.markers`! I'll definitely try it out.
- Leaving out the tests was intentional. I usually leave that for last, since a lot of stuff always changes in PRs, and I hate updated them OVER and OVER again!

kirbyfan64:
@nilnor I changed it to use markers and fixed the `invalidate_from` issue. All done except for tests (which I again usually leave for last).

kirbyfan64:
Actually, I forgot about configs. Darn. :/

kirbyfan64:
Slight problem: config.

This is actually the hardest part! See, a config can only have one base. BUT this causes trouble with sub modes.

Say the `word_pattern` needs to be queried. Just getting `buffer.config.word_pattern` will get always get the config entry from the main mode, not the sub mode. So I added `buffer.config_at`. And it doesn't work correctly! When `init.moon` sets properties on `args.buffer.config`, none of them will have effect in `config_at`. `config_at` creates a new config and chains it to `mode_at(pos).config`. Therefore, none of the config settings from `buffer.config` will come up in `config_at`.

Is this a good thing or a bad thing?

shalabhc:
There seem to be some subtleties here that are not immediately clear, at least to me. I'm fine with just leaving out support for `config_at` and check in the rest of the stuff because it may still be useful (consider `toggle-comment` which would work with sub modes).

Maybe some of the config vars such as `word_pattern` shouldn't really be configs at all, at least not at the buffer scope.

kirbyfan64:
Yeah, I would say `word_pattern` should be a mode thing, and configs should always span the entire buffer. That would definitely be the simplest way to go.

nilnor:
I had a quick look now. Good to see that the markers module seems to work here, that cuts down on a lot of stuff. `invalidate_from` is now covered, but `reset`, `set` and `clear` seems not to be - go through everything and make sure all bases are covered.

`styling.clear_style_offsets` shouldn't be there at all, as this should be transparent from the other operations. At no point should `howl.Buffer` muck around with the internals, for the case where this seems to be used (howl.Buffer.lexer =), this would be done already by the aullar.buffer.lexer assignment.

`styling._insert_sub_markers` can be in-lined for better readability.

While it's not for me to dictate whether tests should be written simultaneously or after, I'll note that retrofitted tests are always almost subpar. For core functionality, as this is, the test coverage needs to be extensive, so make sure to give the test some real love. Considering this, this PR is currently at most halfway done, so better get cracking on those test :)

We can untangle the config business after that.

kirbyfan64:
FYI, I'm sorry for the delays. I've been really busy. Hopefully, I'll be able to pick this up again next week...

nilnor:
@kirbyfan64 No worries, take the time needed!

kirbyfan64:
Almost there! Just need to make nested sub modes work now.

kirbyfan64:
...Git wrecked my history didn't it?? :( Well, it should merge cleanly anyway!

I can't get nested sub modes to work. I'm pretty sure it's actually an issue with the implementation of the fix for #145; it seems to sometimes want to lex the whole block as the first language and sometimes the second.

nilnor:
You merged in lots of commits from master now unfortunately, which makes this diff unreadable. Could you clean up the branch and remove the master commits and force push the clean branch?

PR #153, opened 2016-01-08
Rework buffer titles in `switch-buffer` list

shalabhc:
- Removed title counters
- Default title is always basename
- Title unique-ification moved to switch-buffer command
- Unique-ification uses project name and parent directory name up to two levels

This also removes any counters and directory prefixes from the header bar of the editor - only the basename is displayed when viewing a file.

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -14,23 +14,6 @@ ffi = require 'ffi'
 append = table.insert
 min = math.min
 
-buffer_titles = setmetatable {}, __mode: 'v'
-title_counters = {}
-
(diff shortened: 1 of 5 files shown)


nilnor:
Code looks good as far as I can tell, but I'm seeing spec failures for `janitor_spec` and `inbuffer_completer_spec`. Was this dependent on anything else?

shalabhc:
After my last commit I see everything pass, even on a merged branch. I'm not quite sure why using `open_file` was wedging stuff but it's probalby better to just use new_buffer only for buffer related tests.

PR #154, opened 2016-01-08
Added howl.sys.time for high resolution time

shalabhc:
- uses g_get_real_time for microsecond resolution
- returns time as seconds (float) so comparable with os.time

Replaced use of os.time with sys.time

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ## Unreleased (in master)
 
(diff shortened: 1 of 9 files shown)


nilnor:
Looks reasonable to me, but we should add the new function to the documentation for the sys module.

shalabhc:
Done - please take another look.

nilnor:
:+1:

PR #157, opened 2016-02-09
UI rework

nilnor:
This moves away from the old CSS based theming. CSS theming is still used, but only for the status and indicator bars. All other elements have been updated to to use a new ContentBox class extensively. The ContentBox is a generic, stylable, container class supporting a three-element widget.
It contains a main widget, as well as a optional header and optional footer. It's stylable via themes, and will automatically apply any theme changes for its own type (which defaults to `content_box`).

The major driver for for this is to allow for more advanced themes, and allow things such as border radiuses, transparency everywhere, etc. The current themes have been adapted for the new structure but have been kept pretty much the same. One thing that we're loosing currently is the additional border in the command line between the entry and any widget above, which is since it doesn't really fit into the new structure. I can live with that (it used to be that way earlier as well), but if not something would have to be added to restore this.

The plan now is to add some new custom Howl themes that take advantage of the new possibilities, so as such this can be considered still incomplete. It will be easier to see the effects of this once new themes are added. Still, this is the base and is unlikely to change that much.


shalabhc:
Very nice!

I see a white header and status bar - is that expected?

howl-ui-rework

nilnor:
Hmm, no, that shouldn't be the case. Do you get any errors/warnings? Could be problem with some version of something, so what's the environment?

nilnor:
Never mind, I managed to reproduce this with Ubuntu 14.04, will look at it.

nilnor:
@shalabhc the white bars should be fixed now,

shalabhc:
@nilnor - verified, it's fixed.

nilnor:
So, I finally finished a new custom theme for Howl, "steinom". And with that I'm done with this PR, so please review. Screenshot:

steinom

kirbyfan64:
@nilnor I don't think it's supposed to look like this...

howl-tstein

nilnor:
@kirbyfan64 Indeed not, version incompatibilities again it seems. I will look at it.

shalabhc:
Wow - looks great in the screenshot!

Unfortunately looks like @kirbyfan64's version on my machines :neutral_face:

nilnor:
I whacked at it with the magic compatibility wand so please have another look!

bundles/howl-themes/README.md
```diff
@@ -0,0 +1,38 @@
...
+This bundle contains custom made themes for the Howl editor.
+
+## License
+
+The themes make use of patterns from [Subtle Patterns](subtlepatterns.com),
+licensed under the
+[Attribution-ShareAlike 3.0 Unported](http://creativecommons.org/licenses/by-sa/3.0/deed.en_US)
+license.
+
+The bundle itself is under the following copyright:
```
shalabhc: Lines 14 - 20 have some redundant info and unusual flow.

bundles/howl-themes/init.moon
```diff
@@ -0,0 +1,18 @@
+ --Copyright 2016 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+olive = bundle_file('steinom/steinom.moon')
```
shalabhc: Micro-nit - what is olive? Old name?

nilnor: Ah, indeed, good catch.

shalabhc:
The background issue is now fixed for m and Steinom is very nice!

Overall looks great.

shalabhc:
:+1: 

Do all bundled themes belong in `howl-themes` now?

nilnor:
@shalabhc Yes, that's what I thought. So the plan would be to collect the other two (three with the Monokai theme) themes into this bundle and have one for all bundle themes.

kirbyfan64:
Was the changing of the colors in the auto-complete popup in the Tomorrow Night Blue theme intentional? It looks a little odd to me.

Also, the new color in sub lexing looks slightly...blinding...

nilnor:
@kirbyfan64 Not sure how the embedded background ended up being changed, the border was from some earlier experimenting. Have a look at master now.

PR #158, opened 2016-02-09
Moved directory from process buffer title to directory column

shalabhc:
Process buffer titles in the `switch-buffer` command can really blow up the width of the first column if the directory is deep. This moves the directory from the first 'title' column to the second column, also making it consistent with where the file buffer directories are shown.


nilnor:
:+1:

PR #159, opened 2016-02-10
Sub modes...again

kirbyfan64:
Because Git hates me with a passion.

I'm *still* trying to get nested ones to work; I have a feeling it's going to be one of those one-line fixes that takes forever to figure out.

Also, still need more specs!

### DIFF
--- a/lib/aullar/spec/styling_spec.moon
+++ b/lib/aullar/spec/styling_spec.moon
@@ -253,6 +253,24 @@ describe 'Styling', ->
           6, 's1:s2', 7
         }, styling\get(1, 7)
 
+        assert.same nil, styling\get_nearest_style_marker 1
+        marker =
+          name: 'my_sub'
+          start_offset: 2
+          end_offset: 5
+        assert.same marker, styling\get_nearest_style_marker 2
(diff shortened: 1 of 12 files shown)


kirbyfan64:
...so, there's this REALLY weird behavior with the specs that's not reproducible in a normal Howl session...but, other than that, I think this is *generally* ready for another code review. After, like, 3 months...

shalabhc:
In my experience weird behavior with specs happens because specs don't always clean up perfectly after themselves and may use mocks. If your specs pass when run in isolation but fail when run as part of the suite, that *might* be the reason. Try resetting the relevant state in the `before_` hook for your spec.

kirbyfan64:
I'm still trying to work on this. Right now I'm warring with Git over how to rebase the branch, since 8a3aac836d79356b5b8ec7f2502bb15ba90b1640 broke the support I had put in for nested modes...

kirbyfan64:
Would you guys mind if I squashed this whole thing into one commit? It'll make the merging I'm trying to do *much* easier.

kirbyfan64:
Yeah, I just did it with the commits squashed, and it was a piece of cake.

PR #161, opened 2016-02-13
Keymap changes for #112

### LINKED ISSUE #112: More key bindings
Here are some key bindings I have in my `init.moon` file that might be useful to add to Howl itself:

- `ctrl_shift_k` for `editor-delete-line`: This one is really, REALLY useful. I'm addicted. :)
- `ctrl_shift_x` for `buffer-close`: `ctrl_shift_w` might actually be a better fit for this one. Basically, when using processes, I also use this one a lot.
- `ctrl_shift_h` for `buffer-replace-regex`: Handy for replacing variable names, since you can stuff like `\b` to restrict the replace to just words.

Thoughts?

shalabhc:
- Changed `ctrl_w` to run `buffer-close` not `view-close`. Added `ctrl_shift_w` for `view-close`.
- Added `ctrl_[shift_](home|end)` bindings for moving cursor to start or end of buffer, optionally extending selection.

This fixes #112.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -55,6 +55,10 @@ code-named `aullar`.
 - The `howl-moon-eval` command was improved by automatically adjusting the
 indentation levels to work as a stand-alone code chunk.
 
+### Keymap changes
+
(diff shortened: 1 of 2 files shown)


nilnor:
:+1:

PR #163, opened 2016-02-15
Add delete forward/back word bound to ctrl delete/backspace.

barnybug:


lib/howl/ui/editor.moon
```diff
@@ -959,7 +969,9 @@ for cmd_spec in *{
   { 'smart-tab', 'Inserts tab or shifts selected text right', 'smart_tab' }
   { 'smart-back-tab', 'Moves to previous tab stop or shifts text left', 'smart_back_tab' }
   { 'delete-back', 'Deletes one character back', 'delete_back' }
+  { 'delete-back-word', 'Deletes one word back', 'delete_back_word' }
   { 'delete-forward', 'Deletes one character forward', 'delete_forward' }
+  { 'delete-forward-word', 'Deletes one word back', 'delete_forward_word' }
```
kirbyfan64: "Deletes one word **forward**"

barnybug:
oops, fixed :-)

lib/howl/keymap.moon
```diff
@@ -16,7 +16,9 @@
     tab:              'editor-smart-tab'
     shift_tab:        'editor-smart-back-tab'
     backspace:        'editor-delete-back'
+    ctrl_backspace:   'editor-delete-back-word'
```
shalabhc: Could you please also add the `osx` equivalents in the section below? I think `meta_backspace` and `meta_delete` should work.

barnybug: I think it'd be bound to Option+Backspace/Delete on the Mac - but I noticed this comment re option key in the keymap: https://github.com/howl-editor/howl/blob/master/lib/howl/keymap.moon#L152

shalabhc: OK, Option+* variants will work too. I think Word used Command+Backspace, Atom uses Option+Backspace and Lifehacker says both work:

http://lifehacker.com/5803041/the-mac-os-x-delete-key-it-goes-both-ways

Perhaps just add commented out code using `option_backspace` etc. in the osx section then? Next time I fire up Howl on OSX I'll try and set up the the `option` key translation.

barnybug:
Good idea. Added comment.

nilnor:
Looks good! One nitpick, and it's the behaviour when a selection is active. Even though it probably doesn't make sense to type these shortcuts with a selection active, it would probably indicate an accidentally pressed `ctrl` key rather than anything else, in which case it would probably make sense to just delete the selection instead of attempting anything else.

nilnor:
Oh, and disregard the failing travis for now, the memory tests have been proven to be fickle on Travis and something will have to be done, but it's unrelated to this.

barnybug:
Thanks for the suggestion - selection behaviour improved.

nilnor:
All right, looks good!

PR #162, opened 2016-02-14
Monokai theme

barnybug:
First off - thanks for this great text editor!
I've created the ubiquitous Monokai theme for howl.

PR #164, opened 2016-02-16
Monokai theme ui rework

barnybug:
Rework of #162 for the ui-rework branch.


barnybug:
Oops, ignore - created against wrong base! See #165

PR #166, opened 2016-02-17
Add some more extensions and filenames for ruby mode.

barnybug:
Small change. Was looking for .jbuilder files to use ruby mode. Thought I'd add a few other common ones at the same time (inspired by the list in sublime).


nilnor:
Thanks!

PR #167, opened 2016-02-17
Add editor-{copy,cut}-current to copy/cut the selection or otherwise current line

barnybug:
Similar to editor-duplicate-current and models the behaviour of a few other text editors - if you haven't a selection Ctrl+C/X will copy/cut the current line, instead of doing nothing. I find this very handy for moving a single line of code around.

### DIFF
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -1002,8 +1002,8 @@ for cmd_spec in *{
   { 'scroll-up', 'Scrolls one line up', 'scroll_up' }
   { 'scroll-down', 'Scrolls one line down', 'scroll_down' }
   { 'duplicate-current', 'Duplicates the selection or current line', 'duplicate_current' }
-  { 'cut', 'Cut the selection or current line to the clipboard', 'cut' }
-  { 'copy', 'Copy the selection or current line to the clipboard', 'copy' }
+  { 'cut', 'Cuts the selection or current line to the clipboard', 'cut' }
+  { 'copy', 'Copies the selection or current line to the clipboard', 'copy' }
   { 'cycle-case', 'Changes case for current word or selection', 'cycle_case' }
 }
   args = { select 4, table.unpack cmd_spec }
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -470,6 +470,19 @@ class Editor extends PropertyObject
       @buffer\insert @selection.text, max(anchor, cursor)
       @selection\set anchor, cursor
 
+  cut: =>
+    if @selection.empty
+      clipboard.push text: @current_line.text, whole_lines: true
+      @buffer.lines[@cursor.line] = nil
+    else
+      @selection\cut!
+
+  copy: =>
+    if @selection.empty
+      clipboard.push text: @current_line.text, whole_lines: true
+    else
+      @selection\copy!
+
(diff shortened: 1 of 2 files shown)


shalabhc:
Could you please share which other editors have this behavior for the standard cut/copy bindings? It seems a bit strange to me, though I'm not against new bindings for these.

barnybug:
Sublime text:
https://gist.github.com/eteanga/1736542
and komodo:
http://komodoide.com/blog/2013-01/komodo-tips-tricks/

barnybug:
Visual studio (not been a user in a long time though so can't verify!)

barnybug:
Netbeans too.

Atom and Eclipse both have plugins doing this; (Atom by default deletes (not cuts) a single character forward of the cursor on ctrl+x with no selection, somewhat bizarrely).

shalabhc:
OK then, I vote [+0](https://www.python.org/dev/peps/pep-0010/). @nilnor has the final say, of course.

But please update the `osx` versions too :smiley:

barnybug:
Better than a -0 at least :-)
Split the PR into two commits - one adding the commands, and second updating the key bindings (mac too).
Can drop the keybinding change if it isn't entirely agreeable.

nilnor:
My thought is whether it makes sense to keep the original operations in this case, or whether it 's better to just update the original methods. @shalabhc, @kirbyfan64: any votes on that? :)

kirbyfan64:
Well, you could always change the bodies of `selection\cut` and `selection\copy` to select the current line if there is no active selection, instead of just returning. The problem is just that `Selection` (rightfully) doesn't know anything about the current buffer and cursor...

Maybe the `copy` and `cut` operations on `Selection` could take an argument `default_selection` or something?

shalabhc:
I don't think the `cut` and `copy` methods on `selection` should be changed at all.

However, I'm in favor of changing the existing 'editor-copy' and 'editor-cut' commands instead of adding the new '-current' variants. The new commands are so similar and provide a superset of the functionality - why would anyone need 'editor-cut' but not just use 'editor-cut-current'?

nilnor:
Yes, sorry about being unclear, but what I was attempting to say was what @shalabhc formulated more clearly, that the existing commands would be replaced rather than having two new variants. Well, it seems we think along the same lines though, so unless @kirbyfan64 is vehemently opposed let's go with updating the existing commands instead.

kirbyfan64:
Sure!

barnybug:
Thanks guys - nice to know changes are carefully considered. I've made the suggested changes.

lib/howl/ui/editor.moon
```diff
@@ -989,6 +1002,8 @@ for cmd_spec in *{
   { 'scroll-up', 'Scrolls one line up', 'scroll_up' }
   { 'scroll-down', 'Scrolls one line down', 'scroll_down' }
   { 'duplicate-current', 'Duplicates the selection or current line', 'duplicate_current' }
+  { 'cut', 'Cut the selection or current line to the clipboard', 'cut' }
```
nilnor: Final nitpick - the descriptions should be fixed to be in the same form as the other descriptions, e.g. "Cuts the", "Copies the". After that we're good to go :)

barnybug:
Wording tweaked.

nilnor:
Looks great, thanks!

Issue #156, opened 2016-01-28
Undo is too greedy


kirbyfan64:
Basically:
1. Open Howl.
2. Enter some text.
3. Paste something.
4. Hit `ctrl_z`.

Both the entered AND pasted text will be removed, but only the pasted text should be.

shalabhc:
Seconded.

Additionally, sometimes undo will undo too slowly. E.g. after Howl auto indents a new line on `enter`, invoking undo first undoes the indent, and the next undo removes the newline. I would expect this to be one step for the undo (it was one keystroke for me).

nilnor:
Just for the record the second issue was addressed by 2a217bd.

ozra:
Ah, now I can remove one of the issues on the list I've built up. Reacted to the exact same thing, edit, paste - different snap-points. Ideally every char entered is an event in it self (this can easily be optimized like a common special case, storage wise, when there's a long string of chars entered in linear sequence. Or rather - only the undo-func needs to be changed to pop one char of at a time of every snap-segment that has been entered manually (a paste is of course one unit).

PR #168, opened 2016-02-19
Reset buffer 'modified' status after undo when applicable

shalabhc:
The `buffer.modified` flag is now automatically reset to false if you undo all the way back to the originally synced revision. This avoids unnecessary 'Buffer is modified' warnings on close/exit and also the misleading '*' marker in the buffer list.

(I'm not quite sure if this is the best way to do this, but it works :smile: ).

### DIFF
--- a/lib/aullar/buffer.moon
+++ b/lib/aullar/buffer.moon
@@ -447,6 +447,10 @@ Buffer = {
 
   clear_revisions: => @revisions\clear!
 
+  snapshot: => @revisions\snapshot!
+
+  is_snapshot: (snapshot_id)=> @revisions\is_snapshot(snapshot_id)
+
   notify: (event, parameters) =>
     for listener in *@listeners
       callback = listener["on_#{event}"]
--- a/lib/aullar/buffer.moon
+++ b/lib/aullar/buffer.moon
@@ -447,9 +447,10 @@ Buffer = {
 
   clear_revisions: => @revisions\clear!
 
-  snapshot: => @revisions\snapshot!
-
-  is_snapshot: (snapshot_id)=> @revisions\is_snapshot(snapshot_id)
+  get_revision_id: (snapshot=false) =>
+    if snapshot and @revisions.last
+      @revisions.last.dont_merge = true
+    return @revisions.revision_id
 
   notify: (event, parameters) =>
     for listener in *@listeners
--- a/lib/aullar/revisions.moon
+++ b/lib/aullar/revisions.moon
@@ -30,6 +30,7 @@ define_class {
   new: =>
     @clear!
     @processing = false
+    @snapshot_id = 0
 
   properties: {
     last: => @entries[@current]
@@ -43,7 +44,7 @@ define_class {
     group = @grouping > 0 and @group_id or nil
     entry =  :type, :offset, :text, :prev_text, :meta, :group
     last = @last
-    if last and entry.group == last.group
+    if last and entry.group == last.group and not last.dont_coalesce
       return last if coalesce(entry, last)
 
     @current += 1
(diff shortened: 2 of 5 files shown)


kirbyfan64:
Oooohh!! :+1

nilnor:
It would be nice to fix this, but I think we can do it a little bit more generic. The current approach of syncing it with a file etag means for instance that it won't work with non-file buffers. My thinking is that it could suffice with something similar but simpler. I propose that we add sequential ids for each revision, and have the revisions module maintain a `current_revision` property. The `howl.Buffer` class could then just store the revision at which it synced last (with file or the creation of the buffer). Thoughts?

shalabhc:
Yes, unsaved 'Untitled' buffers don't work with this PR. I think your suggestion makes sense and I'll work on it. 

However I suspect a problem that I haven't verified: revisions can coalesce new changes so the revision id may end up being inconsistent with the revision. This would also be a problem with the current PR.

E.g:

1. Save file (say current_revision = 3)
2. Make a change which happens to get coalesced into the current revision.
3. Make another change which happens to create a new revision (now current_revision = 4)
4. Undo - this jumps back to current_revision = 3 so the modified flag is reset, but the file itself contains the unsaved change made at step 2.

Maybe we needs some kind of explicit 'tag current revision' method which then also disallows further coalescing?

shalabhc:
I've force pushed an alternative implementation. This uses a new `revisions\snapshot` method which attaches a unique snapshot_id to the current revision and also prevents further coalescing. A `revisions\is_snapshot` method checks if the current revision has a given snapshot id.

Fixed `redo` so that can also reset the `modified` flag.

nilnor:
Much nicer. However, we'll still need the general ability to mark a revision as not eligible for coalescing (e.g. for #156), so I think we should add that general ability and use that instead of having it be part of a special snapshot operation. That leaves the special snapshot id - do we need that if we add general ids for revisions, couldn't we just store the current revision id in `howl.Buffer` when saving, etc. and compare to that?

shalabhc:
Couple of things:

- IIUC coalescing is different from grouping and `snapshot()` only prevents coalescing, not grouping. But maybe for #156 we might need to force close the current group?

- I thought of using a general revision id for each revision but felt it didn't make a great API. Either the revision can be modified after the revision_id is read, in which case it's not a very useful property; or the operation of reading the revision_id should also mark the revision as immutable - which is the same as current the `snapshot` operation.

nilnor:
I don't think we need need to close the current group for 156, an explicit group takes precedence I believe. Even if not, revision ids could be allocated for each logical revision and thus stay stable even when sampled within a grouping operation.

On to the other point, how could could be the revision be modified given that coalescing has been disabled for that revision?

shalabhc:
> I don't think we need need to close the current group for 156

OK

> On to the other point, how could could be the revision be modified given that coalescing has been disabled for that revision?

I agree a revision id is useful only if coalescing has been disabled for that revision. The question is what the API should look like. If there is a method `disable_coalesce` and a property `revision_id` then reading `revision_id` without first calling `disable_coalesce` is prone to subtle bugs, but this is not obvious given the API. So what I'm suggesting is that accessing the revision_id should force disable coalescing on any revision, and this is what `snapshot` does.

nilnor:
But `snapshot` is only a fine name for this particular use case, if that. I would expect something called snapshot to do something else, such as constructing a in-memory view of some thing or other.  For the other case, where all we really want is to disable coalescing, we would call `snapshot`? Much better in that case to have something with a name that conveys what it does. I agree that it could be surprising to have a referenced sampled revision being changed, but I'm not sure it really would be a big issue. At any rate, supposing there already is way of just disabling the coalescing we could have something named `get_stable_revision_id` for this purpose.

shalabhc:
I see, good point about the `snapshot` name being to specific. Exposing a way to disable coalescing on a revision is reasonable. Maybe the `buffer` can expose a `get_revision_id` method which automatically disables coalescing, but the underlying revisions object does not disable coalescing automatically.

nilnor:
Sounds good!

shalabhc:
Done - please take another look.

I added a single method `buffer\get_revision_id(snapshot=false)` instead of different methods for stable vs. open revision id, so that its easily discovered by users.

nilnor:
Great!

PR #165, opened 2016-02-16
Monokai theme for ui-rework branch

barnybug:
Rework of #162 for the ui-rework branch.

PR #170, opened 2016-02-22
Monokai theme

barnybug:
Replacement PR for #165 based off master.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Added a new theme, `monokai`.
+
 - Added a new theme, `steinom`.
 
 - Added a new function `sys.time()` which returns the POSIX time for the system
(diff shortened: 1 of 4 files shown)


shalabhc:
If I merge this into the current master, the monokai theme doesn't quite look right. I also see this error when I try and switch to 'monokai':

Error invoking listener for change (gutter_styling: table: 0x40c3dc18 -> table: 0x415c3568): ../lib/aullar/gutter.moon:21: bad argument #1 to 'pairs' (table expected, got string)

barnybug:
That's odd, works fine for me. That error was the same I saw prior to updating the gutter background part (as in the old branch).

shalabhc:
Sorry false alarm - was trying out the 'monokai-theme' branch instead of this one :confused: 

Works great!

nilnor:
Great work, thanks!

PR #172, opened 2016-02-23
Add shell-like tab action on file selection.

barnybug:
From tab completion in the shell, muscle memory always has me expecting tab in file selection to do something.

The idea I've implemented is to make it perform like bash/fish shell - if there is a unique entry, open that entry, otherwise cycle the selection through matched (non-new) entries. Feels pretty natural. Thoughts?

### DIFF
--- a/lib/howl/interactions/file_selection.moon
+++ b/lib/howl/interactions/file_selection.moon
@@ -139,15 +139,13 @@ class FileSelector
         if @list_widget.selection and not @list_widget.selection.is_new
           @_open!
       else
-        -- if we're not on the new entry, go forward once
-        @list_widget\select_next! unless @list_widget.selection.is_new
-        -- if we land on the new entry (or were on it originally), skip it by going forward again
+        @list_widget\select_next!
+        -- if we land on the new entry, skip it by going forward again
         @list_widget\select_next! if @list_widget.selection.is_new
 
     shift_tab: =>
-      -- if we're not on the new entry, go backward once
-      @list_widget\select_prev! unless @list_widget.selection.is_new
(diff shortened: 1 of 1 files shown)


kirbyfan64:
This is *awesome*! My OCD caused me to emit some really specific style notes, but it's still great.

lib/howl/interactions/file_selection.moon
```diff
@@ -111,23 +111,39 @@ class FileSelector
...
+      for item in *@list_widget.items
+        count += 1 unless item.is_new
+      if count == 1
+        -- open a unique completion
+        if @list_widget.selection and not @list_widget.selection.is_new
+          @_open!
       else
-        @_submit file
+        -- cycle non-new entries
+        @list_widget\select_next! unless @list_widget.selection.is_new
```
shalabhc: I'm not sure the `unless` clause here is required and it confused me a bit.

barnybug: If you press tab whilst on the New entry, it would advance two positions otherwise. Hopefully an updated comment will clarify.

shalabhc: It shouldn't because the second `select_next` only fires if it ends up on a New entry *after* the first advance?

nilnor:
One thing: If `tab` cycles the selection downwards, I would expect `shift_tab` to do the opposite.

shalabhc:
Should `shift_tab` go back then? I expected it to when playing with this.

barnybug:
Yeah, good point re shift tab @nilnor @shalabhc Just occurred to me this morning. Added.

shalabhc:
Ok, looks good!

nilnor:
:+1:

PR #173, opened 2016-02-26
Sub modes

kirbyfan64:
Because I work with Git _so_ well...

Yeah, it was _way_ easier to rebase the changes like this. I apologize for the last two PRs, as well as the fact that I just killed the entire history...


lib/howl/buffer.moon
```diff
@@ -267,6 +268,18 @@ class Buffer extends PropertyObject
       b_end_pos = @byte_offset end_pos
       @_buffer\ensure_styled_to pos: b_end_pos
 
+  mode_at: (pos) =>
+    b_pos = @byte_offset pos
+    marker = @_buffer.styling\get_nearest_style_marker b_pos
+    if marker then mode.by_name marker.name else @mode
+
+  config_at: (pos) =>
+    new_config = config.local_proxy!
```
shalabhc: Minor nit but you don't need to create new_config until after the conditional return.

lib/howl/completer.moon
```diff
@@ -54,7 +53,13 @@ class Completer
     seen = {}
     completions = {}
 
-    for completer in *@completers
+    mode = @buffer\mode_at pos
+    completers = if mode == @buffer.mode
+      @completers
+    else
+      load_completers @buffer, context, mode
```
shalabhc: Should this be cached? There will probably be the same few sub-modes in a buffer. `@completers` could be a table mapping `mode` -> result of `load_completers()`

lib/howl/ui/editor.moon
```diff
@@ -212,6 +212,12 @@ class Editor extends PropertyObject
       start, stop = @selection\range!
       @buffer\chunk start, stop - 1
 
+  @property current_mode: get: =>
+    return @last_mode if @last_pos and @cursor.pos == @last_pos and @last_mode == @buffer.mode
```
shalabhc: Would this break after buffer modification?

kirbyfan64:
@shalabhc Complaints addressed.

kirbyfan64:
@nilnor Do you want to take another look? My idea for handling the whole config mess is to just keep it as it currently is in Howl and make `word_pattern` a mode setting, not a normal config option (I don't think I've ever had to change that).

shalabhc:
@kirbyfan64 - thanks. Firstly, I'm really happy that this fixes the commenting for nested lexers. But I had a few more thoughts around this CL in general:

1. What is only the right half of a line has a sub-lexer and your cursor is in that segment (is this even possible?). Now if you do a line operation like `toggle-comment` would it use the incorrect comment?

2. For sub-modes, the buffer config is disregarded. Is this always desirable? E.g. if I do `set-for-buffer indent=5` I might expect it to apply even for sub-modes, but it won't.

lib/aullar/spec/styling_spec.moon
```diff
@@ -253,6 +253,24 @@ describe 'Styling', ->
           6, 's1:s2', 7
         }, styling\get(1, 7)
 
+        assert.same nil, styling\get_nearest_style_marker 1
```
nilnor: Don't throw in the new stuff in the middle of an existing spec, but create specific specs for the styling marker stuff instead. These existing specs deal with styling, and not with markers.

lib/howl/buffer.moon
```diff
@@ -267,6 +268,18 @@ class Buffer extends PropertyObject
       b_end_pos = @byte_offset end_pos
       @_buffer\ensure_styled_to pos: b_end_pos
 
+  mode_at: (pos) =>
+    b_pos = @byte_offset pos
+    marker = @_buffer.styling\get_nearest_style_marker b_pos
+    if marker then mode.by_name marker.name else @mode
```
nilnor: There's no guarantee that a mode with the given name exists (rather the opposite), so we need to fall to `@mode` here. Good candidate for a spec.

nilnor:
Started looking (see comments above), will look more later. Looks like it's getting ready! I still like to see some more specs though :) It touches on a lot of places, so some simple verification specs for at least some of these would be nice.

kirbyfan64:
@nilnor I added some more specs. Do you think we could establish what to do about the config stuff now? That's the only next major thing I can think of. My idea is just to make `word_pattern` a normal buffer option.

lib/aullar/spec/styling_spec.moon
```diff
@@ -329,6 +329,49 @@ describe 'Styling', ->
           2, 's1:s3', 3,
         }, styling\get(1, 6)
 
+  describe 'get_nearest_style_marker(pos)', ->
+    it 'generally works', ->
```
nilnor: Instead of "generally works", try to come up with something describing what it actually does when it works. E.g. does it return the nearest preceding marker, or the nearest subsequent marker?

nilnor:
I will (still) look more later - sorry for the slow progress here, we're on sports break here this week. I think @shalabhc raises a good point regarding how it work when only a segment of a line is sub lexed. Currently, it will not be what is expected. A simple heuristic might be to check what the mode is at the beginning of the line instead, at least for some of the operations such as commenting and indenting?

kirbyfan64:
@nilnor Fixed!

kirbyfan64:
To re-mention my idea for a fix for the config issue (I already said this somewhere in the last 20 PRs for this...), it's basically just changing `word_pattern` to a normal buffer mode setting, not a config option. I don't think anyone's ever really had to manually override the word pattern, and that would basically allow us to "ignore" the config issue for now. Because, other than that, this is ready! Finally! After absolutely forever! :)

shalabhc:
I think making `word_pattern` a mode setting instead of a config option is reasonable. I don't really see the need to make override word_pattern provided by a mode for a buffer. But I don't quite remember what config 'issue' is  being discussed.

lib/aullar/styling.moon
```diff
@@ -128,24 +134,34 @@ define_class {
...
           sub_end_offset = get_sub_style_end sub_start_offset, style
           if sub_end_offset -- no sub_end_offset == empty nested styling blocks
             @set sub_start_offset, sub_end_offset - 1, sub_base, no_notify
-            @apply sub_start_offset, style, base: sub_base, no_notify: true
+            @apply sub_start_offset, style, base: sub_base, no_notify: true, :markers
             styled_up_to = sub_end_offset
+            append markers,
+              name: mode_name
+              start_offset: sub_start_offset
+              end_offset: sub_end_offset + 1
```
shalabhc: Since I don't understand the code very well - why is this `sub_end_offset + 1`, but `@set` above uses `sub_end_offset - 1`?

lib/howl/modes/default_mode.moon
```diff
@@ -14,13 +14,13 @@ class DefaultMode
...
 
-  indent: (editor) =>
+  indent: (editor, lines=editor.active_lines) =>
     indent_level = editor.buffer.config.indent
     dont_indent_styles = comment: true, string: true
     buffer = editor.buffer
     current_line = editor.current_line
 
-    editor\transform_active_lines (lines) ->
+    editor\transform_lines lines, ->
```
shalabhc: I didn't get why this and similar changes were necessary.

lib/howl/ui/editor.moon
```diff
@@ -212,6 +212,8 @@ class Editor extends PropertyObject
       start, stop = @selection\range!
       @buffer\chunk start, stop - 1
 
+  @property current_mode: get: => @buffer\mode_at @cursor.pos
```
shalabhc: I think this name is confusing because the main mode is also 'current'. Maybe `mode_at_cursor`?

lib/aullar/styling.moon
```diff
@@ -170,6 +187,10 @@ define_class {
     ptr = @style_buffer\get_ptr(offset - 1, 1)
     style_map[ptr[0]]
 
+  get_nearest_style_marker: (pos) =>
```
shalabhc: I think it may be better if the marker itself is not exposed via the API. Instead this call should just return the sub mode name directly. Also it could probably be renamed to `get_sub_mode_at(pos)` or something. Markers are an implementation detail here IMO.

shalabhc:
OK, looks great overall! I've added a few minor comments.

lib/aullar/styling.moon
kirbyfan64: @shalabhc The end offset is supposed to represent one past the end of the styling ending.

lib/howl/modes/default_mode.moon
kirbyfan64: @shalabhc `transform_active_lines` always transforms *all* selected lines...even if one of them requires using a different sub mode. I changed the editor to call the correct method on the correct line, but that requires passing in the set of lines it should operate on, like here.

lib/howl/ui/editor.moon
kirbyfan64: @shalabhc Fixed.

lib/aullar/styling.moon
kirbyfan64: @shalabhc Fixed.

kirbyfan64:
Oh, wait, I think I had already fixed the config issue. So this is mostly ready!

shalabhc:
@kirbyfan64 - generally this works great! There are some cases with unexpected behavior:

1. Oepn lib/ljglibs/cdefs/cairo.moon
2. Select lines 5 through 12 (selection includes both Moonscrip and C mode).
3. Press `ctrl_slash` to comment lines.

I see it uses Moonscript comments - this is acceptable in this case though the right behavior is debatable.
However, now select lines 5 through 12 again and press `ctrl_slash` again. This uncomments the Moonscript lines, but double comments the C lines. I would have expected all the lines to get uncommented. After this is it impossible to fully uncomment the C lines using the toggle-comment comment.

There may not be an obviously correct behavior for mixed mode selections, but I think it should be well defined and `toggle-comment` should generally inverse it's behavior. You may want to add tests for this too.

lib/aullar/styling.moon
shalabhc: Right, but `@set` uses `sub_end_offset - 1` above so the of the end of marker is off by 2 from the end of the set region? Not sure if this has practical implications, but maybe the first character in the token closing the sub-mode will be considered part of the sub-mode?

kirbyfan64: @shalabhc I added a comment describing it.

kirbyfan64:
@shalabhc I can't reproduce the exact same issue. What happens for me is that, when I run `toggle-comment` the second time, the MoonScript code gets uncommented, but the C code gets commented with `/* */`.

The reason lies in `_apply_to_line_modes`, and I'm not sure it's a fixable problem. When it uncomments the MoonScript lines, then the sub mode for the C code lines changes to `c`. Therefore, it calls the C mode's `toggle_comment`. But those lines had been commented via the MoonScript comment syntax, so it ends up double-commenting them.

As for why it's unfixable, the only way I know to "fix" it is to get all the modes *ahead* of time, not while the lines are being processed. Problem is, now, for the first `toggle_comment`, the C lines get commented via the C mode. However, when you run `toggle_comment` *again*, then the lines all use the MoonScript `toggle_comment`, so they end up double-commented. Same problem again.

I think the current behavior is probably the least surprising. It's all going to be surprising regardless, but the only alternative is worse for the first case and just as bad for the second.

kirbyfan64:
Oh yeah, and I cherry-picked this entire PR, so now there are no conflicts. :D

shalabhc:
@kirbyfan64 - yes that's what I see, maybe I didn't describe it properly the first time.

What if the commenting worked top down but the uncommenting worked backwards - starting at the last line and moving up. Then when uncommenting the C code would have the moonscript mode because of the `[[` still being commented out. Maybe that is slightly better than now?

kirbyfan64:
@shalabhc The problem is with `toggle_comment`. It's the mode that determines whether or not to comment or uncomment. So you'd need to know which mode to call *before* calling `toggle_comment`. But only `toggle_comment` could say whether to comment or uncomment, which is in turn needed to know which mode to call!

Maybe we could gut `toggle_comment` completely. Instead, the mode would have a method `is_comment` that says whether or not the line is a comment. Then the editor itself could move from there and determine what to do.

kirbyfan64:
Also, I'm not going to be able to work on this at *all* in a few days for the following two weeks, so it would be great if we could figure this out before then!

shalabhc:
@kirbyfan64 - OK I'm thinking of something reasonable but simple now. My main issue is that 'toggle-comment' does not self-revert. How about this: change `_apply_to_line_modes` to only call the line mode's method if the active lines all have the same mode. Otherwise have it call the buffer mode's method. So for mixed mode selections, it always uses the buffer mode comment/uncomment styles and for single mode selections it uses the line's mode.

This is only useful for commenting, not indenting so this can be optional, e.g. to `_apply_to_line_modes('toggle_comment', single_mode_only: true)`. What do you think?

kirbyfan64:
@shalabhc What about a little twist on that, where the mode of the *first* line is used instead? That way it would work better with things like the Markdown mode.

shalabhc:
The problem is if your selection includes a cdef section followed by `]]` and includes some moon code, you will end up with C comments in moon code. My suggestion is essentially default to the old behavior in mixed modes because it is unclear what the correct behavior should be.

Can you give an example of how it is better for Markdown?

kirbyfan64:
@shalabhc Hmmm...

When I said it was better for Markdown, that's because Markdown has no comments, so, in a case similar to above, but where the MoonScript code is within a Markdown document, nothing would end up happening.

But I guess then using the old behavior is the way to go!

kirbyfan64:
@shalabhc Done!

shalabhc:
@kirbyfan64 - thanks - works much better!

I found another issue though. Undo doesn't work as expected. In master if I select a block, toggle-comment and then undo - all comments are undone in one shot and the selection is restored. In this branch undo happens one line at a time.

kirbyfan64:
@shalabhc Fixed.

shalabhc:
It works well now. I think this is ready to merge - @nilnor ?

kirbyfan64:
Bump?

nilnor:
@kirbyfan64 as we already said over email previously, this is too large to go in before 0.4, and that's not cut yet. It's not far off, but I'm doing a blog post on aullar first to go with the release.

kirbyfan64:
Ah, ok.

lib/aullar/styling.moon
```diff
@@ -4,6 +4,7 @@
 {:fill} = require 'ffi'
 {:define_class} = require 'aullar.util'
 GapBuffer = require 'aullar.gap_buffer'
+Markers = require 'aullar.markers'
```
nilnor: I'm missing things from the implementation here - `set` and `insert` does not modify the sub style markers at all, which they reasonably should (since the build is green, specs are probably a good idea for these cases).

lib/aullar/styling.moon
```diff
@@ -157,6 +180,7 @@ define_class {
 
   delete: (offset, count, opts = {}) =>
     @style_buffer\delete offset - 1, count
+    @sub_style_markers\remove_for_range offset, offset + count
```
nilnor: This is not what you want here - this actually shrinks the buffer, but you now only remove some markers and leave others with incorrect offsets. You want to `shrink` the markers.

lib/aullar/styling.moon
kirbyfan64: @nilnor Ah, right, I wanted to ask: is there a reason for `set` to be public? I couldn't find any uses outside of `apply`, and that adds the markers itself more efficiently (it collects them all together and batch-updates them). Could `set` be made into `_set`?

kirbyfan64:
@nilnor Everything fixed except for what I mentioned in the comment.

lib/aullar/styling.moon
nilnor: It's primarily used in the specs, but you missed one usage in `ActionBuffer`.

lib/aullar/styling.moon
```diff
@@ -170,6 +196,10 @@ define_class {
     ptr = @style_buffer\get_ptr(offset - 1, 1)
     style_map[ptr[0]]
 
+  get_mode_name_at: (pos) =>
+    found = @sub_style_markers\at pos
+    found and found[#found] and found[#found].name
```
nilnor: `found` will never be `nil` here by the way, no markers found will result in an empty table.

lib/aullar/styling.moon
kirbyfan64: @nilnor Ah, ok. Fixed.

lib/aullar/styling.moon
kirbyfan64: @nilnor Fixed.

lib/howl/buffer.moon
```diff
@@ -277,6 +278,19 @@ class Buffer extends PropertyObject
       b_end_pos = @byte_offset end_pos
       @_buffer\ensure_styled_to pos: b_end_pos
 
+  mode_at: (pos) =>
```
nilnor: Regarding these two `{config,mode}_at`, I see that some code has been updated to use them, and others not. Is there a reason for that?

lib/howl/modes/default_mode.moon
nilnor: I don't like these changes, or the corresponding ones in Editor. They're invasive, makes things harder to reason about, and I don't really see that they're contributing much to warrant that. I think perhaps these made more sense in earlier revisions, but given the heuristic of using the mode for the first line for mixed selections (which I approve of), then this only seems complicated. It seems like reverting these changes and, for the operations that require a mode, using the mode of the first line would be fine and fully within reasonable expectations. I may of course be overlooking some crucial here, but it seems like this would work fine.

lib/howl/buffer.moon
```diff
@@ -277,6 +278,19 @@ class Buffer extends PropertyObject
...
+    b_pos = @byte_offset pos
+    mode_name = @_buffer.styling\get_mode_name_at b_pos
+    -- Returns @mode if there's no marker or the requested mode doesn't exist.
+    mode_name and mode.by_name(mode_name) or @mode
+
+  config_at: (pos) =>
+    mode_at = @mode_at pos
+    return @config if mode_at == @mode
+    new_config = config.local_proxy!
```
nilnor: So this is something I'm not sure about. As @shalabhc mentioned in an earlier comment this loses any buffer specific configuration when in a sub mode. Is this reasonable, or would one expect something else? Like I said I'm not sure - perhaps this is the way it should work since buffer specific setting were probably set considering the parent mode in the first place, and it might be preferable to use the mode's config directly. To prevent any confusion one could in that case return a read-only proxy, or otherwise it would possibly to return a new read-only chained config respecting both local and sub mode settings. That we could address post-merge however, but we need to figure out what the expected behaviour would be.

nilnor:
Ok, so I'm done! It's been a while in the making, but it will be great to get this in - the commenting in sub modes in particular is a continuous pain. Nice work!

Final  things to address:

- Have a look at existing usages of `Buffer.{config,mode}`, and see whether they should be replaced to work correctly for sub modes. An aside, should `Buffer.{config,mode}` be directly accessible at all?

- The changes for Editor and DefaultMode, revert and simplify if possible (prev. discussion)

- There's no documentation for the new `config_at` and `mode_at`. It should be documented, and if there's a reason for choosing one version over another it needs to be documented (see first point)

And there's the question of how the sub mode config should work, which should be decided upon.

lib/howl/buffer.moon
kirbyfan64: @nilnor The only thing it's used for is `word_pattern`. My proposed fix was to just change `word_pattern` to be a buffer setting, not a config option, since I can't think of a single use case for actually changing that.

lib/howl/buffer.moon
kirbyfan64: Well, my search over the code base may have missed some uses of `buffer.mode`...could you point out where exactly you're referring to?

lib/howl/modes/default_mode.moon
kirbyfan64: @nilnor 

>  but given the heuristic of using the mode for the first line for mixed selections (which I approve of), then this only seems complicated.

Good point. This was for the original version. I'll try to revert this.

kirbyfan64:
@nilnor Man, this is so close! :D

Based on what you said:

1. There are a couple of uses of `buffer.mode`, in particular by `Application` and `Editor`, that can't be cleanly removed. The `Application` uses would partly end up being replaced by something weird like `buffer.mode_at(0)`, and the use in `Editor` (in `_apply_to_line_modes`, which was added in this PR) doesn't really have a suitable equivalent. It's also the only way to *set* a mode, and it's used like heck in the specs.

2. See my above comments for getting rid of `config_at`. TL;DR it only exists for `word_pattern`, which should probably be a normal mode setting anyway. Other than that, it's basically useless.

kirbyfan64:
Linters hate me... It says:

```
line 1: assigned but unused `line`
```

when linting `lib/howl/ui/editor.moon`. Which is weird, because I just checked, and there are no unused `line` variables, let alone on the first line, which is a comment...

kirbyfan64:
Ok; I updated the docs. After dealing with all sorts of Gem-, bundle-, and stupid-Ruby-packaging-related crud...

...I am never, ever, *ever* complaining about [pip](https://pip.pypa.io/en/stable/) again.

lib/howl/ui/editor.moon
```diff
@@ -647,6 +647,21 @@ class Editor extends PropertyObject
...
+    mode = nil
+    modes = [@buffer\mode_at line.start_pos for line in *lines]
+    mode = modes[1]
+    for other_mode in *modes
+      if mode != other_mode
+        mode = @buffer.mode
+        break
+
+    @buffer\as_one_undo ->
+      for line in *lines
```
shalabhc: @kirbyfan64 maybe this block is the lint culprit?

nilnor:
Note that the other travis run shows a broken spec ("Editor / toggle_comment() / uses the buffer mode when the one to use is ambiguous").

lib/howl/buffer.moon
nilnor: Looking now it seems `buffer.mode` is clear, but `buffer.config` still shows occurrences that just uses the plain `buffer.config`. E.g. from doing an `ag 'buffer\.config'`:

- text.moon
- auto_pair.moon
- formatting.moon
- [..]

lib/howl/buffer.moon
nilnor: We could change the `word_pattern` to a mode field instead, but there are still other settings that at least should be used (e.g. indentation, tab width, etc).

lib/howl/ui/editor.moon
```diff
@@ -279,11 +281,10 @@ class Editor extends PropertyObject
     }
 
   transform_active_lines: (f) =>
-    lines = @active_lines
```
nilnor: This actually looks like an earlier oversight - it would be better to use this local var instead of refetching the active lines all the time.

lib/howl/ui/editor.moon
nilnor: Yes, this probably is the source of the linting problem (but I also think that switching to another branch where the config ends up missing might also cause problems). However, the code is wrong anyway - it shouldn't loop over the lines but instead do the op just once, and the `as_one_undo` is not needed.

kirbyfan64:
Linter's happy now! :heart:

nilnor:
@kirbyfan64 did you have a look at the usages of `buffer.config` I commented on earlier in the diff? That still seems unresolved.

kirbyfan64:
@nilnor And, as I also said earlier in the diff, the only reason for `config_at`'s existence is for `word_pattern`, which should probably be a buffer setting anyway. :)

nilnor:
@kirbyfan64: right, but I replied to that before as well. I think other place still needs to use `config_at` for sub modes to really work out well? E.g. in `auto_pair.moon`, it's fetching the auto pairs from the sub mode, but it's checking whether auto pairs are active using the parent buffer's config.

kirbyfan64:
Ok...

So, due to a giant tree of entirely unexpected circumstances (need to use Windows for college work, death of grandfather, *and more*), I'm basically completely locked out of the Linux side of my computer and will be unable to fix that for, at minimum, a week. As a result, unless I manage to get Howl working on Windows (although I'm pretty desperate, so who knows...), I won't be able to finish this for another while. Side note:

@nilnor 

> right, but I replied to that before as well. I think other place still needs to use `config_at` for sub modes to really work out well? E.g. in `auto_pair.moon`, it's fetching the auto pairs from the sub mode, but it's checking whether auto pairs are active using the parent buffer's config.

I don't really know why a buffer would define `auto_pairs` if auto pairs are disabled, though... AFAIK, the only use case would be a user setting it manually in their `init.moon` or in the editor.

kirbyfan64:
@nilnor Updated.

nilnor:
It's been a while in the making, but I think that the magical merge time is now here! @kirbyfan64, has some merge conflicts now, do you want to do the honours of merging this correctly? :)

kirbyfan64:
@nilnor [VICTORY!!!](https://youtu.be/x3m394DZUdg)

shalabhc:
Congratulations on this one!!

Issue #155, opened 2016-01-13
Insertion into read-only buffer works


kirbyfan64:
Basically, I noticed that, if you run an external command with Howl, you can insert text into the `ProcessBuffer`, even though it's read-only. Interestingly, you CANNOT replace text...because it's read-only!

This seems to be because [`howl.Editor.insert`](https://github.com/howl-editor/howl/blob/master/lib/howl/ui/editor.moon#L379) calls [`aullar.View.insert`](https://github.com/howl-editor/howl/blob/master/lib/aullar/view.moon#L310), which always calls [`aullar.Buffer.insert`](https://github.com/howl-editor/howl/blob/master/lib/aullar/buffer.moon#L157) instead of [`howl.Buffer.insert`](https://github.com/howl-editor/howl/blob/master/lib/howl/buffer.moon#L152). This skips the [check to make sure that the buffer is writable](https://github.com/howl-editor/howl/blob/master/lib/howl/buffer.moon#L296).

nilnor:
After looking at it the problem is not that the `Editor.insert` does not check the status (which is also a problem), but that the key presses and corresponding inserts are handled by the aullar view. So, we have to move down some read-only support into aullar, or add some kind of veto-notification thing.

Issue #144, opened 2015-11-29
Allow completers to select text


refi64:
In `ui/completion_popup.moon`, there's this line:

``` moonscript
@editor.cursor.pos = @completer\accept item, @editor.cursor.pos
```

However, this makes it impossible for a completer to do stuff like select the inserted text.

My idea is to change that to:

``` moonscript
old_persistent = @editor.cursor.selection.persistent
@editor.cursor.selection.persistent = true
@editor.cursor.pos = @completer\accept item, @editor.cursor.pos
@editor.cursor.selection.persistent = old_persistent
```

(Of course, with proper error handling and stuff.) Thoughts?

nilnor:
The current structure of this is that completers only complete, they are not expected to do anything else. To better understand why you want this - what is the use case you're thinking of here?

kirbyfan64:
@nilnor In [howl-autoclang](https://github.com/kirbyfan64/howl-autoclang), placeholder text is inserted in `finish_completion`. For instance, if someone selected `puts` from the completion list, the text `(const char* something)` or similar will be inserted afterwards. Although this is useful, it's also slightly irritating to always have to delete the text inside the parens. So I figured I could select the text inside automatically, that way a simple backspace will get rid of it, but it's still there.

shalabhc:
This kind of feature sounds awesome and I'm excited to see C/C++ autocompletion underway :+1: 

But is selection the right thing for this? Presumably one could insert some text with multiple placeholders that the user replaces one-by-one, by tabbing through them, for example. Also it might be useful to let the user replace the placeholder at a later time, something a selection won't do.

nilnor:
Yes, I see the use for this. As Shalabh is suggesting, I don't think the selection is necessarily the best option for handling this. And even if is, I rather introduce an explicit way to assign this kind of information with a completion item itself, rather than having a completer do any modification itself. 

I was also going to bring up the introduction of placeholders - with the advent of aullar and buffer markers I think we now have a pretty good base to build on for creating a more general snippet expansion functionality. And this seems like it could be one such case.

shalabhc:
@kirbyfan64 - see if the table-completions branch helps you a litte bit here. Specifically instead of returning completions as `{'text-to-insert1', 'text-to-insert2'}`, you can now return a list of tables like `{ {'text-to-display1',  completion: 'text-to-insert1'}, ...}` etc. You can even have multiple columns in the table if that helps with formatting.

nilnor:
[...]

PR #175, opened 2016-02-28
Go mode

barnybug:
go bundle adding some go support using the new lexer, smart completion and automatic code reformatting using gofmt.


bundles/go/go_completer.moon
```diff
@@ -0,0 +1,52 @@
...
+    read_stdout: true,
+    read_stderr: true,
+  }
+
+  buffer = ProcessBuffer p
+  editor = app\add_buffer buffer
+  editor.cursor\eof!
+  buffer\pump!
+  
+completions_enabled = true
```
kirbyfan64: I feel like this should be a configuration option.

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,85 @@
...
+    char_constant,
+    float,
+    hexadecimal_float,
+    hexadecimal,
+    octal,
+    R'19' * digit^0,
+  }
+
+  string = c 'string', any {
+    span("'", "'", '\\')
```
kirbyfan64: Won't this conflict with `char_constant`?

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,85 @@
...
+  fdecl = any {
+    sequence {
+      c 'keyword', 'func'
+      ws^1
+      c 'fdecl', ident
+    }
+    sequence {
+      c 'keyword', 'func'
+      ws^1
+      span '(', ')'
```
kirbyfan64: Does this highlight the stars in pointers? Like `func (MyType*) f`. Also, I think this and the previous `sequence` can be put in the same one, like `span('(', ')')^-1`.

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,85 @@
...
+    hexadecimal_float,
+    hexadecimal,
+    octal,
+    R'19' * digit^0,
+  }
+
+  string = c 'string', any {
+    span("'", "'", '\\')
+    span('"', '"', '\\')
+    span('`', '`', nil)
```
kirbyfan64: In Lua/MoonScript, arguments that aren't given are defaulted to `nil`, so you can omit the extra `nil` argument altogether.

bundles/go/go_lexer.moon
barnybug: Fixed

bundles/go/go_completer.moon
barnybug: Good idea, added

bundles/go/go_lexer.moon
barnybug: No the receiver part wouldn't be highlighted.
Simplified fdecl.

bundles/go/go_lexer.moon
barnybug: Sure - it just read better to make it explicit that there are no escape chars in raw strings.

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,76 @@
...
+  
+  builtin = c 'function', word {
+    'append', 'cap', 'close', 'complex', 'copy', 'delete', 'imag', 'len', 'make',
+    'new', 'panic', 'print', 'println', 'real', 'recover'
+  }
+  
+  fdecl = sequence {
+    c 'keyword', 'func'
+    ws^1
+    (span '(', ')' * ws^1) ^ -1
```
kirbyfan64: To get this to highlight correctly, I you can change the whole thing to:

```moonscript
fdecl = sequence {
  c 'keyword', 'func'
  any {
    sequence({
      ws^0
      c 'operator', '('
      (V'all' + P 1 + ws - P')')^0
      c 'operator', ')'
      ws^0
    })
    ws^1
  }
  c 'fdecl', ident
}
```

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,76 @@
...
+    octal,
+    R'19' * digit^0,
+  }
+
+  string = c 'string', any {
+    span('"', '"', '\\')
+    span('`', '`', nil)
+  }
+
+  constant = c 'constant', word { 'true', 'false', 'iota', 'nil' }
```
kirbyfan64: Should these be highlighted as `special` instead? Usually, `constant` seems to be used for user-defined constants. I may be wrong, though...

barnybug: Yeah see what you mean. The golang language def calls them constants (https://golang.org/ref/spec#Predeclared_identifiers), but I agree that doesn't match up with the style naming.

bundles/go/go_lexer.moon
barnybug: Couldn't get the first any part to match (example: "func (self *A) Method() {"), without removing the V'all'

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,58 @@
...
+  append args, file
+  success, out, err, p = pcall Process.execute, args
+  unless success
+    log.error out
+    return
+  unless p.successful
+    log.error err
+    return
+
+reload_buffer = (buffer) ->
```
shalabhc: `buffer` already implements `buffer\reload` - would that work instead of this?

barnybug: That assigns @file, which blitzes the undo history

shalabhc: I see. Could the undo preserving reload be implemented by `buffer\reload` as an option?

bundles/go/init.moon
```diff
@@ -0,0 +1,52 @@
...
+with config
+  .define
+    name: 'go_fmt_command'
+    description: 'Command to run for go-fmt'
+    default: 'gofmt -w'
+    scope: 'global'
+
+  .define
+    name: 'go_complete'
+    description: 'Whether to use go completions in go mode'
```
shalabhc: Would be useful to include 'requires gocode' here.

bundles/go/go_fmt.moon
barnybug: Something fundamental would have to change around the @file setter I guess?
Felt look changing too much core stuff just for this mode!

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,58 @@
...
+    else
+      new_pos += 1
+  new_pos-1
+
+fmt = (buffer) ->
+  return unless buffer.mode.name == 'go'
+
+  file = buffer.file
+  before = file.contents
+  run_command file
```
shalabhc: Will this lose unsaved changes? If so, it should probably warn or ask to save file first. Alternatively, can it just run on the buffer text instead of the saved file? A generic function for the latter functionality could be very useful in other bundles too.

bundles/go/go_fmt.moon
shalabhc: The `@file` setter doesn't have to change, just the implementation of  `buffer\reload`. E.g. it could just do what is being done here, if  `{preserve_undo: true}` is passed in. This seems like a case where improving the underlying API will have future benefit.

I also wonder if undo preserving reload is should be the default, but that requires a wider conversation.

shalabhc:
Very happy to see another proper language support, with its own auto-complete!

bundles/go/init.moon
barnybug: It doesn't, since it offers to install if not present. If it said 'requiries gocode' I'd assume that required me to go off and install it manually.

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,58 @@
...
+      i += 1
+      new_pos += 1
+    elseif before[i]\match '%s'
+      i += 1
+    else
+      new_pos += 1
+  new_pos-1
+
+fmt = (buffer) ->
+  return unless buffer.mode.name == 'go'
```
shalabhc: Minor nit, but I feel this check should go in the command handler (and also log something so user is not confused if the command did nothing).

bundles/go/go_fmt.moon
barnybug: Good point, hadn't considered whether gofmt could even operate just on stdin/stdout. Looking into it, it can, so I'll change this to be more flexibly work 'sans-files'.

So this means adding a pre-save signal in order to run before saving happens.

bundles/go/go_fmt.moon
barnybug: This one will go away, changing to work from stdin/out as suggested below

bundles/go/go_fmt.moon
barnybug: Ack, so having troubles doing it this way. Because the signal emit happens asynchronously it completes before calling the external process. Which means it just modifies the buffer text after the save has happened.

Seems like the 'pre-save' hook would have to be able to block the save.

Not sure how best to proceed, ideas?

barnybug: Figured out a route - see ab48d6

bundles/go/go_fmt.moon
barnybug: agree https://github.com/howl-editor/howl/commit/d49cb3d99d99c7e4fc602389ae543c04d64755e6

bundles/go/go_fmt.moon
shalabhc: Ah, I see. From the code it looks like signals run synchronously only until the first coroutine yield happens (which would happen when running an external process). Did the earlier implementation have a race condition too? Say if you start typing immediately after 'save' and 'gofmt' takes longer to run, the buffer contents would be switched out, discarding your edits?

A quick solution would be run goftm on 'buffer-saved' and immediately save it again. This of course has the same race as above.

Given we might want to do similar transforms on save for other modes, maybe a better solution is warranted. @nilnor - any ideas?

bundles/go/init.moon
shalabhc: Maybe then 'based on gocode'? I'm assuming there are multiple completion engines out there and having this info in the description would be useful to anyone just browsing the config.

barnybug: description adjusted

bundles/go/go_fmt.moon
barnybug: I tried setting buffer.read_only = true around the gofmt running (with my own very slow gofmt (sleep 5 :-) ), but this doesn't appear to prevent typing into the buffer.
Should read_only=true prevent typing?

Worst case I think this is very unlikely to happen - time reports gofmt taking 0.00s consistently for me :-)

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,84 @@
...
+      fdecl,
+      type,
+      keyword,
+      builtin,
+      operator,
+      number,
+      constant,
+      identifier,
+    }
+  }
```
kirbyfan64: Would it be possible to highlight type definitions? Something like:

```moonscript
type_def = sequence {
  c 'keyword', 'type'
  ws^1
  c 'type_def', ident
}
```

Or are there other ways to define types in Go?

bundles/go/go_lexer.moon
```diff
@@ -0,0 +1,84 @@
...
+    'new', 'panic', 'print', 'println', 'real', 'recover'
+  }
+
+  fdecl = sequence {
+    c 'keyword', 'func'
+    any {
+      sequence({
+        ws^0
+        c 'operator', '('
+        (P 1 + ws - P')')^0
```
kirbyfan64: @barnybug  Sorry, I'm an idiot. :/ The reason the version I gave didn't work was because this line is supposed to be `((V'all' + P 1 + ws) - P')')^0` (note the extra parens).

barnybug: So this would lex everything inside as any symbol? Not sure this makes
sense.

kirbyfan64: @barnybug You can try it out yourself. It's basically like `span`, but it correctly highlights everything inside. That way, with stuff like `func (*MyCoolType) MyCoolFunctionMethodThingy()`, the `*MyCoolType` will get colored normally (so the `*` will be highlighted as an operator).

barnybug: ok, I guess it doesn't matter to much, but if I write:
func (copy Type) Method() {
the receiver name 'copy' is highlighted incorrectly as a builtin, because we're lexing anything inside the receiver parens with 'all'.

kirbyfan64: @barnybug What would that code normally do?

barnybug: It would define 'copy' as the name of the parameter receiving 'self', overriding the builtin at the same time. (It's valid compilable code, but you'd probably need a head examination if you were writing it!)

Can't see any real problem using 'all' though, so will go with that.

bundles/go/go_fmt.moon
shalabhc: read_only = true should prevent typing but see #155 

BTW, I feel saving is such a critical operation that having it depend on an external command is not ideal. In the rare case where `gofmt` breaks or the `Process` machinery errors out, the save could be cancelled and maybe even prevent saving permanently. What do you think about always saving first and then applying 'gofmt` to save again?

Even if in this case it is super reliable, this introduces a pattern that may break if another bundle adds a less reliable `before_save`. I'd prefer to see a very simple and predictable code path between pressing `ctrl_s` and the actual save.

bundles/go/go_lexer.moon
barnybug: You can define multiple ones at once like so:
type (
    A int
    B struct { ... }
)

But I got stumped trying to lex this too:

  type_def = sequence {
    c 'keyword', 'type'
    any {
      sequence {
        ws^1
        c 'type_def', ident
      }
      sequence {
        ws^0
        c 'operator', '('
        ws^0
        c 'type_def', ident
        sequence({
          ((P 1) - P'\n')^0
          P'\n'
          ws^0
          c 'type_def', ident
          ((P 1) - P'\n')^0
        })^0
        ws^0
        c 'operator', ')'
      }
    }
  }

bundles/go/go_fmt.moon
barnybug: Have to say I'm not dead keen on double saving. What are others thoughts? @nilnor

barnybug:
sure, added

nilnor:
One thing all new language bundles should have is a comprehensive example file with examples of all syntax that the lexer handles. This is a great help when building the lexer (split windows and the `bundle-reload` command then gives rapid feedback for the lexer changes), and also later when changing it to make sure any changes work and does not break existing lexing. It also provides a nice overview of what the lexer handles.

bundles/go/go_completer.moon
```diff
@@ -0,0 +1,50 @@
+import app, config, interact from howl
```
nilnor: Don't forget to add the copyright stanzas for new files.

bundles/go/go_completer.moon
```diff
@@ -0,0 +1,50 @@
...
+import ProcessBuffer from howl.ui
+import Process from howl.io
+import Matcher from howl.util
+append = table.insert
+
+install_gocode = ->
```
nilnor: Instead of doing lifting ourselves here we could re-use the command module. E.g.:

```moonscript
howl.command.exec nil, 'go get -u -v github.com/nsf/gocode'
```
The first argument is the working directory, so we could the current file's directory if we want to.

bundles/go/go_completer.moon
```diff
@@ -0,0 +1,50 @@
...
+  buffer = ProcessBuffer p
+  editor = app\add_buffer buffer
+  editor.cursor\eof!
+  buffer\pump!
+  
+run_gocode = (context) ->
+  cmd = { 'gocode', '-f=csv', 'autocomplete', context.pos-1 }
+  status, out, err, process = pcall Process.execute, cmd, stdin: context.buffer.text
+  unless status
+    error out unless out\match 'No such file'
```
nilnor: Instead of attempting to run this and then checking the output, `howl.sys.find_executable` could be used to check for the existence of `gocode`.

bundles/go/go_completer.moon
```diff
@@ -0,0 +1,50 @@
...
+    read_stderr: true,
+  }
+
+  buffer = ProcessBuffer p
+  editor = app\add_buffer buffer
+  editor.cursor\eof!
+  buffer\pump!
+  
+run_gocode = (context) ->
+  cmd = { 'gocode', '-f=csv', 'autocomplete', context.pos-1 }
```
nilnor: So it's been a while since I last had a look at Go, but can we actually rely on `gocode` being available in `$PATH`? It will be installed into `$GOPATH/bin` by the install command, so perhaps that's what we should try to run instead?

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,54 @@
...
+
+append = table.insert
+
+run_command = (contents) ->
+  args = {}
+  for arg in config.go_fmt_command\gmatch '%S+'
+    append args, arg
+  args.stdin = contents
+  success, out, err, p = pcall Process.execute, args, stdin: contents
+  unless success
```
nilnor: The two `unless`:s can be combined.

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,54 @@
...
+fmt = (buffer) ->
+  log.info "Running #{config.go_fmt_command}..."
+  before = buffer.text
+  after = run_command before
+  log.info "#{config.go_fmt_command} completed"
+  return if not after or after == before
+
+  editor = app\editor_for_buffer buffer
+  if editor
+    -- reload the contents, adjusting position
```
nilnor: It seems like `Editor:with_position_restored` could be used instead of doing our own custom thing here for the positioning?

nilnor:
Given that this is Go, with its strict indentation and formatting policies, I think it makes sense to include a default mode configuration that sets this up by default (i.e. `use_tabs = true`, `tab_width = 4`, `indent = 4`). That way one does not have to fight with `gofmt` unnecessarily or worry about configuring stuff if you want to follow the go conventions.

In general, I'll echo what @shalabhc said, very nice to see a proper language bundle for this! Will be great to get this in :)

bundles/go/go_completer.moon
shalabhc: Ah I forgot about `command.exec`. I'll also add that the `exec` command can let the user modify the 'go get' command if desired. To do that, it needs to be invoked this way:

```
howl.command.run 'exec go get -u -v github.com/nsf/gocode'
```

bundles/go/go_completer.moon
shalabhc: Or even make it configurable via `config.gocode_path`? So this would let you customize on a per-file basis (and per-project, when that's implemented)

bundles/go/go_fmt.moon
shalabhc: Wasn't it double saving earlier too though? :wink: (Except the second save was outside howl.)

nilnor: Well I agree that it of course need to be very stable. Either it's a policy that a `before_save` handler need to always handle all errors and avoid destroying buffer contents, or we could of course also add guards for the hook invocation. The latter could be somewhat complicated, so I'm thinking that for now we can start with the optimistic approach of thinking the best of the hook handlers.

shalabhc: OK. Can we `pcall` protect the `before_save` call? Then we can worry about hangs and slowness later, but errors at least will not cancel the save.

shalabhc:
Tangential, but interesting: https://github.com/Microsoft/vscode-go

bundles/go/go_fmt.moon
nilnor: We could, but note that this would not be 100% either, since the hook might possibly have modified the buffer before crashing. A more reliable guard could make sure to automatically undo any changes for a failing handler as well. More complicated though, and I'm not sure we need to that in this PR.

shalabhc: Good point. I don't think that needs to be done in this PR. Maybe we need a transnational update method e.g. `buffer\transactional_update -> muck(buffer)` - clearly a separate PR. It also ties in a bit with the auto saved revisions that has come up before.

bundles/go/go_completer.moon
barnybug: done

bundles/go/go_completer.moon
barnybug: done

bundles/go/go_completer.moon
barnybug: Yep, GOPATH is the standard place it would be - changed.

bundles/go/go_fmt.moon
barnybug: fixed

bundles/go/go_fmt.moon
barnybug: I don't think this would adjust for the whitespace changes that go fmt makes?

barnybug:
I've caught up the changes with the comments, so hopefully this is good to go now (optimistic :-) )!

barnybug:
re: vscode-go - looks nice. There's a great amount of tooling now for go that we could consider hooking in.
Also, as an enhancement to completions, it'd be great to allow associated hint text and styling to the entries. gocode returns a load of extra info (symbol type, function signature), that would be good to show alongside. One for the future..

shalabhc:
Good idea. I think completions could be made to support tables and/or styled text objects to enable  hints etc. in the completion drop down. Another related feature is inserting text with replaceable placeholders - #144.

barnybug:
One remaining thing troubling me is how to best to show formatter reported errors to the user when saving. These currently get clobbered by the "file.go: n lines..." info message in the save command.

go fmt produces helpful information about syntax errors, etc. to stderr which I'd like the user to see, even though the save succeeds.

Is there a different/better way to show the error than log.error?
Should the info message be changed when saving was 'partially' successful?
What would be the best way to go about this?

shalabhc:
Could you provide an example of a log message you would like to show? E.g. how long is a typical message and does it include one or more references to lines in the file?

Maybe the logger should accumulate messages and not hide any until it ensures that they were displayed once (or for x amount of time).

I use a custom solution for linter style output that includes references to line numbers. It runs a linter on every save and attaches markers to the buffer lines that produces warnings. These markers use a flair for visual indication. I can then navigate to any of these lines and press 'ctrl_i' to display the message for that line in a popup. The message is stored internally in the marker itself. Alternatively, I can press 'ctrl_shift_i' to show all messages for the current buffer in a buffer-grep/buffer-structure style list view - letting me jump to any location. Reruns of the linter clear the existing markers, of course, before adding new ones. We could standardize on something like this for linter/compiler warnings and errors. What do you think, @nilnor.

shalabhc:
Wanted to clarify that I can't think of anything that would immediately solve the messaging on save issue. My suggestions are a couple of PRs away, of course.

barnybug:
They pretty succinct one liners - some examples:
<standard input>:3:8: expected 'IDENT', found '{'
<standard input>:4:1: expected '(', found '}'
<standard input>:6:2: missing ',' before newline in parameter list

Crucially you always get line/col information.
Your linter marker functionality sounds like exactly the ticket - would certainly be keen to add go support for something standardized.

bundles/go/go_fmt.moon
nilnor: I believe it will (or should). Have you tried it?

nilnor:
Re the linting/messages support, having something standardized for supporting that is at the top of my prio list (post 0.4). Let's leave that for now and revisit once we have it in place.

bundles/go/go_fmt.moon
barnybug: Yes, it doesn't. It just maintains the same line and column position.

bundles/go/go_mode.moon
```diff
@@ -0,0 +1,35 @@
...
+  
+  default_config:
+    use_tabs: true
+    tab_width: 4
+    indent: 4
+  
+  lexer: bundle_load('go_lexer')
+
+  structure: (editor) =>
+    [l for l in *editor.buffer.lines when l\match('^%s*func%s') or l\match('^%s*struct%s') or l\match('^%s*type%s')]
```
nilnor: Nit-picking: This line ends up being very long. Could be rewritten using regex with alternation as one way of getting it shorter, e.g. `r'^\\s*(?:func|struct|type)\\s`

bundles/go/go_fmt.moon
barnybug: eg, to illustrate gofmt this code, with the cursor at [^]:
```
package main

import "fmt"


func    [^]main() {
	fmt.Println("Hello, world!")
}
```
becomes:
```
package main

import "fmt"

func main() {
	fmt.[^]Println("Hello, world!")
}
```
That's where the cursor ends up for with_position_restored - it's only adjusting for indent changes.

barnybug: That's to say, I don't think with_position_restored should necessarily try to do this - the fact gofmt is just changing whitespace (and the trick calculate_new_pos uses to discount this the whitespace change) wouldn't generalise to any other changes in characters. So bit too much of an assumption generally.

nilnor:
All right, almost ready to go in :)

As the final thing we should sort out the cursor position thing. It might be just switching to `with_position_restored` would work, in which case we're fine. We might also have to tweak it if it isn't smart enough. In any case it's preferable to use, and possibly supplement, that instead of a custom written helper without specs inside a bundle.

bundles/go/go_fmt.moon
nilnor: All right, fair enough. Should we ever need something like this anywhere else we can look at unifying it then.

bundles/go/go_fmt.moon
```diff
@@ -0,0 +1,65 @@
...
+  log.info "Running #{config.go_fmt_command}..."
+  before = buffer.text
+  buffer.read_only = true
+  success, result = run_command before
+  buffer.read_only = false
+  unless success
+    log.error "#{config.go_fmt_command} error: #{result}"
+    return
+  log.info "#{config.go_fmt_command} completed"
+  return if result == before
```
nilnor: This does a string comparison, perhaps just checking the sizes would suffice?

nilnor:
Good work @barnybug! I'm merging this now - further tweaking, if any, can be done in master.

PR #176, opened 2016-03-01
Change matching brace highlight behavior a bit

shalabhc:
- prefer to match brace before cursor rather than after, when both
present
- highlight both braces of the matched pair

## Reasoning
(`|` represents cursor, `^` represent hightlight)

### Old behavior

When typing away, the last typed brace is matched in the common case. This is ideal because one usually wants to know which brace was just closed:
```
( ( ... ))|
^
```

However if you happen to be editing something and type a brace that happens to be just before another brace, the brace to the right of the cursor is matched:

```
( ( ... )|)
^  
```

This is confusing because it is inconsistent - one would still want to know (or expect to see) the brace that was just closed. To add to the confusion, there is no indication of which brace is being matched.

### New behavior

This PR always matching the brace *before* cursor, if present.

As a convenience, the PR also matches the brace after the cursor, if there is no brace before cursor. To let user know exactly which brace is being matched in all cases, it highlights both braces in the pair, instead of just the far-away matched brace.

### DIFF
--- a/bundles/howl-themes/monokai/monokai.moon
+++ b/bundles/howl-themes/monokai/monokai.moon
@@ -163,6 +163,13 @@ return {
       background: '#0064b1'
       height: 'text'
 
+    brace_highlight_secondary:
+      type: flair.RECTANGLE
+      foreground: '#0064b1'
+      text_color: foreground
+      line_width: 1
+      height: 'text'
+
     list_selection:
       type: flair.RECTANGLE
       background: current
--- a/bundles/howl-themes/solarized_light/solarized_light.moon
+++ b/bundles/howl-themes/solarized_light/solarized_light.moon
@@ -178,6 +178,12 @@ return {
       background_alpha: 0.6
       height: 'text'
 
+    brace_highlight_secondary:
+      type: highlight.RECTANGLE
+      foreground: blue
+      line_width: 1
+      height: 'text'
+
     list_selection:
       type: highlight.RECTANGLE
       foreground: blue
--- a/bundles/howl-themes/steinom/steinom.moon
+++ b/bundles/howl-themes/steinom/steinom.moon
@@ -159,6 +159,13 @@ return {
       background: '#0064b1'
       height: 'text'
 
+    brace_highlight_secondary:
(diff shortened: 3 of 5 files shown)


nilnor:
Plus one for matching both braces, I see how that is useful. Since I use the VI stuff I always have a block cursor and hadn't thought about that. I'm less positive about the change in behaviour though. The cursor is actually (I would say at least) not between characters but at a particular character. Again, I use the VI stuff, and with a block cursor it's very clear what the current character is (i.e. the cursor is not between characters but actually at a character). To illustrate:

brace

This does not feel correct to me, since I would expect the pair corresponding to the current character to be highlighted. After checking other editors quickly this seems to be undecided though, VIM uses Howl's current behaviour while Emacs seems to do what this PR introduces. @kirbyfan64, @barnybug, thoughts?

shalabhc:
I see how it looks wrong with the block cursor. No wonder VIM (and Emacs) chose what they did. Options I can think of:

1. Highlight both pairs with different flair - needs yet another flair
2. Make this a config option `config.matching_braces_highlighted = 'left' | 'right'`
3. Make VI mode behavior different from regular mode
4. Leave current matching behavior as-is

shalabhc:
After playing around some, I ended up liking the highlight-both-pairs feature. The parens at cursor use the usual `brace_highlight` flair and the parens just before the cursor use a new `brace_highlight_secondary` flair.

Here are some examples - let me know what you think @howl-editor/developers:

parens1
parens2
parens3
parens4
parens5
parens6
parens7

kirbyfan64:
@shalabhc Oooh, that looks nice! :+1;

nilnor:
That looks promising, but I would have to try it out to be sure, could you push the changes? At any rate, should it not feel 100% we could have that but have the double matching be configurable.

shalabhc:
@nilnor - pushed

nilnor:
It feels interesting at any rate, I'll have to try it out more to be sure. But lets merge it - we can tweak it later if needed, e.g. making it configurable.

shalabhc:
:+1:

PR #177, opened 2016-03-02
Make Buffer.read_only work properly with aullar

### LINKED ISSUE #155: Insertion into read-only buffer works

nilnor:
Moves read-only status down in to aullar.Buffer.

Fixes #155.

### DIFF
--- a/bundles/vi/state.moon
+++ b/bundles/vi/state.moon
@@ -1,5 +1,5 @@
 import bindings from howl
-import getmetatable, setfenv, pairs, callable, print, tostring, pcall from _G
+import getmetatable, setfenv, pairs, callable, print, tostring, pcall, error from _G
 
 _G = _G
 _ENV = {}
--- a/lib/aullar/buffer.moon
+++ b/lib/aullar/buffer.moon
@@ -90,6 +90,7 @@ Buffer = {
     @_collect_revisions = false
     @text = text
     @_collect_revisions = true
+    @read_only = false
 
   properties: {
     size: => tonumber @text_buffer.size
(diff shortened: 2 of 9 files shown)


shalabhc:
Cool - it works. But I get different messages for different operations:

- insert, paste or backspace: warning 'Attempt to modify read-only buffer'
- delete: error 'Attempt to modify read-only buffer *buffer-title*'
- select and cut: error '/lib/howl/ui/selection.moon:76: Attempt to modify read-only buffer'

Would be nice if all operations had the same warning as insert.

nilnor:
@shalabhc Have a look now, should be more uniform. I also fixed a related problem with the ProcessBuffer.

bundles/vi/state.moon
```diff
@@ -1,5 +1,5 @@
 import bindings from howl
-import getmetatable, setfenv, pairs, callable, print, tostring, pcall from _G
+import getmetatable, setfenv, pairs, callable, print, tostring, pcall, error from _G
```
shalabhc: 'error' unused?

shalabhc:
Yup - looks great! :+1:

bundles/vi/state.moon
nilnor: The opposite actually, a previously missing import.

PR #178, opened 2016-03-04
Add icons to switch-buffer list

shalabhc:
This adds icons in the first column of the buffer list for `switch-buffer`. It also removes the status column which showed '*' for modified buffers since the icon indicates modification state. Custom icons also show status of process buffers - running, successful or error.

Some screenshots below (the icons should be self-explanatory, I hope):

buffer-icons
buffer-icons2

### DIFF
--- a/lib/howl/interactions/buffer_selection.moon
+++ b/lib/howl/interactions/buffer_selection.moon
@@ -3,9 +3,18 @@
 
 import app, interact, Project from howl
 import File from howl.io
+import icon, markup from howl.ui
 import Matcher from howl.util
 
 append = table.insert
+
+icon.define_default 'buffer', 'font-awesome-square'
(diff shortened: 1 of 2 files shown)


nilnor:
Looks great, very nice :)

For the already existing icons there's a configuration variable for controlling them, `file_icons`. Perhaps something similar here then, `buffer_icons`?

shalabhc:
Ah, right. Should `buffer_icons = false` then display the status column with `*`, `modified-on-disk` etc., or just hide that info altogether?

nilnor:
Preferably show something like the current solution, as we'll otherwise lose information unless icons are enabled.

shalabhc:
@nilnor - OK, added `config.buffer_icons` which lets user switch to the text status as in the current solution.

PR #180, opened 2016-03-07
Rework buffer title code

### LINKED ISSUE #179: Useless directory showing
dir_lst

Not sure why `[howl]` is appearing there, considering there isn't even any ambiguity and `howl` isn't actually the direct parent of the shown paths (it's `lib`).

The command I used to open the files was (run in the Howl source directory):

```
$ howl spec/buffer_spec.moon spec/editing/auto_pair_spec.moon lib/aullar/buffer.moon lib/howl/buffer.moon
```

kirbyfan64:
I *was* going to just add another `if` branch, but then an alarm started going off in my head, considering that that would have resulted in *three* nested `if`s! Two is fine, but three felt like code smell to me.

Closes #179.

### DIFF
--- a/lib/howl/interactions/buffer_selection.moon
+++ b/lib/howl/interactions/buffer_selection.moon
@@ -59,9 +59,10 @@ get_buffer_list = ->
     continue if #buffers == 1
 
     options_list = {
+      { project: true }
       { project: false, parents: 1 }
-      { project: false, parents: 2 }
       { project: true, parents: 1 }
+      { project: false, parents: 2 }
       { project: true, parents: 2 }
     }
 
--- a/lib/howl/interactions/buffer_selection.moon
+++ b/lib/howl/interactions/buffer_selection.moon
@@ -58,11 +58,17 @@ get_buffer_list = ->
   for basename, buffers in pairs(basenames)
     continue if #buffers == 1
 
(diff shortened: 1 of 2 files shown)


shalabhc:
Yes the loop is better :+1: 

But this probably needs an update to the related spec.

lib/howl/interactions/buffer_selection.moon
```diff
@@ -58,11 +58,17 @@ get_buffer_list = ->
...
 
-    titles = [make_title(buffer, project:true) for buffer in *buffers]
-    if has_duplicates(titles)
-      titles = [make_title(buffer, project:true, parents:1) for buffer in *buffers]
-      if has_duplicates(titles)
-        titles = [make_title(buffer, project:true, parents:2) for buffer in *buffers]
+    options_list = {
+      { project: false, parents: 1 }
+      { project: false, parents: 2 }
+      { project: true, parents: 1 }
```
shalabhc: There should also be just a {project: true} somewhere in here? i.e. just the project name with no parents.

kirbyfan64:
@shalabhc Updated!

shalabhc:
:+1:

@nilnor - going to merge since it a simple one

PR #181, opened 2016-03-07
Add a Pascal bundle

kirbyfan64:
Since the existing one was 99% garbage...


bundles/pascal/README.md
```diff
@@ -0,0 +1,12 @@
+# Nim bundle
```
shalabhc: 'Pascal'

kirbyfan64: @shalabhc I have *no* clue how the hell I managed to miss something that obvious...but...fixed!

bundles/pascal/pascal_lexer.moon
```diff
@@ -0,0 +1,210 @@
+howl.aux.lpeg_lexer ->
```
nilnor: Don't forget the copyright stanzas :)

bundles/pascal/misc/example.pas
```diff
@@ -0,0 +1,53 @@
+{ Comments }
```
nilnor: These comments (at least) can be multi-line. Not covered in the example or lexer.

nilnor: Hmm, or it seems to be partially working. I might have lucked out with a weird example here: [test file](https://github.com/graemeg/freepascal/blob/master/compiler/aoptobj.pas). When I open this the first comment isn't highlighted correctly, but any edits, etc. makes it work. A similar file seems to work out fine.

kirbyfan64: @nilnor Well, that's odd...

Here's a minimal test case:

```pascal
{




















}
```

Basically, it has to be a multiline comment over 20 lines long. I'm not quite sure what the problem is; it also occurs with any other forms of multi-line comment.

Looking into it, thought this may take a while...

kirbyfan64: @nilnor I think this is actually some sort of bug in the styling itself. I managed to make it happen once with a Python source file. Maybe it's just that the nested comments in Pascal bring it up?

nilnor: @kirbyfan64 that seems likely - I will have a look at it later.

kirbyfan64: @nilnor Maybe this is somehow related to #185?

kirbyfan64:
@howl-editor/developers Ready for review!

nilnor:
I think we can merge this now. One remaining thing though is still the copyright stanzas, all code files should have one, and the year should be 2016 (so update the stanza for the lexer as well).

kirbyfan64:
@nilnor Done.

kirbyfan64:
Bump?

nilnor:
The year of copyright should still be 2016, but I'll merge this now and adjust it in master, :+1:

bundles/pascal/misc/example.pas
nilnor: Blast from the past: I had a another look at this while looking at #185, and adjusted the lexer to deal with this in 7111796.

PR #182, opened 2016-03-22
Brace matching improvements

shalabhc:
- Made brace matching style aware, so operators are not matched with string literals, etc.
- Refactored brace matching implementation 
- Added cursor-match-brace command that moves cursor to matching brace, if any

### DIFF
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -145,8 +145,8 @@ command.register
   handler: (line_no) -> app.editor.cursor\move_to line: line_no, column: 1
 
 command.register
-  name: 'cursor-match-brace'
-  description: 'Jump to matching brace'
+  name: 'cursor-goto-brace'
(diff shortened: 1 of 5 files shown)


lib/howl/commands/edit_commands.moon
```diff
@@ -143,3 +143,12 @@ command.register
     return tonumber line_str
 
   handler: (line_no) -> app.editor.cursor\move_to line: line_no, column: 1
+
+command.register
+  name: 'cursor-match-brace'
```
nilnor: I'm not sure about the `match` part, since that sounds more like a highlighting thing than something that will actually change the position. Perhaps `cursor-matching-brace` or `cursor-goto-brace` (the latter to align with `cursor-goto-line`)?

nilnor:
Apart from the nitpick about the command name it looks good, :+1:

shalabhc:
Addressed - please take another look.

nilnor:
Great, merging!

Issue #150, opened 2015-12-28
Backspace before whitespace doesn't move cursor as expected


shalabhc:
1. Open a new buffer
2. Type 3 spaces followed by 'a'
3. Position cursor after first space (i.e. column 2)
4. Press `backspace`

The space is deleted, however the cursor is now at column 3 instead of column 1.

nilnor:
This is also by design. I'm not saying it necessarily has to be this way, but it's working as intended. When the configuration variable `backspace_unindents` is true and we're in leading indentation we unindent and move the cursor to the first non-blank column. Not something new in master either, this behaviour is the same as it always has been.

shalabhc:
I see.

The unindent action is expected, but moving the cursor right when I'm asking it to go left seems a bit odd. In this particular case I was trying to merge the current line into the previous so pressing `backspace` twice seemed the obvious way (I was on the column 2). However the cursor jump foiled my plan. Given other readily available commands - `ctrl_right` to jump to the start of the word  and `shift_tab` to unindent - is there a specific reason behind the jump right on `backspace`, or is it just incidental?

shalabhc:
FWIW, I also find it a bit jarring that `shift_tab` makes the cursor jump to the first non blank character. In my mental model, deletion of text should only move the cursor if the text before the cursor was deleted. So successive `shift_tab`s on a line would look like this (cursor shown as `|`):

```
    |    abc
    |  abc
    |abc
  |abc
|abc
```

nilnor:
I'm open for a PR changing this behaviour. It's currently the way it is since that's how it worked with Scintilla, and I kept the behaviour as much as possible when replacing it with aullar. Both these cases are now handled in `Editor`, so the code there should be changed along with the accompanying specs.

shalabhc:
OK. I have a couple of other things in the pipeline so I'll leave the bug open and get to a PR for this later.

shalabhc:
Fixed by #184

PR #184, opened 2016-04-01
Change cursor behavior when unindenting line

shalabhc:
This PR adjusts the behavior as per #150.

Using `backspace` to unindent now never moves the cursor to the right and always moves the cursor left. The conceptual idea is that the spaces to the left of the cursor are being deleted. Successive backspacing looks something like this:

```
    |    abc
  |    abc
|    abc
```

Using `shift_tab` to unindent now never moves the cursor to the right. It may move the cursor to the left to the first non blank character. The conceptual idea is that the rightmost spaces are being deleted. Successive shift-tabbing now looks something like this:

```
    |    abc
    |  abc
    |abc
  |abc
|abc
```

### DIFF
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -384,9 +384,10 @@ class Editor extends PropertyObject
 
     conf = @buffer.config
     if conf.tab_indents and @current_context.prefix.is_blank
+      cursor_col = @cursor.column
       cur_line = @current_line
       cur_line\unindent!
-      @cursor.column = cur_line.indentation + 1
+      @cursor.column = math.min(cursor_col, cur_line.indentation + 1)
     else
       return if @cursor.column == 1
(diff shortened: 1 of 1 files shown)


lib/howl/ui/editor.moon
```diff
@@ -384,9 +384,10 @@ class Editor extends PropertyObject
 
     conf = @buffer.config
     if conf.tab_indents and @current_context.prefix.is_blank
+      cursor_col = @cursor.column
       cur_line = @current_line
       cur_line\unindent!
-      @cursor.column = cur_line.indentation + 1
+      @cursor.column = math.min(cursor_col, cur_line.indentation + 1)
```
nilnor: `min` and `max` are locally available, so need to ref `math` here and below.

nilnor:
Looks good except the minor nitpick above, :+1:

shalabhc:
Addressed - please take another look.

nilnor:
Looks great!

PR #186, opened 2016-04-12
Command line persistent keymap

shalabhc:
The purpose of this PR is to remove the special case handling of keymaps by `interact.select` and introduce a more generic way for interactions to install keymaps that are active even when nested interactions are running.

This adds `command_line.persistent_keymap` - a keymap that is active for the interaction that sets this property, and also while any nested interactions are running. This also adds `command_line\refresh!` which invokes `refresh`, if present on all running interactions. These two are used to implement buffer closing in the `interact.select_buffer` list and the old implementation that passed a keymap to `interact.select` has been replaced.

### DIFF
--- a/lib/howl/interactions/buffer_selection.moon
+++ b/lib/howl/interactions/buffer_selection.moon
@@ -142,7 +142,7 @@ interact.register
         current_selection = selection
 
     command_line = howl.app.window.command_line
-    command_line.keymap = {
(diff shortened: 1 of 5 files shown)


shalabhc:
Withdrawn. Redoing this API to be more consistent with the forthcoming help API. Will submit a combined PR for both.

PR #188, opened 2016-04-19
Auto screenshots

nilnor:
This adds a new script for automatically generating screenshots, based on an original script from @shalabhc. The script has been used to produce new sets of screenshots for all four bundled themes, and the manual has been updated to use these. The manual also has gotten a theme selector where one can pick a theme to view the manual images in. Finally, four new automatically generated screenshots pages (one for each theme) has been added to the site, where one can view all available screenshots.


lib/scripts/screen-shooter.moon
```diff
@@ -0,0 +1,317 @@
...
+      dispatch.launch -> command.run 'buffer-grep resume'
+      app.editor.line_at_top = 8
+  }
+
+  {
+    name: 'buffer-replace'
+    ->
+      open_files { 'lib/howl/application.moon' }
+      dispatch.launch -> command.run 'buffer-replace /showing/'
+      app.editor.line_at_top = 10
```
shalabhc: The replacements start at line 50 so they're not visible. Maybe set this to 48?

shalabhc:
:+1: looks great - except for the one comment on the buffer-replace screenshot.

PR #189, opened 2016-04-21
Add 'buffer-sort-lines' command to sort lines in-place

shalabhc:
Self explanatory. This command sorts the selected lines in-place, or all lines of the buffer.

### DIFF
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -98,6 +98,25 @@ command.register
       app.editor.line_at_top = replacement.line_at_top
 
 command.register
+  name: 'buffer-sort-lines'
+  description: 'Sort the selected lines or the entire buffer'
+  handler: ->
+    editor = howl.app.editor
+    local lines
+    if editor.selection.empty
+      lines = editor.buffer.lines
+    else
+      lines = editor.active_lines
+    return unless #lines > 1
+
+    texts = [line.text for line in *lines]
+    table.sort texts, (a, b) -> a.stripped < b.stripped
+    editor.buffer\as_one_undo ->
+      for idx, line in ipairs lines
+        line.text = texts[idx]
+
+
+command.register
   name: 'editor-paste..',
   description: 'Pastes a selected clip from the clipboard at the current position'
   input: interact.select_clipboard_item


nilnor:
I was thinking we should add another command that replaces the selection or entire buffer with the output of a command who is feed the content, e.g. `replace-with-piped` or something similar. This could then be achieved with `replace-with-piped sort`. Would we still need or want to have special code for this then?

shalabhc:
I think that other command should also exist, but sorting functionality is core and it should be consistently available, independent of the presence or behavior or a `sort` command on your machine or platform (e.g. consider Windows)

lib/howl/commands/edit_commands.moon
```diff
@@ -98,6 +98,25 @@ command.register
       app.editor.line_at_top = replacement.line_at_top
 
 command.register
+  name: 'buffer-sort-lines'
+  description: 'Sort the selected lines or the entire buffer'
+  handler: ->
+    editor = howl.app.editor
+    local lines
```
kirbyfan64: Super minor nitpick...but, stylistically, why not just do:

```
local lines
lines = if editor.selection.empty
  editor.buffer.lines
else
  editor.active_lines
```

They compile to the same thing, but I think this reads a bit better.

nilnor: You wouldn't need the `local lines` in that case. FWIW, I think it's fine as is as well, even though I've used the if-statement-as-expression myself it reads a little bit funny to me.

nilnor:
Ok, so I'm fine with adding this command. However, it doesn't sort properly, and needs to be fixed to work with unicode. For example:
```
Ä
Å
ö
å
```

Should sort as 

```
Å
Ä
å
ö
```

Another thing that isn't working ideally is when you sort the entire buffer, and the last empty line ends up as the first line.

lib/howl/commands/edit_commands.moon
shalabhc: I have strong Python habits so the if expression looks odd to me too, specially since it's on multiple lines.

shalabhc:
Agree on both points. I was hoping to tackle unicode later, but might as well handle it properly now. The last empty line issue is debatable - since all other blank lines will end up at the top, I assume we still want the last empty line to stay there?

lib/howl/commands/edit_commands.moon
kirbyfan64: Well, I also have strong Python habits, but my Haskell side just loves expressions. ;)

shalabhc:
I'm having trouble with `ustring.ucompare` which seems to ignore case and also ignore leading non word characters. This may be OK so I'll revisit later but I'm withdrawing this for now and will submit another PR for an `editor-replace-exec' command.

PR #191, opened 2016-04-25
Fix matching of file extensions containing symbols

kirbyfan64:
Currently, files like `x.c++` won't be highlighted with the C/C++ mode. When I looked into it, it turned out that the extension pattern wasn't matching `c++`, causing the correct mode to not be loaded. This fixes that by making sure that the pattern correctly matches any extension.


nilnor:
Looks good! Remove the issue reference in the comment though, then merge at will.

kirbyfan64:
FYI, the build errors are just because I deleted the branch before Travis had a chance to complete. `master` is still green.

PR #194, opened 2016-05-18
Implement command line help

shalabhc:
- interact.select() keymap replaced with a more generic
`command_line\add_keymap()` method.
- New `command_line\add_help()` method to attach help to running
commands and interactions.
- Command line help displayed in a popup when `f1` is pressed from the
command line.
- Updated existing help text and added new help for many commands.

Screenshot of new help popup:
image


kirbyfan64:
Maybe the help screen could also show which keybindings the command is bound to like the autocomplete does?

shalabhc:
@kirbyfan64 sounds good. What do you think of this:

command-line-help

kirbyfan64:
@shalabhc :+1:

nilnor:
Looks good! I thought about one thing though - do we want to add some other indication that the `f1` key can be used, perhaps appending the command line's title?

shalabhc:
It could be nice but the problem is finding a good place for it. Just appending to the title looks cluttered. Also not all commands have a title (e.g. `buffer-replace`). Finally help is available on *all* commands so showing it every time can be somewhat redundant.  So I feel the manual is sufficient here.

nilnor:
OK, sounds good, let's content ourselves with that. I'll leave the merging and Changelog additions in your capable hands :)

PR #195, opened 2016-05-18
Website redesign

shalabhc:
Redesigned to a flatter, more spacious and more blue look.
Tweaked content on index and about pages.

Site now looks like this:
howl-site


nilnor:
Stellar :cake:

shalabhc:
Nice to see this live :tada:

shalabhc:
Thanks to @nilnor and @kirbyfan64 for the great suggestions and all the feedback that went into this!

PR #196, opened 2016-05-25
Add editor-replace-exec command

shalabhc:
Replaces selection or buffer with output of external command fed the
selection or buffer.

### DIFF
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -152,3 +152,23 @@ command.register
     pos = app.editor\get_matching_brace cursor.pos
     cursor\move_to(:pos, :extend) if pos
 
+
+howl.command.register
+  name: 'editor-replace-exec'
+  description: 'Replace selection with output of selection fed into external command'
+  input: ->
+    chunk = app.editor.active_chunk
+    app.window.command_line.title = 'External command'
+    working_directory, cmd = howl.interact.get_external_command!
+    return unless working_directory
+    return chunk, working_directory, cmd
+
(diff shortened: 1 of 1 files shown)


lib/howl/commands/edit_commands.moon
```diff
@@ -152,3 +152,23 @@ command.register
...
+  name: 'editor-replace-exec'
+  description: 'Replace selection with output of selection fed into external command'
+  input: ->
+    chunk = app.editor.active_chunk
+    app.window.command_line.title = 'External command'
+    working_directory, cmd = howl.interact.get_external_command!
+    return unless working_directory
+    return chunk, working_directory, cmd
+
+  handler: (chunk, working_directory, cmd)->
```
kirbyfan64: Probably overkill, but shouldn't there be a space before `->`?

shalabhc:
Another option is to use a shortcut that explicitly requests preview? E.g. `alt_enter` to preview (and `enter` to apply). 

The reason is 'run the command as you type' seems a bit scary to me if we don't get it exactly right. The problem is knowing the full command has been specified. I don't want to Howl to try and run a (possibly harmful) `sor` command on my way to running a 'sort' command. I do see the utility if I'm constructing a pipeline, e.g. `sort | uniq | awk ...`. Then, if I can fiddle with the order and arguments to things and see the output in real time - that would be very nice.

lib/howl/commands/edit_commands.moon
shalabhc: Yes there should - not overkill, this is a valid style issue and I think nilnor fixed this after merging.

PR #197, opened 2016-05-27
Better class name, constant and member highlighting for Python

shalabhc:


bundles/python/python_lexer.moon
```diff
@@ -46,12 +46,16 @@ howl.aux.lpeg_lexer ->
   operator = c 'operator', S'+-*/%~&^=!<>;:,.(){}[]|`'
 
   name = (alpha + '_')^1 * (alpha + digit + P'_')^0
+  type_name = c 'type', (upper^1 * (alpha + digit + '_')^0)
```
kirbyfan64: I don't know about this...a lot of Python scripts will begin functions with uppercase letters. In particular, I know a lot of Python scripts, including [several within Chromium](https://chromium.googlesource.com/chromium/src/build/+/master/gn_helpers.py) and [SCons](https://bitbucket.org/scons/scons/src/1e0827962a2bb114041df47d44be2384c86f3918/examples/src_packaging/SConstruct), use PascalCase for functions.

shalabhc: OK that's a good point - should I reduce this to just apply to the 'class' statement, or just leave it out?

kirbyfan64: @shalabhc What exactly do you mean by "apply to the 'class' statement!?

shalabhc: @kirbyfan64 - nevermind I'm confused. We already have `type_def` for that.

kirbyfan64: @shalabhc Actually, I got an idea.

Python [PEP 484](https://www.python.org/dev/peps/pep-0484/) adds type annotations.

The function annotations syntax is always like:

```
('(' | ',') identifier [':' annotation]
```

So, maybe a rule could be added like:

```moonscript
S'(,' * name * ':' * type_name
```

where `type_name` would be either one of the built-int types (e.g. `str`, `int`, `bytes`) *or* something matching the above rule that you put.

shalabhc:
@kirbyfan64: But the syntax allows any expression on the right hand side of the colon, no? So maybe it should just highlight as a regular expression?

PR #198, opened 2016-05-27
Blueberry blend theme

shalabhc:
I'm happy with how this looks now, but also open to further suggestions.


nilnor:
Great :+1: 

Let's have this in 0.4 since it does not affect anything else.

shalabhc:
Yay!

So - should we generate a new website then?

nilnor:
Yes, sounds like a good idea! You want to do that, and generate the screenshots?

shalabhc:
Yes, on it...

shalabhc:
Done - pushed commits and updated website for new theme.

PR #202, opened 2016-06-01
Fix compilation under FreeBSD

maxc01:
Tested under FreeBSD 10.2, of course, you should gmake instead make.

===
WoW, this is the first native modern editor for FreeBSD


nilnor:
Thanks a lot, great to hear about it running on FreeBSD!

maxc01:
FreeBSD needs more love :smile:

Issue #200, opened 2016-05-31
Window overflows in Elementary OS Freya


finnbar:
I don't know whether this is some weird Elementary OS thing, but whatever theme is set, the window seems to overflow to outside of the window controls. The window controls are still within the inside bit as well, it just seems to have some extra stuff on the outside. I've attached an image showing this.

![screenshot from 2016-05-31 20 07 49](https://cloud.githubusercontent.com/assets/4217801/15687260/9c5e9e50-276b-11e6-9d39-4c64956b1bd6.png)

Sorry if this is an easily fixable user setting rather than an actual bug, I've just picked up Howl (and it's super nice!) and cloned it from the repo.

nilnor:
(I'm assuming you refer to the window's title bar extending past Howl's editor view - please correct me if I'm wrong).

This is not actually a bug per se, but a theming setting. The outer padding is an option in themes (defaulting to a little bit of padding), which typically works out well. For your setup not so much however since in this case a full width would blend in better (nice screenshot by the way).

It's unfortunately not an easily fixable user setting at this point. At some point in the future we'll add easy theme overriding, but until this is in place it's slightly more work to customize a theme. For this example (customizing the Monokai theme) you can do the following:

- Copy the existing monokai theme file to your local ~/howl directory (`cp bundles/howl-themes/monokai/monokai.moon ~/.howl/monokai-custom.moon` from within the Howl directory)

- Modify the theme settings to your liking. In this case you would add `outer_padding: 0` under the `window:` section, like so:

```mooonscript
  window:
    outer_padding: 0
```

- Tell Howl to use your new customized theme in your init file (~/.howl/init.moon or ~/.howl/init.lua). Here's a snippet that should work regardless of whether you're using Lua or Moonscript:

```lua
howl.ui.theme.register('Monokai Custom', user_file('monokai-custom.moon'))
howl.config.theme = 'Monokai Custom
```

Let me know how it works out for you!

kirbyfan64:
@nilnor I thought the issue was referring to the giant black box outside of the window?

finnbar:
@kirbyfan64 Yes, I was referring to the black box, although I'm going to try your fix. Weirdly when opening a file in Howl by right-clicking in Files and selecting "open with Howl", the box disappears. I've got no idea what's happening there, but I'll report back on trying the fix.

nilnor:
[...]

PR #204, opened 2016-06-01
Make scrollbars themable

nilnor:
It didn't show itself to be much of a problem earlier, but probably only since the overlay scrollbars were often a visible orange, which wasn't that pleasing but at least clearly visible across all themes. When trying out Elementary OS however (for issue #200), whose theme uses black scrollbars, it's apparent that there is a problem if the scrollbars aren't visible.

So this makes scrollbars themeable. As ever, the Gtk support for CSS styling is a work of progress - I verified that this works on Ubuntu 16.04 and Elementary OS (Freya release, Gtk 3.14). I checked Ubuntu 14.04 and 12.04, where it's a no-op. I welcome any alternative choice for the actual colours choosen for the scrollbars of course.

Examples:

Blueberry

Monokai

### DIFF
--- a/bundles/howl-themes/blueberry_blend/blueberry_blend.moon
+++ b/bundles/howl-themes/blueberry_blend/blueberry_blend.moon
@@ -84,6 +84,10 @@ return {
   }
 
   editor: delegate_to content_box, {
+    scrollbars:
+      slider:
+        color: keyword
+        alpha: 0.8
 
     background: color: dark_blue
     indicators:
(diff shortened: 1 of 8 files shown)


shalabhc:
:+1: Sweet!

On my 'ugly scrollbars' machine, this is a little better than the old stuff, but it's still a thick pill. I'm trying to see what tweaks work but don't wait for the PR. (GTK CSS is even less fun than Web CSS, IMO).

thick-scrollbars

nilnor:
I agree regarding the CSS. There's probably some class somewhere that would deal with the remaining border as well if one looks deep enough.

Anyway, I'm merging this now then to make the scrollbars actually visible with black-scrollbar themes.

PR #216, opened 2016-06-08
Allow process buffer to show color codes in stderr

shalabhc:
This uses a highlight for the stderr and styles ANSI color codes
correctly.

Also added an opts.title parameter to the process buffer allowing a
custom title.

### DIFF
--- a/lib/howl/ui/process_buffer.moon
+++ b/lib/howl/ui/process_buffer.moon
@@ -144,7 +144,5 @@ highlight.define_default 'stderr',
   type: highlight.UNDERLINE
   foreground: 'red'
   line_width: 1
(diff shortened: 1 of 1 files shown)


nilnor:
Looks good, but we should make sure that the highlight looks good for the themes. The current single default looks too dark in all of the themes currently. Not sure whether we can find a good default that works in all themes, but otherwise we should define the new highlight specifically for the themes where it does not.

shalabhc:
Updated to make the highlight more prominent. I tested all the themes and it looks good to me - not too bold but also hard to miss. Let me know what you think?

nilnor:
:+1:

PR #217, opened 2016-06-09
Make sure we show any files opened with the --reuse parameter

### LINKED ISSUE #213: Focus on reuse
Currently the editor does not focus the newly opened file via `howl --reuse filename` and it has to be switched to from the buffer view.

The focus could be made default as it is pretty useful with a command line workflow.

Reference: https://www.reddit.com/r/howleditor/comments/4mry1q/focus_on_reuse/

nilnor:
Previously those were just added to the list, which meant that they
ended up at the bottom of the buffer list. Now we make sure that we display
any new files, filling out any existing editors with the new files.

Previous loading didn't do any existing check either, we we really need to
when files are opened with the --reuse parameter.

Fixes #213.

### DIFF
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -174,10 +174,10 @@ class Application extends PropertyObject
           editor.buffer = @next_buffer
 
   open_file: (file, editor = @editor) =>
-    for b in *@buffers
-      if b.file == file
-        editor.buffer = b
-        return b, editor
+    buffer = @_buffer_for_file file
+    if buffer
+      editor.buffer = buffer
+      return buffer, editor
 
     buffer = @new_buffer mode.for_file file
     status, err = pcall ->
@@ -292,8 +292,16 @@ class Application extends PropertyObject
 
     @settings\save_system 'session', session
 
+  _buffer_for_file: (file) =>
+    for b in *@buffers
+      return b if b.file == file
+
+    nil
+
   _load: (files = {}) =>
     local window
+
+    -- bootstrap if we're booting up
     unless @_loaded
       @settings = Settings!
       @_load_core!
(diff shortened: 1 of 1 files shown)


lib/howl/application.moon
```diff
@@ -316,12 +324,29 @@ class Application extends PropertyObject
...
+
+    -- files we've loaded via a --reuse invocation should be shown
+    if #loaded_buffers > 0 and @_loaded
+      for i = 1, math.min(#@editors, #loaded_buffers)
+        @editors[i].buffer = loaded_buffers[i]
+
+    -- all loaded files should be considered as having been viewed just now
+    now = howl.sys.time!
+    for b in *loaded_buffers
+      b.last_shown = now
```
shalabhc: I don't think you need this because `editor\_show_buffer` updates this already. I see that the 'switch-buffer' command does not set this either.

lib/howl/application.moon
```diff
@@ -316,12 +324,29 @@ class Application extends PropertyObject
...
     for path in *files
       file = File path
-      buffer = @new_buffer mode.for_file file
-      buffer.file = file
-      signal.emit 'file-opened', :file, :buffer
+      buffer = @_buffer_for_file file
+      unless buffer
+        buffer = @new_buffer mode.for_file file
+        buffer.file = file
+        signal.emit 'file-opened', :file, :buffer
```
shalabhc: Similar code in `open_file` above is pcall protected but here it is not - should it be?

lib/howl/application.moon
nilnor: We need it since we just load buffers from the command line straight into buffers without passing any editor. For the --reuse case we then explicitly fill any already open editors.

nilnor:
Added error handling for the loading case now, and improved the initial log message to indicate startup errors better.

shalabhc:
:+1:

PR #219, opened 2016-06-09
ListWidget and StyledText enhancements

shalabhc:
The main motivation here is to make list selections more aesthetically pleasing.

1. Added following features to StyledTable, and consequently, to `interact.select` and related lists:
  - Column `min_width`: Extra padding is added to make columns at least this wide. Specified as number of characters.
  - Column `align: 'right'`: Cells are right aligned instead of left aligned.

2. Changed scrolling behavior of `ListWidget` to be row-wise instead of page-wise. This makes it look more like how buffers scroll when you hold down the `up` or `down` key. IOW, it scrolls in one line at a time, instead of jumping an entire page and re-positioning the selection at the top of that page.

3. Finally the above features have been activated in a couple of places:
  - Added min_width for the command list columns, this makes the scrolling appear more smooth
  - Added right alignment and min_width for the line number field in line selections (`buffer-grep`, etc.)

### DIFF
--- a/lib/howl/command.moon
+++ b/lib/howl/command.moon
@@ -238,12 +238,18 @@ interact.register
   evade_history: true
   handler: (opts={}) ->
     opts = moon.copy opts
+    command_items = get_command_items!
+    name_width = 0
+    shortcut_width = 0
+    for item in *command_items
+      name_width = math.max(name_width, item[1].ulen)
+      shortcut_width = math.max(shortcut_width, item[2].ulen)
     with opts
-      .items = get_command_items!
+      .items = command_items
       .headers = { 'Command', 'Key binding', 'Description' }
       .columns = {
-        { style: 'string', min_width: 30 }
-        { style: 'keyword', min_width: 15 }
+        { style: 'string', min_width: name_width }
+        { style: 'keyword', min_width: shortcut_width }
         { style: 'comment' }
       }
     result = interact.select opts
--- a/lib/howl/command.moon
+++ b/lib/howl/command.moon
@@ -242,8 +242,8 @@ interact.register
       .items = get_command_items!
       .headers = { 'Command', 'Key binding', 'Description' }
       .columns = {
-        { style: 'string' }
-        { style: 'keyword' }
+        { style: 'string', min_width: 30 }
+        { style: 'keyword', min_width: 15 }
         { style: 'comment' }
       }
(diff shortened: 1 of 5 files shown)


lib/howl/command.moon
```diff
@@ -204,8 +204,8 @@ interact.register
       .items = get_command_items!
       .headers = { 'Command', 'Key binding', 'Description' }
       .columns = {
-        { style: 'string' }
-        { style: 'keyword' }
+        { style: 'string', min_width: 30 }
```
nilnor: Perhaps we could detect the actual widths to use instead of using a hard coded value? As it is this almost makes the list not move but not quite.

lib/howl/interactions/line_selection.moon
```diff
@@ -97,6 +97,10 @@ interact.register
     opts.matcher = matcher
     opts.on_change = line_match_highlighter(editor, matcher.explain)
     opts.force_preview = true
+    opts.columns = {
+        {align: 'right', style: 'comment', min_width: 3},
```
nilnor: Ditto for the hard coded value.

lib/howl/command.moon
shalabhc: That means one very long command could push the entire column to be pretty wide - maybe not a big deal. I'll try it out.

lib/howl/interactions/line_selection.moon
shalabhc: That would require scanning the entire list of lines up ahead - maybe not a big deal for most cases but could occasionally be pretty large.

shalabhc:
OK I'll take a deeper look to determine what's causing the lag. I definitely don't want to make it slower even a little. In theory it should have been the same speed but something is amiss.

lib/howl/command.moon
shalabhc: Done!

lib/howl/interactions/line_selection.moon
shalabhc: Done!

shalabhc:
I've taken care of the adjusting the min widths based on the actual widths of the items. Let me know if that looks better.

nilnor:
The lagging turned out to be dependent on some other loading activity at the time - I could not reproduce the lag later. So let's merge this, and we'll keep an eye open for regressions.

PR #222, opened 2016-06-13
Making the escape key configurable

RBlafford:
This change uses the binding_for method to allow users to customize
the key that will abort commands, close the command line, file
browser etc.

### DIFF
--- a/lib/howl/command.moon
+++ b/lib/howl/command.moon
@@ -158,7 +158,8 @@ class CommandInput
     tab: => @command_completion!
     up: => @run_historical!
     ctrl_r: => @run_historical!
-    escape: => self.finish!
+    binding_for:
+      ["cancel"]: => self.finish!
     enter: =>
       if @command_line.text
         cmd = resolve_command(@command_line.text)
--- a/lib/howl/interactions/file_selection.moon
+++ b/lib/howl/interactions/file_selection.moon
@@ -152,9 +152,10 @@ class FileSelector
       return false unless @command_line.text.is_empty
       @_chdir @directory.parent if @directory.parent
 
-    escape: =>
-      app.editor\cancel_preview!
-      self.finish!
+    binding_for:
+      ["cancel"]: =>
+        app.editor\cancel_preview!
+        self.finish!
 
     ctrl_s: =>
       @show_subtree = not @show_subtree
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -4,7 +4,7 @@
 ffi = require 'ffi'
 
 import Window, Editor, theme from howl.ui
-import Buffer, Settings, mode, bundle, bindings, keymap, signal, interact, timer, clipboard, config from howl
+import Buffer, Settings, mode, bundle, bindings, keymap, signal, interact, timer, clipboard, config, command from howl
 import File, Process from howl.io
 import PropertyObject from howl.aux.moon
 Gtk = require 'ljglibs.gtk'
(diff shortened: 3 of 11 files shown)


lib/howl/command.moon
```diff
@@ -158,7 +158,8 @@ class CommandInput
     tab: => @command_completion!
     up: => @run_historical!
     ctrl_r: => @run_historical!
-    escape: => self.finish!
+    binding_for:
+      ["cancel"]: => self.finish!
```
kirbyfan64: Out of curiosity, is there a reason this can't just be `cancel: => self.finish!`?

lib/howl/interactions/file_selection.moon
```diff
@@ -152,9 +152,10 @@ class FileSelector
       return false unless @command_line.text.is_empty
       @_chdir @directory.parent if @directory.parent
 
-    escape: =>
-      app.editor\cancel_preview!
-      self.finish!
+    binding_for:
+      ["cancel"]: =>
```
kirbyfan64: Same here.

lib/howl/command.moon
RBlafford: Just tried that. Didn't work. I followed the documented instructions on bindings found here http://howl.io/doc/api/bindings.html

RBlafford: After speaking to some helpful people on the gitter channel, this suggestion was further clarified. Will fix and push

shalabhc:
In general I like the motivation behind this change - to make the action keys within interactions configurable.

My concern is that `cancel` is not a command so it's confusing to see it in the keymap as well as under `binding_for`.

One option is to implement key translation aliases. For instance something like `bindings.alias_key "escape", "cancel"`. This would then make the escape key have an additional translation `"cancel"`.

Now `cancel` can be used in keymaps directly, outside of `binding_for`. The confusing thing here might be that `cancel` isn't a key on the keyboard. So maybe these could follow the convention of always using a prefix, e.g. `action_cancel` - or maybe not, I'm not sure yet.

nilnor:
I too like the motivation behind the change, but it does end up being slightly weird. The new `cancel` command is just a place holder, and it's in effect not possible to assign multiple key bindings to this new command since only the one would be used anyway.

So, some other way of doing it would be nice, but like @shalabhc I'm not sure exactly what it would be. The virtual thing could be a way to go, but it does end up with a value that isn't really a key if one was to choose `cancel` as a virtual key.  Still, if we we're to translate it to `escape` instead and treat it as a key remapping mechanism it might not be so bad.

shalabhc:
OK, so based on @nilnor's comment I have another suggestion. 

1. Leave the existing keymaps as-is.

2. Implement a `remap_key` function that works as follows:
  
    ```moonscript
    bindings.remap_key 'ctrl_g', 'escape'
    ```

    Now, pressing `ctrl_g` will add `escape` to the translations. So all existing keymaps should just work.

I think the advantage of this approach is that keymap authors do not have to know the names of any *virtual* keys that they must use and they always use the real keys. Also, it enables key binding modifiers (such as emacs) to do any appropriate mappings. 

Another advantage is that this leaves no room for inconsistency - consider that for both previously suggested options, it was possible for an author to still use `escape` and then their keymap would not conform to any re-bindings.

nilnor:
I would prefer the approach outlined by @shalabhc fwiw.

nilnor:
Closing due to inactivity - please reopen if/when ready again.

PR #223, opened 2016-06-21
Howl linting

nilnor:
Adds a new `howl-lint` script based upon the Moonscript linter, adjusts the code as per the complaints, and adds linting as a build step for Travis.


shalabhc:
Wow - lot of good fixes in here. I had noticed a couple of these recently and glad to see that the linter found them.

:+1:

PR #225, opened 2016-06-22
Buffer inspections

nilnor:
This adds a first version of inspection support for Aullar. Inspectors are registered wit the `inspection` module, and the `inspect` module deals with running these according to configuration settings and displaying the results.

<hr/>
[previously] 

This is a work-in-progress branch for getting a generic inspections framework in place for Howl. It allows inspectors to be present for both buffers and their modes. All inspectors are run, with the results being coalesced and displayed visually for the buffer. As a first proof-of-concept a Moonscript inspector has been implemented (not coincidentally since that works for Howl code).

Current things to consider:

- Currently the inspections are run on save, but work is in progress to add inspections-on-idle as well.
- The inspectors are currently assumed to be Lua functions, but we want to support just specifying an external command to run as well with automatic parsing of the output
- Should we look at generalizing this further to allow arbitrary code to jack into the same structure?
- Inspections are currently cleared upon modification, but that makes it jarring, so that should be changed

### DIFF
--- a/lib/howl/inspect.moon
+++ b/lib/howl/inspect.moon
@@ -208,6 +208,16 @@ highlight.define_default 'warning',
   line_type: 'solid'
 
 command.register
+  name: 'buffer-inspect'
+  description: 'Inspects the current buffer for abberations'
+  handler: ->
+    buffer = app.editor.buffer
+    criticize buffer
+    editor or= app\editor_for_buffer buffer
+    if editor
+      update_inspections_display editor
+
+command.register
   name: 'cursor-goto-inspection'
   description: 'Goes to a specific inspection in the current buffer'
   input: ->
--- a/lib/howl/inspect.moon
+++ b/lib/howl/inspect.moon
@@ -219,7 +219,9 @@ command.register
           unless i == #inspections
             pbuf\append "\n"
 
-      popup.view.cursor.line = 1
+      with popup.view
+        .cursor.line = 1
+        .base_x = 0
 
       spans = selection.spans
       highlight.remove_all 'search', buffer
--- a/lib/howl/inspect.moon
+++ b/lib/howl/inspect.moon
@@ -1,7 +1,7 @@
 -- Copyright 2016 The Howl Developers
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
-{:app, :config, :log, :signal, :config} = howl
+{:app, :config, :log, :signal, :config, :timer} = howl
 {:highlight} = howl.ui
 {:pcall} = _G
 {:concat, :sort} = table
@@ -81,6 +81,7 @@ mark_criticisms = (buffer, criticisms) ->
 
   criticisms.count = #ms
   buffer.data.inspections = criticisms
+  buffer.data.last_inspect = buffer.last_changed
 
   if #ms > 0
     markers\add ms
@@ -106,13 +107,21 @@ criticize = (buffer, criticisms) ->
   mark_criticisms buffer, criticisms
 
 update_buffer = (buffer, editor) ->
-  return unless buffer.config.auto_inspect
-  return unless buffer.size < 1024 * 1020 -- 1MB
+  data = buffer.data
+  if data.last_inspect and data.last_inspect >= buffer.last_changed
+    return
 
   criticize buffer
   editor or= app\editor_for_buffer buffer
   if editor
     update_inspections_display editor
+
+on_idle = ->
+  b = app.editor.buffer
+  return unless b.config.auto_inspect == 'idle'
+  return unless b.size < 1024 * 1024 -- 1MB
+  update_buffer app.editor.buffer, app.editor
+  timer.on_idle 0.5, on_idle
 
 signal.connect 'buffer-modified', (args) ->
   with args.buffer
@@ -124,10 +133,27 @@ signal.connect 'buffer-modified', (args) ->
     editor.indicator.inspections.text = ''
 
 signal.connect 'buffer-saved', (args) ->
-  update_buffer args.buffer
+  b = args.buffer
+  return unless b.config.auto_inspect == 'save'
+  return unless b.size < 1024 * 1024 -- 1MB
+  update_buffer b
 
 signal.connect 'after-buffer-switch', (args) ->
   update_buffer args.current_buffer, args.editor
+
+signal.connect 'app-ready', (args) ->
+  timer.on_idle 0.5, on_idle
+
+config.define {
+  name: 'auto_inspect'
+  description: 'When to automatically inspect code for abberrations-'
+  default: 'idle'
+  options: {
+    { 'manual', 'Only inspect when explicitly asked' }
+    { 'idle', 'Inspect on idle' }
+    { 'save', 'Inspect when saving a buffer' }
+  }
+}
 
 highlight.define_default 'error',
   type: highlight.UNDERLINE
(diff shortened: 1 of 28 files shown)


shalabhc:
Very nice!

Some comments:

* There should some way to jump to next message and/or show all buffer messages. I had 10 inspections in a buffer but finding them all wasn't easy.

* How generic should the message annotations be? Is it only for inspections? What if I want to attach other messages such as code review comments or test results? A simple way to add annotations has wide applicability.

    * Why is `buffer.data.inspections` needed? Could we just use markers only and then any code could add the right markers, making it more generic?

* Not a fan of showing the message as a log message but this is subjective. If possible, some kind of pluggable display would be nice. Note this will also pollute the log entries if we ever show them.

* How would an inspector be set for a buffer directly? Would I set it in 'after-buffer-switch'?

* In the Python world (and possibly other languages too) there is not one but many popular linters and meta-linters. Having one bound to the mode seems limiting. Ideally I want to do something like `set-for-mode inspect=flake8`. Having this a config option may automatically provide project level configuration when that lands.

* Would be nice to have a command to explicitly run inspections for the current buffer.

lib/howl/inspect.moon
```diff
@@ -0,0 +1,147 @@
...
+  criticisms
+
+criticize = (buffer, criticisms) ->
+  criticisms or= inspect buffer
+  buffer.markers\remove name: 'inspection'
+  mark_criticisms buffer, criticisms
+
+update_buffer = (buffer, editor) ->
+  return unless buffer.config.auto_inspect
+  return unless buffer.size < 1024 * 1020 -- 1MB
```
shalabhc: Not quite 1MB.

Also 1MB seems a bit small - why limit this?

nilnor:
@shalabhc:

- Yes, we need some way of showing all message. I was thinking about something like the buffer-structure display, showing all lines with mesages on them. 

- We can make it more generic. I originally only used the markers, it now currently uses the data.inspections for performance reasons - since we need to lookup messages on each cursor change we don't want to scan all the markers all the time. We could later look at optimizing that and then go back to only use markers again.

- For showing the messages log message was chosen as a starting point as it's so easy, but that could be changed to something else (popups comes to mind).

- Yes, an inspector for a buffer would currently be set using a signal handler (probably `file-opened` though?). For projects it would be nice to get project-specific properties in place, which I think would help with most use cases.

- We could make inspectors registrable, similar to completers. One could then make it a config option, but it would be a list. 

- And yes, having a command is a good thing. Once we have the idle stuff in there that needs to be configurable anyway (when to run it). So if one option would be to never run it automatically the command can be used explicitly.

lib/howl/inspect.moon
nilnor: The limit itself is somewhat arbitrary, but I think a limit makes sense just so that one doesn't open a larger file and things hang on linting. I'm open to suggestions on what it should be though (and will fix the typo meanwhile). How big should it be though, how large a file would one open for actual editing where linting would helpful?

shalabhc: Even with slowish linters (1s) I don't experience any hang because the linter is an external process. This is only a concern for in-process linters then?

shalabhc:
* I do use a buffer structure type display (specifically, using `interact.select_location`) and it works quite well for me.

* (data.inspections) - OK

* I do like popups but the downside is you either auto-show the popup and obscure the next line or show-on-demand and require a keystroke invocation. Anyway, we can keep this as a log for now and revisit later because this seems easy to change.
   * A tangential idea came to mind - to modify the logging system to be more configurable by adding some more context to the logging call and the ability to install different log handlers. For instance `log.info 'line too long', source: 'buffer-message'` could be displayed as a popup and `log.info 'file saved', source: 'command'` could displayed in the status bar. You could install handlers with something like `log.bind 'buffer-message', (msg) -> display_in_popup(msg)`. What do you think - worth discussing separately?

* I wasn't sure if `file-opened` would work with preexisting buffers after a restart.

* I like the idea of registrable inspectors, and yes - a list config option. 

* SGTM

nilnor:
Another point to remember: When we remove the removal of marker upon modification, we need to remove any inspection markers being affected by the edit.

lib/howl/inspect.moon
```diff
@@ -124,24 +133,38 @@ signal.connect 'buffer-modified', (args) ->
...
 
 signal.connect 'after-buffer-switch', (args) ->
   update_buffer args.current_buffer, args.editor
 
+signal.connect 'app-ready', (args) ->
+  timer.on_idle 0.5, on_idle
+
+config.define {
+  name: 'auto_inspect'
+  description: 'When to automatically inspect code for abberrations-'
```
shalabhc: Typo '-'

shalabhc:
So I'm now thinking we don't need registered inspectors as discussed earlier in this PR, but rather we just use commands as inspectors. 

Here's the idea: On idle or on save, the buffer inspector is run. The inspector is just a standard command, say 'py-lint' or 'moon-lint' that could be invoked manually if needed. The command's job is to update the 'inspection' markers. You can have a generic command that runs an external process, parses the output to add the markers, e.g. `external-linter /usr/bin/flake8`

In fact, the inspector command doesn't have to add inspections, it it just a command that is run on-save or on-idle. So maybe these could be generic config options:

```
set-for-mode on_save_command=external-linter flake8
set-for-mode on_idle_command=py-quick-linter
```

Thoughts?

nilnor:
I think it's a good idea to have the actual linting being separate from the displaying of results (so writing a new linter does not include having to worry about markers). To me, having them as separate commands does not make that much sense unfortunately, and it seems like it wouldn't solve the problem of easily choosing multiple linters for different  code files. Thus, I'm still in favour of having it either as a config option or by querying both buffers and modes similarly to how completions are handled.

shalabhc:
OK decoupling the markers from linters sounds good. I'll wait and see how the config develops.

shalabhc:
I integrated flake8 (a Python linter) with this system and it works great. I just had to write a function to run flake8, parse the output and return criticisms. I enabled it by doing this `howl.mode.by_name('python').inspectors = {flake8}`. Kludgy, but works.

shalabhc:
After using it for a bit I have the following feedback:

* I like the cursor-goto-inspection command.
* I don't like the status bar message when the cursor is on an inspection. I find using both the indicator and status bar for the inspections messages is somewhat jarring, and too noisy for me. I think it may look better if:
  * either, all inspection messages were shown only in the indicator and never in the status bar. It could combine all the info, e.g. "(2 of 3) unused local".
  * or, inspection messages weren't shown in the indicator/status bar at all and always shown only in the popups. The popup appears automatically if the cursor sits there for 2 seconds. This would make it most consistent with the cursor-goto-inspection command.

lib/howl/inspect.moon
```diff
@@ -140,13 +141,18 @@ show_popup = (editor, inspections, pos) ->
 
 display_inspections = (editor) ->
   pos = editor.view.cursor.pos
+
+  -- if we've already displayed the message at this position, punt
+  return if pos == last_display_position
```
shalabhc: Isn't last_display_position tied to the current editor and not global?

lib/howl/inspect.moon
```diff
@@ -139,7 +139,12 @@ show_popup = (editor, inspections, pos) ->
     keep_alive: true,
   }
 
-display_inspections = (editor) ->
+display_inspections = ->
+  timer.on_idle (config.display_inspections_delay / 1000), display_inspections
```
shalabhc: Isn't this a perpetual infinite loop?

Instead, why not trigger in 'cursor-changed' and from on_idle?

nilnor: It's a reoccurring idle timer, so I wouldn't call it a loop. I think it's easier to just use the existing idle timer functionality than to attempt managing the interval / resetting / cursor state handling registering manually, and it's not in practice very inefficient either.

lib/howl/inspect.moon
nilnor: It is, I will look at handling editor changes as well.

lib/howl/inspect.moon
shalabhc: Ah OK.

nilnor:
This first version should now be ready for merging.

shalabhc:
:+1: 

A great addition to the feature set!

PR #227, opened 2016-06-25
Line motion commands and bindings

shalabhc:
This adds the following commands:

* `editor-move-lines-up` - moves current or selected lines up by one line
* `editor-move-lines-down` - moves current or selected lines down by one line

See a preview here: https://www.youtube.com/watch?v=w9FPK4XtESo

This also contains the following API changes:

1. Adds a new API method `editor\with_selection_preserved` - this uses the markers to preserve the selection through buffer edits. 

2. Implements the trimming feature for markers that have `preserve: true` set - this causes markers to be trimmed rather than removed when text at either boundary is deleted.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased (in master)
 
+- Added new commands `editor-move-lines-up` and `editor-move-lines-down` bound
+to `alt_up` and `alt_down` by default. These move the current or selected lines
+up (or down) by one line while preserving the selection.
+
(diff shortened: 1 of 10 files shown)


nilnor:
Looks great I think! Only possible addition I can think is the change for the buffer lines indentation setter - perhaps it's worth to add a spec for the case this previously didn't handle?

Apart from that, merge at will (and add a Changelog entry if you will).

shalabhc:
OK, added another spec now and also updated both Changelog and the API doc for editor.

PR #229, opened 2016-06-25
Proper OS X keymap

hvze:
Rebound some keys to work on OS X.

### DIFF
--- a/lib/howl/keymap.moon
+++ b/lib/howl/keymap.moon
@@ -114,7 +114,7 @@
 
     osx:
       editor: {
-        meta_shift_a:     'editor-select-all'
+        meta_a:           'editor-select-all'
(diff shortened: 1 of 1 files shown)


shalabhc:
:+1: 

Some comments might only make sense in the context of this diff so they could probably be adjusted. In general I like all changes I see there (even though I haven't actually tried them out).

hvze:
Yeah, I'm going to do some more testing with keys and see if I can increase efficiency.

shalabhc:
@hvze - are you happy with this keymap now?

hvze:
I'm gonna find my laptop and do some more modifications, I've been using it for a while but I will make the longer keybinds feel more comfortable.

PR #230, opened 2016-06-27
Generalize the sub_lex_by_inline lexer helper

nilnor:
Where it previously internally used scan_until this is now left for the
user. This makes it possible to use arbitrary patterns for matching the
text to be sub lexed, e.g. `scan_through_indented`.


nilnor:
It should be noted that this changes the semantics of the helper, but it was added for 0.4 with only one internal use so far.

PR #231, opened 2016-06-29
Better cross platform shell compatiblity; Not all default shells are bash

Immortalin:


nilnor:
What part actually requires `bash`? Or did you mean bourne shell? If the latter, `/bin/sh` would be better as I rather not have it depend on bash at all.

Immortalin:
`sh` doesn't always point to Bourne shell on all systems

kirbyfan64:
Isn't it usually Bourne-compatible, though? On Ubuntu/Elementary OS, it points to Dash, which is fully compatible.

shalabhc:
Note that `bash` isn't even installed by default on FreeBSD, but `sh` is available. https://www.freebsd.org/doc/en/articles/linux-users/shells.html

PR #232, opened 2016-06-29
Add editor-move-text-left|right commands

shalabhc:
Adds commands to move the selection (or current character) left and right by one character. Bound to `alt_right` and `alt_left`


nilnor:
:1, in it goes.

PR #236, opened 2016-07-03
Added project-other-open command

shalabhc:
Lets you first select one project and then a file in the project. It doesn't matter what your current project is  (or if you have one).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ## Unreleased (in master)
 
+- Added a new command `project-other-open` that allows opening a file in any
+  project.
+
 - Added command line help which is invoked by pressing `f1` while any
 interactive command is running. This displays a popup containing a description
 
--- a/lib/howl/commands/file_commands.moon
+++ b/lib/howl/commands/file_commands.moon
@@ -84,6 +84,20 @@ command.register
     app\open_file file
 
 command.register
+  name: 'project-other-open',
+  description: 'Open a file in another project'
+  input: ->
+    selected = interact.select
+      title: 'Project'
+      items: [{root.basename, root.path, :root} for root in *Project.roots]
+      columns: {{style: 'keyword'}, {style: 'comment'}}
+
+    return unless selected
+    return interact.select_file_in_project project: Project.for_file selected.selection.root
+  handler: (file) ->
+    app\open_file file
+
+command.register
   name: 'save',
   description: 'Save the current buffer to file'
   handler: ->
--- a/lib/howl/keymap.moon
+++ b/lib/howl/keymap.moon
@@ -94,6 +94,7 @@
   ctrl_o:           'open'
   ctrl_shift_o:     'open-recent'
   ctrl_p:           'project-open'
+  ctrl_shift_p:     'project-other-open'
   ctrl_shift_r:     'exec'
   ctrl_alt_r:       'project-exec'
   ctrl_shift_b:     'project-build'


nilnor:
How about addding the ability to switch projects from within the ordinary `project-open` instead, since we already have the ability to e.g. switch to different command from within the command line? Also, we should store the known roots in the session to make the switching work fine.

shalabhc:
> How about addding the ability to switch projects from within the ordinary project-open instead

Do you mean we should remove the special keymap for project-other-open or that we should forego the command entirely and integrate the functionality in project-open?

nilnor:
Something of the latter kind, I'm thinking that if we store the projects root so that they're already known there might be a better way of selecting another project than explicitly navigating to it first, since we already have a list available. So one idea (that I'm not sure how well it work out in practice) would be to allow for switching projects directly in the prompt.

shalabhc:
OK how about this - when you launch project open you see the current project name in the prompt ('|' being the cursor):
```
project-open howl-project:|
```

Then you can type away to search the current project, or press backspace and you see this:
```
project-open |
```
With the completion list showing the list of all known projects. When you select a project, you go back into the project specific search:

```
project-open other-project:|
```

Essentially this makes projects behave list list list of top level directories. One advantage is you can directly bind a shortcut to a specific project:

```
howl.bindings.push
   f2: 'project-open my-project:'
```

However, the problem is the base directory name of each project may not be globally unique, so we need a way to disambiguate them.

equim:
Sounds good to me.  I guess the completion behaviour becomes quite similar to the set command where you can backspace past the = to edit the setting you want to change.  Also quite similar to the open command in that the current directory is pre-populated.

Could we just use the absolute path for the project?  That would resolve the uniqueness and allow you to open a file from a project that isn't currently open.  We could still have suggested completions for the known / open projects along with the normal directory completions.

Taking this further we could maybe do the same for the open command - typing a colon would enter recursive mode.  This keeps the commands consistent and adds protection against accidentally backspacing up a directory with recursive mode still on (we do have a cap on how many files we look for in recursive mode but still).

equim:
Eclipse has functionality to open a file in the current 'workspace' where you start entering the file name and you get completions for all matching files across all open projects.  Perhaps we can add that functionality with this proposed syntax by allowing no project name to be specified before the colon:
```
project-open :|
```

shalabhc:
Using ':' for recursive search in `open` is in interesting idea! If we make `project-open` and `open` similar, the question arises whether we need two commands for this at all? Or would `open` subsume all the functionality? Consider you could have a shortcut that lands you at `open /path/to/current/project:` and perhaps a shortcut to let you select one of many projects to 'open'.

Re eclipse - what constitutes a workspace? It seems too expensive and not particularly useful to do a cross project search. E.g. I could have 10s of unrelated projects on one machine.

shalabhc:
On further thought I think a dedicated project open command may be better because you need functionality like creating a new project. Eventually we may want a single project with multiple roots, etc.

nilnor:
I think that's a great idea! Seems like it would be quite intuitive (for me at least), that you would just back up to get a full list. And as you said, that list could have an option for opening a new project directory as well.

shalabhc:
Any thoughts on how to resolve duplicate named, @nilnor?Maybe just do 'name <2>', etc?

nilnor:
If possible, I would prefer to prepend the parent directory, e.g. `parent/<project-dir>`.

equim:
I might be misunderstanding but I don't get why it wouldn't be the absolute path to the project directory.

Regarding eclipse, basically a workspace is a persisted session / eclipse instance.  A workspace is a directory containing settings, references to projects and session state (all persisted).  Each eclipse instance (process) has exactly one workspace.  A workspace can only be opened in one eclipse instance at a time, though I think a project can be referenced from more than one workspace.  Eclipse also keeps a cache of all the file names across all open projects in the workspace so its project-open equivalent is fast after the first access.

I almost always only have one workspace - probably because eclipse is a heavyweight IDE so switching workspaces is painful.  This means I tend to have multiple unrelated projects in the same workspace.  Even there though, they are all projects that I am working on, and it's nice to be able to just type in a filename and jump to it regardless of the project.  If there are collisions I either type more or just select the right one from the available completions.

I see howl as already pretty similar to eclipse here - we can open multiple projects in a session/instance (though I don't think they are persisted yet) and we can jump to any file in a project with completions.  Being able to jump to any file across any currently open project would be useful to me if I were using howl as much as I use eclipse, but caching the file lists would make it a lot more usable.

shalabhc:
> I might be misunderstanding but I don't get why it wouldn't be the absolute path to the project directory.

Well, the full path will work to disambiguate it fully, of course. But I feel it might be not too convenient usability wise.

My first concern is around path length and the visual weight it carries. Full paths can be long and showing long strings (with common prefixes) in completion lists and the command line can be bit hard to read. When I think of a project I just think of the base name (e.g. 'howl') and not really the full path, so seeing that as-is in the project list is nice. We can show the full path next to name as a comment (like we do for buffers), but the main name and command line prompt will contain the base name. 

My second concern is that seeing the full path in the command line may (wrongly) indicate to the user that they can backspace to the parent directory, when in fact I'm proposing that backspace go up to the full projects list.

Also, project base names tend to not collide too often, at least for me - so paying the usability price of full paths just for disambiguation may not be worth it.

In any case, these things evolve as we use them so lets try out base names and see how it works?

shalabhc:
@equim  - thanks for the description of an Eclipse workspace. Yes it does seem similar to a Howl 'session'. If we index the project file names, we should be able to provide a *super fast* project-open and also across-project-open. Actually we can already implement a cross project file matcher, but I'm not sure how usable it will be depending on the number of files you have.

@nilnor - have you given thought to the idea of indexing file names and/or contents? A fast project-wide file-grep would be nice too. (Yes I know about `ag` but it still doesn't beat incremental real-time matching across the project). If we also tag structure lines in the index, it could possibly work as a fast jump-to-definition as well.

shalabhc:
Withdrawing this PR. Will send out another one based on the discussions here.

equim:
Agreed that the full path could become a bit unwieldy if it was always required - but can we still specify a full path so we can open a previously unknown project with the same prompt to open a file in that project?

At the moment project-open will first prompt for the project directory if the current buffer file isn't under a known project and then prompt for a file under that project.  The way I imagine the new project-open behaviour, we would instead allow you to open a project and open a file within that project at the same time.  I imagine the portion before the : can be an absolute path to a project that was previously unknown or a shortened path (e.g. just the base name) to an already known project.

Is that how you imagine it too?

shalabhc:
The way I imagined it there would be a 'create new' option in the list of
projects, which would drop you into a directory selection interaction.

Does that sound OK?

equim:
Yeah I think that's OK.  It's one extra key press compared to what we have now but it's not an every day command and it's probably better to be explicit to make it more obvious.

shalabhc:
Yes that's was I was thinking too - selecting 'Create new' will be an explicit action so hopefully more obvious what is is going on.

PR #241, opened 2016-07-13
Fix drag select misbehaving for edge cases

### LINKED ISSUE #240: Drag select misbehaving for edge cases
There are a few related issues regarding mouse selection:

* Drag select to the top of the view selects several lines down.
* Double click then drag doesn't select full words.
* Triple click then drag doesn't select full lines.
* Triple click does not select the line ending, meaning triple click -> delete leaves an empty line.

The first one is a strange bug.  The rest, I believe, do not do what the user expects and are not how most text editors behave.

equim:
Fixes #240.

### DIFF
--- a/lib/aullar/view.moon
+++ b/lib/aullar/view.moon
@@ -774,11 +774,13 @@ View = {
 
   _on_button_release: (_, event) =>
     event = ffi_cast('GdkEventButton *', event)
+    return true if notify @, 'on_button_release', event
     return if event.button != 1
     @_selection_active = false
 
   _on_motion_event: (_, event) =>
     event = ffi_cast('GdkEventMotion *', event)
+    return true if notify @, 'on_motion_event', event
     unless @_selection_active
       if @_cur_mouse_cursor != text_cursor
         if event.x > @gutter_width
@@ -793,8 +795,11 @@ View = {
     pos = @position_from_coordinates(event.x, event.y)
     if pos
       @cursor\move_to :pos, extend: true
-    elseif event.y < 0
-      @cursor\up extend: true
+    elseif event.y < 10
+      if @_first_visible_line == 1
+        @cursor\move_to pos:1, extend: true
+      else
+        @cursor\up extend: true
     else
       @cursor\down extend: true
 
--- a/lib/aullar/view.moon
+++ b/lib/aullar/view.moon
@@ -795,13 +795,16 @@ View = {
     pos = @position_from_coordinates(event.x, event.y)
     if pos
       @cursor\move_to :pos, extend: true
-    elseif event.y < 10
-      if @_first_visible_line == 1
+    elseif event.y < @margin
+      if @first_visible_line == 1
         @cursor\move_to pos:1, extend: true
       else
         @cursor\up extend: true
     else
-      @cursor\down extend: true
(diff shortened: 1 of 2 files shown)


shalabhc:
While this looks ok to me, do you think it may be useful to implement the selection snap to whole word (or whole line) as separate methods on the editor so it is easily testable and usable elsewhere?

lib/aullar/view.moon
```diff
@@ -793,8 +795,11 @@ View = {
     pos = @position_from_coordinates(event.x, event.y)
     if pos
       @cursor\move_to :pos, extend: true
-    elseif event.y < 0
-      @cursor\up extend: true
+    elseif event.y < 10
```
nilnor: I saw the discussion of this on the issue, and yes, having a hard coded `10` here isn't ideal. The explanation for this is the `margin`, so the original code should have been `event.y <= @margin` I believe.

equim:
@shalabhc yeah I think it would be a good idea to extract methods to expand the selection to word/line boundaries - and/or maybe methods for rounding a position to next/previous word/line boundary.  I suspect there's already something similar for cursor-word-left/right commands tho it might not be exactly the same.  I'll look into it.

I forgot to mention that this PR also fixes another misbehaving edge case: drag selecting off the top of the document does not select to the beginning of the first line (unless the mouse cursor happens to be there).  Drag selecting off the bottom of the document should similarly select to the end of the last line - I'll look into that too.

lib/aullar/view.moon
equim: Thanks @nilnor - I've tested and indeed it should indeed be `@margin`.  I felt a bit dirty submitting a PR with a magic number, but I couldn't quite understand what was causing the offset.  I'll fix.

nilnor:
I agree with @shalabhc in that it would be great to get this extracted out of the handler, it would make the code cleaner and as @shalabhc mentioned make it easer to write specs for it. The behavioural changes all look good to me.

lib/howl/ui/editor.moon
```diff
@@ -628,6 +630,38 @@ class Editor extends PropertyObject
...
+  _expand_to_line_starts: (pos1, pos2) =>
+    line1 = @view.buffer\get_line_at_offset pos1
+    line2 = @view.buffer\get_line_at_offset pos2
+    if pos1 <= pos2
+      @buffer.lines[line1.nr].start_pos, @_next_line_start(line2)
+    else
+      @_next_line_start(line1), @buffer.lines[line2.nr].start_pos
+
+  _next_line_start: (line) =>
+    next_line_nr = line.nr + 1
```
shalabhc: `line.next` could be used here for brevity

equim:
Thanks to the suggestion to use line.next from @shalabhc I looked a bit deeper into what "line" I actually had and then ultimately found that I was mixing types/concepts from aullar with howl.  I was calling lower level methods in aullar to get byte offset from coordinates and then line from byte offset.  I was lucky that the triple click drag logic used byte offsets consistently - but drag select and double click drag select did not meaning they were broken when handling positions after multi-byte chars.

Latest commit fixes it and is cleaner now I can follow @shalabhc's suggestion.

nilnor:
All right, looks good, thanks!

PR #242, opened 2016-07-14
Command history persistence

shalabhc:
Implements command line history persistence.

This saves the command line history and reloads it when Howl is started so pressing `up` after `alt_x` will show entries from the previous run. History is synced every time any command is run, so a crash or kill of the Howl process will leave the history saved as well. The saving mechanism attempts to merge new history with saved history so multiple running instances of Howl should not clobber each other's history, but instead produce a merged history.

This also implements full deduplication of history. Rerunning any historical command removes the old record and creates a new one.

### DIFF
--- a/lib/howl/ui/command_line.moon
+++ b/lib/howl/ui/command_line.moon
@@ -25,9 +25,10 @@ class CommandLine extends PropertyObject
     @showing = false
     @spillover = nil
     @running = {}
-    @_command_history = {}
     @aborted = {}
     @next_run_queue = {}
+    @_command_history = {}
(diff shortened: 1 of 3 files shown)


nilnor:
There's no pruning as of yet, but there should be I think - particularly if we want to sync the file with every command run.

shalabhc:
Ah right - overlooked that.  I'll add that along with a config option.

shalabhc:
Added `config.command_history_limit` that defaults to 100. The most recent 100 unique commands are persisted.

nilnor:
:+1:

PR #245, opened 2016-07-20
Don't scroll to bottom on select all

### LINKED ISSUE #244: Select all scrolls to bottom
The editor-select-all command scrolls to the bottom of the file.  This is different to most editors I've used.

I often select all to copy the contents of a file elsewhere.  Having it scroll to the bottom makes me lose my place.

equim:
Fixes #244.

### DIFF
--- a/lib/howl/ui/selection.moon
+++ b/lib/howl/ui/selection.moon
@@ -74,7 +74,6 @@ class Selection extends PropertyObject
     @set start_pos, end_pos
 
   select_all: =>
-    @_view.cursor.pos = @_buffer.size + 1
     @_sel\set 1, @_buffer.size + 1
 
   range: =>
--- a/lib/aullar/view.moon
+++ b/lib/aullar/view.moon
@@ -774,11 +774,13 @@ View = {
 
   _on_button_release: (_, event) =>
     event = ffi_cast('GdkEventButton *', event)
+    return true if notify @, 'on_button_release', event
     return if event.button != 1
     @_selection_active = false
 
   _on_motion_event: (_, event) =>
     event = ffi_cast('GdkEventMotion *', event)
+    return true if notify @, 'on_motion_event', event
     unless @_selection_active
       if @_cur_mouse_cursor != text_cursor
         if event.x > @gutter_width
@@ -793,8 +795,11 @@ View = {
     pos = @position_from_coordinates(event.x, event.y)
     if pos
       @cursor\move_to :pos, extend: true
-    elseif event.y < 0
-      @cursor\up extend: true
+    elseif event.y < 10
+      if @_first_visible_line == 1
+        @cursor\move_to pos:1, extend: true
+      else
+        @cursor\up extend: true
     else
       @cursor\down extend: true
 
--- a/lib/aullar/view.moon
+++ b/lib/aullar/view.moon
@@ -795,13 +795,16 @@ View = {
     pos = @position_from_coordinates(event.x, event.y)
     if pos
       @cursor\move_to :pos, extend: true
-    elseif event.y < 10
-      if @_first_visible_line == 1
+    elseif event.y < @margin
+      if @first_visible_line == 1
         @cursor\move_to pos:1, extend: true
       else
         @cursor\up extend: true
     else
-      @cursor\down extend: true
+      if @last_visible_line == @buffer.nr_lines
+        @cursor\move_to pos:@buffer.size+1, extend: true
+      else
+        @cursor\down extend: true
 
   _on_scroll: (_, event) =>
     event = ffi_cast('GdkEventScroll *', event)
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -75,6 +75,8 @@ class Editor extends PropertyObject
     listener =
       on_key_press: self\_on_key_press
       on_button_press: self\_on_button_press
+      on_button_release: self\_on_button_release
+      on_motion_event: self\_on_motion_event
       on_focus_in: self\_on_focus
       on_focus_out: self\_on_focus_lost
       on_insert_at_cursor: self\_on_insert_at_cursor
@@ -660,6 +662,9 @@ class Editor extends PropertyObject
     return true if bindings.process event, 'editor', maps, self
 
   _on_button_press: (view, event) =>
+    @drag_press_type = event.type
+    @drag_press_pos = @view\position_from_coordinates(event.x, event.y)
+
     if event.type == Gdk.GDK_2BUTTON_PRESS
       group = @current_context.word
       group = @current_context.token if group.empty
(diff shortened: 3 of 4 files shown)


equim:
Hrmph, this PR includes #241 commits.  Sorry, I can do a clean PR if wanted.

lib/howl/ui/selection.moon
```diff
@@ -74,7 +74,6 @@ class Selection extends PropertyObject
     @set start_pos, end_pos
 
   select_all: =>
-    @_view.cursor.pos = @_buffer.size + 1
```
shalabhc: You may want to remove or adjust the corresponding spec.

equim:
Having the cursor in the middle of the selection does go against all the usual editor selection norms.  Most (all?) editors move the cursor to the the beginning or end of the file.

For your example, @shalabhc, with this patch pressing shift+left or shift+down etc moves the cursor in the normal way and changes the selection to start from the beginning of the file but end at the cursor.  That behaviour is probably unintuitive and not very useful.  At the same time if the cursor were still placed at the end I'm not sure it's very useful to be able to do something like shift+left to select all but the last character.

It still seems more useful to me to leave the cursor where it is, allowing an easy and natural way to 'undo' the select all from the keyboard, but perhaps the cost of it appearing strange to the user is too great.  Perhaps it's better to create a new PR that fixes the scroll issue while still moving the cursor to the bottom.

I've gone through checking some other editors to see what they do to the cursor pos and scroll pos when selecting all.  Since some hide the cursor when there is a selection I also checked what happened when pressing left or right arrow keys to see where the cursor would end up.  Interestingly most do something different to howl here too - moving the cursor to the beginning/end of the selection before the selection is cleared:

Editor                | Select all cursor pos | Select all scroll pos | Left key with selection | Right key with selection
----------------------|-----------------------|-----------------------|-------------------------|-------------------------
SciTE                 | Start of file         | Unchanged             | Start of selection      | End of selection
Textadept             | Start of file         | Unchanged             | Start of selection      | End of selection
Atom                  | End of file           | Unchanged             | Start of selection      | End of selection
VS Code               | End of file           | Unchanged             | Start of selection      | End of selection
Eclipse               | End of file           | Unchanged             | Start of selection      | End of selection
KDevelop              | End of file           | End of file           | Cursor left             | Cursor right
Gedit                 | Invisible             | Unchanged             | Start of selection      | End of selection
Firefox text area     | Invisible             | Unchanged             | Start of selection      | End of selection
Howl                  | End of file           | End of file           | Cursor left             | Cursor right
Howl with this patch  | Unchanged             | Unchanged             | Cursor left             | Cursor right

shalabhc:
Wow - extensive research!

Perhaps we can also consider other ways in which we could address the same issue. The problem right now IIUC is the cursor position (and scroll context) are lost when you do a select-all. I noticed even pressing `esc` doesn't cancel the selection and revert the cursor position, maybe something it should do.

One possible feature is to drop a breadcrumb for the cursor location on specific events. Then have a shortcut to 'go back' the trail of breadcrumbs. 'select-all' could be one command that adds the breadcrumb, but this feature may be useful in other commands too (e.g. `cursor-start|eof`). The downside is the user has to remember this shortcut to walk back. I think this exists in emacs: https://www.gnu.org/software/emacs/manual/html_node/emacs/Mark-Ring.html

Also, if we had a command for 'buffer-copy' then maybe it would handle your use case and not lose the context at all.

equim:
Yeah, leaving a breadcrumb might be a good idea.  Some of the editors I've used - usually the ones towards the IDE end of the spectrum - have forward and back commands (typically bound to alt+right / alt+left) for navigating through the view history which can include going back to previous files as well as to previous locations in the current file which I guess is like the emacs mark ring you mentioned.

Howl already has the switch-buffer command which I like for going back to a previously viewed buffer.  Perhaps there should be another command (switch-location?) that would be the same but includes the line and column number for each and can have multiple entries per file.

That kind of thing would become even more useful if/when we start getting more navigation features such as go to definition.

A buffer-copy command could be another interesting workaround though I'm not confident I'd remember it when I need it since it's a fairly niche thing - as opposed to navigating within/between buffers.

I'm happy to close / ignore this PR for now.

shalabhc:
> Perhaps there should be another command (switch-location?) that would be the same but includes the line and column number for each and can have multiple entries per file.

I like this idea a lot. This is somewhat related to bookmarks, but it is automatic and not explicit. I often bounce around between a handful of locations across files at one time and quickly jumping to one of them seems very useful.

A shorctut for 'go back' in particular seems like a good idea too.

Is this something you are interested in doing @equim?

equim:
Yeah I'm interested in doing it, but probably won't get to it short term.  I now have a list of 83 things I'd like to do for howl!  Most of them are much smaller changes than this so they are probably better for easing me into the code base.

shalabhc:
Sounds good. Looking forward to your other changes!

nilnor:
So, back to the original change for this PR: I'm actually not that concerned with Howl being the one slightly out of place with having the cursor stay in place, especially as the other editors in @equim's excellent research table aren't agreed with what to do. So, anyone against merging this PR?

Related to the research, we might want to change the cursor left/right behaviour though to match the majority of the other editors in moving the cursor to the start or end of the selection.

nilnor:
Actually, disregard the last part of my last comment - if we keep the cursor where it is it makes sense that cursor left and right ends up right to the current location.

equim:
Yeah, the cursor left/right moving to selection start/end would reduce the effectiveness of this PR - it would become just a way to maintain the scroll position.

We could perhaps still have both this PR and make cursor left/right move to selection start/end only if the cursor is already at one of the selection ends (as it would be in every case except select-all for now).

One thing I thought about when first making this change was that perhaps this new select-all cursor behaviour would be useful for other selection commands in the future, e.g. an expand-selection command that would first select the current word (or expand to word boundaries if there is already a selection), then the next expand-selection would expand to select the ()/[]/{}/""/'' delimited block the cursor is in, etc. Preserving the cursor position in the middle of the selection becomes useful then with a matching collapse-selection command that would gradually reduce the selection down through the blocks the cursor is in.  If the cursor position wasn't preserved we wouldn't know which block to collapse the selection to if there were multiple blocks within the selection.

I'm still not sure if cursor in middle of selection is a good idea - I like the extra functionality but it's not good if people think it's a bug when they see it.

I'll have a go at implementing those expand-selection/collapse-selection commands so we have another potential use case.

shalabhc:
While I'm not against the cursor not consistently being at the end of the selection, I wanted to point out that the API exposes `.anchor` and `.cursor` properties on the selection, which could confusing?

I think the cursor being mid-selection is fine as long as we follow a consistent guideline. For instance, we can say the cursor is at the end of the selection when one of the cursor motion commands is used to extend the selection (`cursor-up-extend`, etc). Commands that select something directly and don't have a corresponding cursor motion don't move the cursor (`select-all` and possibly `select-line`, `expand-selection` etc. in the future).

As an aside, having `left` and `right` move the cursor to the corresponding end of the selection sounds like a great idea. I like the suggestion from @equim that this should only happen if cursor is already at one end of the selection.

nilnor:
@shalabhc good point about the `cursor` property. I think it's useful to be able to have the cursor in the middle of a selection (for instance VI's line visual mode actually requires it to work properly), but the `cursor` name becomes confusing in these cases. It probably makes sense to look at renaming this at some point to something slightly more generic.

@equim yes, I agree with Shalabh that the idea of only doing the selection left/right handling when at one of the boundaries makes sense.

Anyway, let's go with this for now - should people believe it to be a bug then we can revisit it, but the functionality itself is actually useful for other cases as well (as the aforementioned VI mode).

PR #251, opened 2016-07-29
Fix location selection for all line numbers.

shalabhc:
Preview was only loading the first 8K bytes which did not guarantee
that line numbers in select_location would be covered. This forces
loading the entire file if a line number is given.

### DIFF
--- a/lib/howl/interactions/location_selection.moon
+++ b/lib/howl/interactions/location_selection.moon
@@ -19,7 +19,8 @@ interact.register
       preview = Preview!
       opts.on_change = (selection, text, items) ->
         if selection
-          buffer = selection.buffer or preview\get_buffer selection.file
+          buffer = selection.buffer or preview\get_buffer selection.file,
+            full: selection.line_nr and selection.line_nr > 1
           editor\preview buffer
           if selection.line_nr
             editor.line_at_center = selection.line_nr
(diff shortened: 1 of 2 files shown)


nilnor:
But the reason for having the limit in the first place was to avoid hang-like lags for previewing huge files, and with this we have that problem again? Could we avoid this by doing some kind of staggered load, or perhaps just showing some text indicating that the file is unviewable similarly to how we do it for binary files?

shalabhc:
io.File does not support async reading at the moment so there's no good way to avoid the hang if we read the full file afaict.

Note that `select_file` behavior is unchanged - i.e. still only loads 8K. Only if you use `select_location` *with* a line number in an item, it will load that file fully. The trade-off is you either get a quick preview with and error (because the line isn't loaded) or you get a full preview that works, which may or may not be slow, depending on file size.

shalabhc:
So I'm looking at `io.File` and `gio.File` and it seems all the underlying pieces to implement non blocking reads are in place. However `io.File` uses Lua's 'io' module in a couple of methods (`read` and `open`) and the blocking `gio_file_load_contents` in `.contents`.

I can try to make `File.read` non blocking and switch `.contents` to use that. What are your thoughts on the API here?

nilnor:
(for reference we had some out of band discussion on this, with the outcome that some more things would be tried for the content loading).

shalabhc:
The next step is to try switching some File api methods to use async io underneath. That would be a separate PR so closing this one.

PR #253, opened 2016-07-30
Makes toggle-comment comment if any line not commented

equim:
Toggle comment with multiple selected lines only considered the first line
when deciding whether to comment or uncomment.  If the first line was
commented then it attempted to uncomment all lines - leaving already
uncommented lines unchanged.

### DIFF
--- a/lib/howl/modes/default_mode.moon
+++ b/lib/howl/modes/default_mode.moon
@@ -94,7 +94,13 @@ class DefaultMode
     return unless prefix
     pattern = r"^\\s*#{r.escape prefix}.*"
 
-    if editor.active_lines[1]\umatch pattern
+    all_active_lines_commented = true
+    for line in *editor.active_lines
+      unless line\umatch pattern
+        all_active_lines_commented = false
+        break
+
+    if all_active_lines_commented
       @uncomment editor
     else
       @comment editor
--- a/lib/howl/modes/default_mode.moon
+++ b/lib/howl/modes/default_mode.moon
@@ -96,7 +96,7 @@ class DefaultMode
 
     all_active_lines_commented = true
     for line in *editor.active_lines
-      unless line\umatch pattern
+      if not line\umatch(pattern) and not line.is_blank
         all_active_lines_commented = false
         break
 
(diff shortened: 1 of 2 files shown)


nilnor:
I'm not so sure about this, as the current behaviour has it's advantages. For example, given a code snippet such as:

```mooonscript
important = 2
-- the following is needed for blah blah
other = important * 2 / 3
```

When coding, it happens that one wants to comment out an entire block. The current behaviour works fine for this - the previously uncommented lines becomes commented and the already commented one gets a double comment. When un-commenting later everything is restored correctly with existing comments and all, whereas with this change one would just end up with broken code. In short, I'm not convinced this new behaviour is better, but am as always open to persuasive arguments :)

equim:
Yes, toggle-comment with those 3 lines selected should result in them all getting commented out (with double comment on the second line) and then toggle-comment on those 3 lines again should reverse the first toggle-comment leaving us with the original 3 lines.  This PR doesn't change that.

The one case where this PR should have an affect is where you have code like:

```moonscript
-- the following is needed for blah blah
important = 2
other = important * 2 / 3
```

and you toggle-comment with all three lines selected.  With the old behaviour it would uncomment the first line and leave the other 2 lines alone.  With this PR all lines will be commented out (with the first line double commented).  The reason I wouldn't just select the last 2 lines to toggle-comment is that then I would have all 3 lines comments out once and I would have to pay attention to what I was uncommenting to avoid uncommenting the non-code comment.

Unfortunately, after double checking the scenario you described I found this new PR gives undesirable behaviour with code like:

```moonscript
important = 2

-- the following is needed for blah blah
other = important * 2 / 3
```

Selecting those 4 lines and doing toggle-comment comments out all lines except the blank one.  Then toggle-comment on the 4 lines again comments them out again since it detected that one line wasn't commented out.

I'll fix this by detecting and ignoring the blank line in toggle-comment.

nilnor:
Sorry, I read that diff way to fast. Good catch re the blank lines, apart from that I think it all looks great!

nilnor:
Great, thanks!

PR #261, opened 2016-09-29
Windows

### LINKED ISSUE #201: Windows build
[](https://gfycat.com/DesertedSpryBobcat)

Changes: https://github.com/howl-editor/howl/commit/7a551fbc3b2a49da0fc5e3bf4219b2198b8863ab

Related to #100.

I was interested in doing this port for fun, so this changeset is just to show what I did.  I did not attempt to run the tests, I only opened some files in the editor.

The build environment is [msys2's](http://msys2.github.io/) `mingw32` environment.  I had to change LuaJIT's Makefile to build the .a file (change `BUILDMODE` in src/Makefile from `mixed` to `static`).  There are a ton of gtk dlls that would need to be packaged to make a release out of it.

The biggest pain by far is the fact that LuaJIT's ffi doesn't function the same on Windows as it does on Posix systems with regards to being able to load a name without knowing the module it comes from.  I solved this hackily by adding over 9000 exports to the output binary (objdump and regex).  gcc doesn't like this as much as I don't like this.

refi64:
Closes #201. I still have to figure out why the tests are failing on Travis, but I figured I might as well put this up so that you guys can review.

TL;DR of this pull request: Windows is weird.


kirbyfan64:
Build's green now! :D

I was thinking we should probably add AppVeyor CI to make sure Windows doesn't break; it has support for [MSYS2](http://help.appveyor.com/discussions/suggestions/615-support-for-msys2/page/1).

lib/ljglibs/gobject/object.moon
```diff
@@ -7,9 +7,9 @@ C, ffi_cast, ffi_new = ffi.C, ffi.cast, ffi.new
 lua_value = types.lua_value
 
 core.define 'GObject', {
-  new: (type) ->
+  new: (type, ...) ->
```
nilnor: Good catch! There's an additional twist when invoking functions though - the arg list needs to be `nil`-terminated or bad things will happen (see e1bec1c).

spec/buffer_spec.moon
```diff
@@ -1,6 +1,8 @@
 import Buffer, config, signal from howl
 import File from howl.io
 import with_tmpfile from File
+ffi = require 'ffi'
+ffi.cdef [[ void g_usleep(unsigned long ms); ]]
```
nilnor: Let's move the cdef into the ordinary `glib.moon` cdef file inside of ljglibs instead.

nilnor (commented):
Very happy to see this happen, kudos! :)

lib/howl/io/process.moon
```diff
@@ -77,7 +95,7 @@ pump_stream = (stream, handler, parking) ->
       handler ret
       if ret == nil
         stream\close!
-        dispatch.resume parking
+        pcall dispatch.resume, parking
```
kirbyfan64: Whoops, that was just left behind when I was debugging the EOF errors.

lib/howl/io/process.moon
```diff
@@ -147,13 +175,15 @@ class Process
       stderr = {}
       on_stderr = (err) -> stderr[#stderr + 1] = err
 
-    stdout_done = on_stdout and dispatch.park "process-wait-stdout-#{@pid}"
-    stderr_done = on_stderr and dispatch.park "process-wait-stderr-#{@pid}"
-    pump_stream(@stderr, on_stderr, stderr_done, true) if on_stderr
-    pump_stream(@stdout, on_stdout, stdout_done) if on_stdout
+    @stdout_done = on_stdout and dispatch.park "process-wait-stdout-#{@pid}"
```
kirbyfan64: Oddly enough, the code seems to work now without my changes. I don't actually know why exactly it didn't work before, but I'm guessing something else I did fixed this...

lib/howl/init.lua
```diff
@@ -162,12 +164,27 @@ local function main(args)
...
   require('howl.globals')
+
+  local font_dir = table.concat({app_root, 'fonts'}, path_separator)
+  if ffi.os == 'Windows' then
+    require 'howl.cdefs.windows'
+    local fonts = howl.io.File(font_dir).children
+    for _, font in ipairs(fonts) do
+      local loaded = ffi.C.AddFontResourceExA(font.path, ffi.C.fr_private, nil)
+      if loaded == 0 then
+        print('failed to load font ' .. font.path)
```
kirbyfan64: @nilnor Still needs the flush. Not really sure why...

lib/howl/io/process.moon
```diff
@@ -127,7 +148,14 @@ class Process
 
   send_signal: (signal) =>
     signal = signals[signal] if type(signal) == 'string'
-    C.kill(@pid, signal)
+    if jit.os == 'Windows'
```
kirbyfan64: @nilnor Good point. I was thinking that I could add a table `platform` to `howl.sys`. It would have the following methods/attributes/functions/you-get-the-point:

- `platform_tmpname` from `howl/io/file.moon`, under the name `tmpname`.
- `fd_to_stream(win_type, unix_type, fd)` from `howl/io/input_stream.moon` and `howl/io/output_stream.moon`. It could be called like `fd_to_stream(Win32InputStream, UnixInputStream, fd)`, and it would return the correct type, converting the file descriptor as necessary. Alternatively, it could be `fd_to_stream(kind, fd)` and be called like `fd_to_stream('Input', fd)` and automatically index the streams from `ljglibs.io` (like `require 'ljglibs.io'["#{platform}#{kind}Stream"]`), but I'm not sure if you would consider that too magical.
- `default_shell` and `send_signal` from `howl/io/process.moon`.
- `load_font` from the font-loading code in `howl/init.lua`.

I'm not sure whether the font name code in `font_awesome.moon` would really fit in `howl.sys`, though, as well as the file separator handling code in `file.moon`.

How does all that sound?

README.md
```diff
@@ -30,6 +30,8 @@ how to install Howl from source.
 based system).
 - `C compiler`: Howl has a very small C core itself, and it embeds
 dependencies written in C.
+- If you're using Windows, you *need* to build Howl inside of
```
kirbyfan64: Should it be in a readme (e.g. `README.windows.md`), or do you think it would be better as a page on the site (e.g. `getit.windows.html`), as a parallel to `getit.html`?

nilnor: Perhaps a separate page on the site would be best, we could just reference to that from the README.

lib/howl/io/process.moon
```diff
@@ -108,17 +124,20 @@ class Process
...
     error 'opts.cmd missing or invalid', 2 unless @argv
     @_process = launch @argv, opts
+    @true_pid = @_process.true_pid
     @pid = @_process.pid
     @working_directory = File opts.working_directory or get_current_dir!
     @stdin = OutputStream(@_process.stdin_pipe) if @_process.stdin_pipe
     @stdout = InputStream(@_process.stdout_pipe) if @_process.stdout_pipe
     @stderr = InputStream(@_process.stderr_pipe, PRIORITY_LOW - 10) if @_process.stderr_pipe
+    @stdout_done = nil
+    @stderr_done = nil
```
nilnor: This is now old cruft

lib/howl/io/process.moon
nilnor: I think that sounds good, that will clean up the diffs a lot.

lib/ljglibs/gio/win32input_stream.moon
```diff
@@ -0,0 +1,22 @@
+-- Copyright 2014-2015 The Howl Developers
+-- License: MIT (see LICENSE.md at the top-level directory of the distribution)
+
+require 'ljglibs.gio.input_stream'
+ffi = require 'ffi'
+core = require 'ljglibs.core'
+{:ref_ptr} = require 'ljglibs.gobject'
```
nilnor: We need to run this PR through the new linter later before merging btw to catch all of these things (e.g. unused var above).

nilnor:
Re AppVeyor, that would be great. The specs are bound to break pretty fast by mistake otherwise.

kirbyfan64:
Well, at some point I guess I broke processes again...ugh...

kirbyfan64:
Wow, this is really, really weird...

So, `TerminateProcess` kind of kills the process, except it doesn't. Apparently, the process gets "killed", but it doesn't actually stop running. If I run something like `sleep 2; touch`, the exit callback is called, but `stdout` and `stderr` stay waiting for `sleep 2` to finish. But `touch` is never run. So it appears the shell is killed, but it waits for the first process to finish first.

And it seems that the exit callback could somehow potentially be called before the process data is read somehow? Now I remember why I put that weird `pcall`...

mfxmfx:
I would like to try your Windows patch (and possibly add support for cross-compilation in .travis.yml).

Could you please rebase your branch on current master? Thanks.

shalabhc:
@kirbyfan64 - is this PR still good? Maybe we can merge even if not *everything* on windows is fixed?

kirbyfan64:
@shalabhc Uhh...yeah...

I *still* can't figure out what the hell is going on with processes. Meanwhile, I just shoved a nasty retry-three-times loop when the troublesome error appears. Messy, but [Windows I/O seems to generally only be solvable with retry loops](https://github.com/python/mypy/pull/3239).

As an aside, here's a zip file with the binaries:

https://www.dropbox.com/s/3d3extidx56h1t7/howl.zip?dl=0

(created using the new `make windist`) and a nice, shiny Inno Setup installer:

https://www.dropbox.com/s/7whewn5d43ug6o8/howl-setup.exe?dl=0

soapdog:
@kirbyfan64 tried both the zip file and the installer on my windows 10 machine. The executable just exited without any error, nothing appeared on the screen. (PS: I don't have msys2 installed, thought maybe it is not needed for running it)

kirbyfan64:
Okay guys, this is really, really bad...

### The problem

Basically, everything's broken again, especially since I got a new computer with an upgraded MSYS2 (since [the old one kinda...met an unfortunate end](https://refi64.com/posts/when-replacing-a-hard-drive.html)):

- Icons don't work anymore.
- glib process crud is still broken.
- [Pixbufs don't work](https://twitter.com/refi_64/status/922208599081979905), even though I can see the symbols present in the resulting binary.
- There's this persistent, ugly border that I can't get rid of.
- I have legit tried to make a binary distribution for several hours, and I still cannot figure out why some (again pixbuf-related) DLLs aren't getting picked up.

Here's the fundamental problem I've noticed: *using MSYS for this is never going to work correctly*. The entire concept of a UNIX environment that's not really UNIX inside of Windows has caused nothing but endless pain, and I'm incredibly sick of trying to mess with it. It surprises me that these broken binaries are what the GTK+ project recommends using...

### The new plan

Screw MSYS2.

I've stumbled upon [this GTK+ building project](https://github.com/wingtk/gvsbuild). The new idea is to instead use Visual Studio for building. I have no clue how this is going to end up, *but it can't possibly be any worse*.

### Meanwhile...

I literally set up Howl with my entire custom config to work *flawlessly* under Windows Subsystem for Linux w/ VcXsrv in under an hour. I'll probably write up somewhere how to install it, but if you're looking for a quick way to get Howl working under Windows until I get everything here working, **this is probably the way to go**.

nilnor:
@kirbyfan64 it sounds like you've had good success with the subsystem for Linux and VcXsrv. Could we (and I really mean you :)) do a detailed write-up of this that we can put on the site as installation instructions?

nilnor:
Thanks for this — it was a serious piece of work, and it's on me that it never got a decision.

Ten years on, master has diverged far past the point where this can be rebased. The two changes that matter most here:

- **The Gtk-4 port.** The whole UI layer moved from Gtk+ 3 to Gtk 4. Files this PR touches are gone or unrecognisable: `lib/howl/ui/theme.moon` and the `.moon` themes were replaced by 100% CSS themes, `lib/howl/ui/content_box.moon` and `lib/ljglibs/util/background.moon` no longer exist, several ljglibs modules this builds on were removed (`gtk/container`, `gtk/bin`, `gtk/event_box`, `gdk/atom`, `gtk/clipboard`, ...), and `lib/aullar/view.moon` / `gutter.moon` were substantially rewritten.
- **The command interaction refactor.** `lib/howl/interactions/` was replaced by the new `interact` API, explorers and the command panel, so `interactions/file_selection.moon` and its spec are gone.

On top of that the cdefs, process/IO and spec layers have all moved, the LuaJIT baseline changed (`luajit-2.1-2025-04-09`), and CI moved from Travis to GitHub Actions — so the failing-Travis question in the description no longer has a meaningful answer either.

Closing as unmergeable rather than unwanted. Windows support is still a legitimate goal, but it would have to be redone from scratch against current master rather than rebased. The platform-specific pieces here (`lib/howl/cdefs/windows.moon`, the win32 stream bindings, `src/process_helpers.c`, the installer script) remain a useful reference for anyone taking that on, and the PR stays readable in the archive.

For context: master is currently unreleased and Gtk-4 based — `0.6` is the last release and `0ca4ffa2` the last Gtk-3 master.

PR #271, opened 2016-10-15
Add a buffer-next command mapped to Ctrl+Tab

paly2:
Hello ! Sorry for my bad English ;-)

I'm new to howl, and I found that switching between buffers was quite long (you need to press Ctrl+B, then to select the buffer, and then to press Enter). So I tried to create a very fast-to-use buffer-next command, mapped to Ctrl+Tab (the view-next command being remapped to Ctrl+Shift+Tab, I use it less often because using the mouse to change active view is often faster). Note that I have never programmed in MoonScript before, so I guess that my code is very dirty.

This PR proposes this change to merge.

Note: I love howl, because it is simple but powerful !

### DIFF
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -46,6 +46,17 @@ command.register
   handler: (buf) -> app.editor.buffer = buf
 
 command.register
+  name: 'buffer-next',
+  description: 'Jump to the next buffer'
+  handler: ->
+    buffers = moon.copy app.buffers
+    table.sort buffers, (a, b) -> a.title < b.title
+    for i, buffer in ipairs buffers
+      if buffer.title == app.editor.buffer.title
+        app.editor.buffer = buffers[i % #buffers + 1]
+        break
+
+command.register
   name: 'buffer-reload',
   description: 'Reload the current buffer from file'
   handler: ->
--- a/lib/howl/keymap.moon
+++ b/lib/howl/keymap.moon
@@ -39,6 +39,7 @@
     ctrl_shift_end:   'cursor-eof-extend'
 
     ctrl_b:           'switch-buffer'
+    ctrl_tab:         'buffer-next'
     ctrl_c:           'editor-copy'
     ctrl_d:           'editor-duplicate-current'
     ctrl_shift_e:     'cursor-goto-inspection'
@@ -65,7 +66,7 @@
     ctrl_q:           'show-doc-at-cursor'
     ctrl_space:       'editor-complete'
     ctrl_slash:       'editor-toggle-comment'
-    ctrl_tab:         'view-next'
+    ctrl_shift_tab:   'view-next'
 
     shift_insert:     'editor-paste'
     ctrl_insert:      'editor-copy'
@@ -121,6 +122,7 @@
       editor: {
         meta_shift_a:     'editor-select-all'
         meta_b:           'switch-buffer'
+        meta_tab:         'buffer-next'
         meta_c:           'editor-copy'
         meta_d:           'editor-duplicate-current'
         meta_f:           'buffer-search-forward'


kirbyfan64:
Honestly, I have never used view-next, but I could definitely see going to the next buffer being quite useful!

shalabhc:
@paly2 - did you try `switch-to-last-hidden-buffer`? Most of the time, that is what I want, and not switching to the next buffer in alphabetical order.

BTW, I do agree we need a better shortcut for that and `ctrl_tab` seems reasonable to me.

paly2:
@shalabhc I did try it, but is allows you to switch between only two buffers... It is often what you want to do, but when you want to switch between three or more buffers, you need the long ^B - selecting - Enter method.

Maybe you could map by default `switch-to-last-hidden-buffer` to `ctrl_tab`, `buffer-next` to `ctrl_shift_tab`, and removing the default shortcut for `view-next` which is useless.

shalabhc:
`buffer-next` seems to go through the buffers in alphabetical order, is that correct?

I think a slightly different order may be better - go through the buffers in the last-seen order. The problem or course is it would just switch between two buffers. What I really want is if press `ctrl_tab` twice in succession, it go to the third buffer in the buffers list. So if my buffers list looks like this:

A <-- current
B
C

I press `ctrl_tab` twice, I want to go to C. Then my buffer list looks like this:

C <-- current
A
B

I press `ctrl-tab` once and can go back to A.

Would that work for you? I'm not quite sure the best way to implement it is though.

paly2:
@shalabhc Yes, that's exactly what I need !

mfxmfx:
FYI, I have a switch-to-{next,prev}-buffer implementation that uses a sequential Buffer.id for sorting much like @shalabhc is proposing in his [config](https://github.com/howl-editor/howl/tree/config) branch.

Useful when starting with something like "howl foo.h *.c" . Implementation is pretty straightforward.

shalabhc:
Ah yes, I needed the id to generate a unique key for any buffer not associated with a file.

Back to this issue, currently there is no way to handle pressing `ctrl_tab` twice without releasing the `ctrl` key. I'm not even sure what the best way to even represent that in the `bindings` API is.

Also the state transition would look something like [buffer-order1] -> [tabbing through buffer-order1] -> [commited to a buffer, order changed to buffer-order2]

Basically I'm saying I'm not sure where to start implementing this idea, but I'll give it more thought.

shalabhc:
One way this *may* work is if first the key handling detects repeated keypresses. E.g. If I press `ctrl_a`, release `a` while holding down `ctrl`, then press `a` again, it should be identified as a repeat and not a new keystroke. This could be then fed into the keymap handler, which then could pass it as an argument to a command.

Another way I think this might work is the key-up state is also invoked on keymaps. So you could have a new interaction which implements `ctrl_keyup`: -> self.finish!` - basically ending the  interaction as soon as the ctrl key is released.

shalabhc:
@kirbyfan64, @nilnor - any thoughts on the above ideas?

nilnor:
Thinking about this, the desired outcome here seems to be to have some way to cycle between buffers in the last viewed order, without requiring one to use multiple/different key strokes. If it would be OK to use a list for this, like `switch-buffer`, then yet another approach would be to have the triggering key stroke also cycle between list entries. E.g. `ctrl_b` brings up the list and subsequent `ctrl_b`s cycle through items. By default releasing the keys would currently cancel everything though, but perhaps it would be OK with one final `enter` being required?

I think identifying repeated key strokes is interesting as well, and right now more so than identifying key-ups.

shalabhc:
I think the desired use case here is minimizing keystrokes while being able to switch:

1) Switch to the next buffer with 1 keystroke (press ctrl_b, release ctrl_b)
2) Switch back to the first buffer with 1 keystroke (press ctrl_b, release ctrl_b)
3) Switch to the next+1 buffer with 2 keystrokes (press ctrl_b, release b but hold ctrl, press b again, release ctrl_b)

It's basically how alt_tab works with apps on Ubuntu.

shalabhc:
So I was playing with the new breadcrumbs branch that @nilnor recently added and noticed it drops breadcrumbs on `buffer-switch`, which is great. What this means is if you use `buffer-switch` to go from buffer A to buffer B and then to buffer C, you can then navigate between these three buffers using the `navigate-back` and `navigate-forward` commands (which could be bound shortcuts). The commands also navigate to last position within the current buffer, if you used search, so it's not strictly between buffers. 

Maybe we could add new navigation command `navigate-next-buffer` and `navigate-previous-buffer` - which jump to the closest breadcrumb that is in another buffer? This might satisfy the fast buffer switching use case pretty well, without changing the underlying model.

nilnor:
@shalabhc: Yes, that's actually something that will work pretty well I think, and the support for this shouldn't be to hard to add once the first version of breadcrumbs is in place.

shalabhc:
I'm going to close this PR as obsolete. The better way to implement this is now using the breadcrumbs feature as described in the comments above.

PR #273, opened 2016-10-20
Update the font's full name for osx

### LINKED ISSUE #68: OS X support

hmadison:
While working on getting Howl to a state where it can live as an application bundle for OSX, I noticed that the font name was incorrect for the way it is referenced. Currently, the font's full name is set internally as "FontAwesome", but [referenced as](https://github.com/howl-editor/howl/blob/6002d9910b917737bc1249cde0673df03b58108c/lib/howl/ui/icons/font_awesome.moon#L686) "Font Awesome". 

I changed the font's name to "Font Awesome". Now, when used in combination with `ATSApplicationFontsPath` the icons [display correctly](http://imgur.com/a/z8mdb) without modification.

This partially resolves #68.


nilnor:
Great, thanks!

PR #276, opened 2016-11-10
Clean up .travis.yml now that luarocks is gone.

mfxmfx:


nilnor:
Great, thanks!

mfxmfx:
Thanks for the quick merge.

I've only recently discovered howl after using textadept for the last 2 years, and I'm quite impressed!

PR #278, opened 2016-11-10
VI: implement 'N': repeat the last search in the opposite direction

mfxmfx:


mfxmfx:
Hi,

this is my first small contribution, to get an idea about your workflow:

- do you want me to create a new command for patches like these?
- I have not added a spec case because there was none for 'n' either, and I'm new to Moonscript

Cheers,
Markus

nilnor:
It's not possible to categorically say that commands should be added but it's on a case by case basis. A general rule of thumb would be that if it's a generic and worthwhile addition, and not VI specific, than it probably makes sense to add it as a command. 

In this case I think a new command is warranted, e.g. `buffer-repeat-search-opposite`. It probably makes sense to move the implementation to the searcher itself, i.e. `Searcher#repeat_last_opposite`. And yes, the VI spec is lacking. As an excuse I'll point to the fact that it was one of the first ones. The searcher has a more comprehensive spec though, so no excuse there :)

Apart from this this looks good, I am always happy to see VI improvements.

mfxmfx:
I've started work on improving vi_spec - please see my other pull request #285 .

And are you sure we want new commands for trivial things like this?

We'd actually need 4 variations 

- buffer-repeat-search (means buffer-repeat-search-same)
- buffer-repeat-search-opposite
- buffer-repeat-search-forward
- buffer-repeat-search-backward

which leads to a (confusing) inflation of available commands.

nilnor:
If we another command, I think that just adding `buffer-repeat-search-opposite` would be sufficient. I'm fine with not adding it as well though, we can hold off on it as well and add if it's every requested.

mfxmfx:
Ok, then please merge this as is - we can always add a command later.

nilnor:
Merged!

PR #285, opened 2016-11-16
VI: vi_spec: enhance the local press function to support simplified test cases.

mfxmfx:


nilnor:
Not sure this is an improvement - it's shorter but less clear. `press` was easier to understand before it accepted magic strings. How about having another helper to cut down on one of the extra lines instead? Alternatively, to make it clearer, use a options table, like so?

```moonscript
press 'i', and_insert: 'my text'
```

mfxmfx:
Well, this is Vi and not [Smalltalk](https://en.wikipedia.org/wiki/Smalltalk) (though I love both). :wink:

Actually I'd like to add lots of Vi tests from [ktexteditor](https://github.com/KDE/ktexteditor/blob/master/autotests/src/vimode/modes.cpp), but I have not looked at the inner workings of bindings.process yet - for some reason ```press 'i', 'a', 'escape'``` didn't work for me.

mfxmfx:
I think that vi-mode needs some love - maybe I'll cook up something over the holidays. Closing for now.

PR #288, opened 2016-11-23
NEED HELP: VI: '/' and '?' should use "match_at_cursor: false".

mfxmfx:
Hi Nils,

I need some help on how to pass "match_at_cursor" down to interactions/search.moon.

Any hints?

Thanks,
Markus


nilnor:
I don't think there's a way of passing in input parameters for commands. Trying to shoehorn something in would not really be worth it either I think, at least not if these are the only cases. I think the easiest way to move forward would be to define custom search commands for VI, e.g. `vi-search-forward` that passes in the flag correctly (have a look at `edit_commands.moon` for the implementations.

PR #289, opened 2016-11-24
VI: '/' and '?' should not match at the current cursor position

mfxmfx:


bundles/vi/base_map.moon
```diff
@@ -118,8 +118,8 @@ map = {
     F: (editor) -> bindings.capture back_to_char
     t: (editor) -> bindings.capture forward_till_char
     T: (editor) -> bindings.capture back_till_char
-    '/': 'buffer-search-forward'
-    '?': 'buffer-search-backward'
+    '/': 'vi-buffer-search-forward'
```
nilnor: Do we want these versions or just `vi-search-forward`, etc.? They end up being slightly long.

mfxmfx:
You are correct about '?'. BTW, I'm having a similar problem while trying to implement '#' (match current word backward), so closing this for now.

PR #293, opened 2016-12-01
Upgrade LPeg to version 1.0.0.

mfxmfx:
LPeg 1.0.0 was released in Sep 2015, and Textadept is also using it since Oct 2015, so it might be a good idea to upgrade.

There are currently 2 test-suite failures which may indicate bugs in the grammars.

Please try it and give it some review.

### DIFF
--- a/lib/howl/util/lpeg_lexer.moon
+++ b/lib/howl/util/lpeg_lexer.moon
@@ -178,7 +178,7 @@ export *
 -- helper patterns
 eol = eol_p
 blank = blank_p
-line_start = -B(1) + B(eol)
+line_start = B'\n' + B'\r\n' + B'\r' + -B(1)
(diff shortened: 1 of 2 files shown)


nilnor:
As you say it's not done. The failures are not only in the test suite but also manifest when running Howl. I did a quick fix for the issues in the Howl lexer stuff (see https://gist.github.com/nilnor/80312a399fb98ca60810d6e660aec60b) but the other issue is for the imported Scintillua lexers, so one would have to go through and fix those as well.

mfxmfx:
Just for reference, here is the related scintillua hg commit from 2014-07-18:

https://foicica.com/hg/scintillua/rev/bfe6008eb013

mfxmfx:
I think this is a high priority issue, but I'm not going to work on this. Closing.

PR #294, opened 2016-12-06
Major config implementation refactor

shalabhc:
While buffer.config, mode.config and the set* commands can still be
used the same way, the underlying implementation has been changed.

The config is now organized along two dimensions - a primary dimension
called scope which is a hierarchial namespace, and a secondary
dimension called layer which can be any string. When defined, a layer
may be associated with another layer as the parent. For each scope,
multiple layers may exist and for each layer within the scope multiple
name/value pairs may exist.

Values are looked up in a scope, and if not found, the parent scopes
are checked, all the way from the immedidate parent to the global
scope, represented by the emtpy string ''. For each scope, value
resolution may check multiple layers.

Proxy config objects on buffer and mode hide most of this machinery by
exposing config values as attributes.

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -10,16 +10,25 @@ ffi = require 'ffi'
 append = table.insert
 min = math.min
 
+buffer_id = 1
+
+_next_id = ->
+  buffer_id += 1
+  return buffer_id
+
 class Buffer extends PropertyObject
   new: (b_mode = {}) =>
     super!
 
+    @id = _next_id!
     @_buffer = aullar.Buffer!
-    @config = config.local_proxy!
     @markers = BufferMarkers @_buffer
     @completers = {}
     @inspectors = {}
     @mode = b_mode
+    @_set_config!
+    @config\clear!
+
     @properties = {}
     @data = {}
     @_eol = '\n'
@@ -54,12 +63,12 @@ class Buffer extends PropertyObject
     set: (new_mode = {}) =>
       old_mode = @_mode
       @_mode = new_mode
-      @config.chain_to new_mode.config
       if new_mode.lexer
         @_buffer.lexer = (text) -> new_mode.lexer text, @
       else
         @_buffer.lexer = nil
 
+      @_set_config!
       signal.emit 'buffer-mode-set', buffer: self, mode: new_mode, :old_mode
 
   @property title:
@@ -120,6 +129,9 @@ class Buffer extends PropertyObject
   @property read_only:
     get: => @_buffer.read_only
     set: (v) => @_buffer.read_only = v
+
+  @property _config_scope:
+    get: => @_file and ('file' .. @_file.path) or ('buffer/' .. @id)
 
   chunk: (start_pos, end_pos) => Chunk self, start_pos, end_pos
 
@@ -284,9 +296,7 @@ class Buffer extends PropertyObject
   config_at: (pos) =>
     mode_at = @mode_at pos
     return @config if mode_at == @mode
-    new_config = config.local_proxy!
-    new_config.chain_to @mode_at(pos).config
-    new_config
+    return config.proxy @_config_scope, mode_at.config_layer
 
   add_view_ref: =>
     @viewers += 1
@@ -295,8 +305,14 @@ class Buffer extends PropertyObject
     @viewers -= 1
 
   _associate_with_file: (file) =>
+    scope = @_config_scope
     @_file = file
+    config.copy scope, @_config_scope
+
     @title = file and file.basename or 'Untitled'
+
+  _set_config: =>
+    @config = config.proxy @_config_scope, 'default', @mode.config_layer
 
   _on_text_inserted: (_, _, args) =>
     @_on_buffer_modification 'text-inserted', args
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -10,7 +10,7 @@ ffi = require 'ffi'
 append = table.insert
 min = math.min
 
-buffer_id = 1
+buffer_id = 0
 
 next_id = ->
   buffer_id += 1
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -12,7 +12,7 @@ min = math.min
 
 buffer_id = 1
 
-_next_id = ->
+next_id = ->
   buffer_id += 1
   return buffer_id
 
@@ -20,7 +20,7 @@ class Buffer extends PropertyObject
   new: (b_mode = {}) =>
     super!
 
-    @id = _next_id!
+    @id = next_id!
     @_buffer = aullar.Buffer!
     @markers = BufferMarkers @_buffer
     @completers = {}
--- a/lib/howl/config.moon
+++ b/lib/howl/config.moon
@@ -175,18 +175,15 @@ proxy = (scope, write_layer='default', read_layer=write_layer) ->
   }
   setmetatable proxy, proxy_mt
 
-
 copy = (scope, new_scope) ->
-  scopes[new_scope] = scopes[scope]
+  scopes[new_scope] = {}
+  if scopes[scope]
+    for k, v in pairs scopes[scope]
+      scopes[new_scope][k] = moon.copy v
 
 delete = (scope) ->
   error 'Cannot delete global scope' if scope == ''
   scopes[scope] = nil
-
-move = (scope, new_scope) ->
-  error 'Cannot move global scope' if scope == ''
-  copy(scope, new_scope)
-  delete(scope)
 
 config = {
   :definitions
--- a/lib/howl/config.moon
+++ b/lib/howl/config.moon
@@ -2,7 +2,8 @@
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 append = table.insert
 
-values = {}
+scopes = {}
+layer_defs = {'default': {}}
 defs = {}
 watchers = {}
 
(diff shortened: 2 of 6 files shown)


lib/howl/config.moon
```diff
@@ -140,25 +162,44 @@ watch = (name, callback) ->
...
+  proxy = {
+    clear: => scopes[@_scope] = {}
+    _scope: scope
+    _write_layer: write_layer
+    _read_layer: read_layer
+  }
   setmetatable proxy, proxy_mt
 
+
+copy = (scope, new_scope) ->
```
nilnor: I'm missing a spec for this (ditto for `delete` and `move`). I'm missing it particularly much for this one as it doesn't seem right - this is not a copy operation but an aliasing operation. I'll comment further where it's actually used in Buffer.

lib/howl/buffer.moon
```diff
@@ -295,9 +305,15 @@ class Buffer extends PropertyObject
     @viewers -= 1
 
   _associate_with_file: (file) =>
+    scope = @_config_scope
     @_file = file
+    config.copy scope, @_config_scope
```
nilnor: Given that `copy` really isn't a copy but an aliasing op, it seems like this would just have the effect of clobbering a scope? E.g. if a new buffer `b` has the config scope "buffer/1234", and is associated with file "/tmp/foo.moon", this would just overwrite any file scope with the buffer scope?

lib/howl/config.moon
```diff
@@ -101,36 +102,57 @@ define = (var = {}) ->
...
+  unless scopes[scope][name]
+    scopes[scope][name] = {}
 
-proxy_set = (name, value, proxy) ->
-  def = get_def name
+  scopes[scope][name][layer] = value
 
-  if def.scope and def.scope == 'global'
-    error 'Attempt to set a local value for global variable "' .. name .. '"', 2
+  broadcast name, value, (scope != '' or layer != 'default')
```
shalabhc: I'm not sure if that would be useful. If anything, `watch` could potentially take a scope and/or layer but I think we should add that only if needed.

lib/howl/config.moon
```diff
@@ -101,36 +102,57 @@ define = (var = {}) ->
   defs[var.name] = var
   broadcast var.name, var.default, false
 
-set = (name, value) ->
+define_layer = (name, def={}) ->
+  layer_defs[name] = moon.copy(def)
+
+set = (name, value, scope='', layer='default') ->
```
shalabhc: Done!

spec/config_spec.moon
```diff
@@ -256,49 +256,121 @@ describe 'config', ->
...
+        local proxy_inner_layered, proxy_outer_layered, proxy_inner_mixed, proxy_inner_sub
+
+        before_each ->
+          config.define_layer 'layer:one'
+          config.define_layer 'layer:sub', parent: 'layer:one'
+          proxy_inner_layered = config.proxy '/outer/inner', 'layer:one'
+          proxy_outer_layered = config.proxy '/outer', 'layer:one'
+          proxy_inner_mixed = config.proxy '/outer/inner', 'default', 'layer:one'
+          proxy_inner_sub = config.proxy '/outer/inner', 'layer:sub'
+
```
shalabhc: Fixed.

lib/howl/buffer.moon
shalabhc: OK, I've fixed copy, removed move (didn't need it) and added specs.

lib/howl/config.moon
shalabhc: Done.

site/source/doc/api/config.md
```diff
@@ -69,9 +131,8 @@ Defines a new config variable. Options can contain the following fields:
 - `description`: A description of the configuration variable (_required_)
 
 - `scope`: An optional value specifying the scope of the variable. One of
```
shalabhc: Do you think this should be changed to specify the valid scopes by pattern? E.g '^$' for global scope and '.+' for non global scope, but also '^file/' for file scopes?

shalabhc:
OK, addressed all comments - please take another look!.

lib/howl/buffer.moon
```diff
@@ -10,16 +10,25 @@ ffi = require 'ffi'
 append = table.insert
 min = math.min
 
+buffer_id = 1
+
+_next_id = ->
```
kirbyfan64: Is there a reason this is preceded by an underscore? I know that's usually done in Python modules, but here, everything's hidden except for `Buffer` itself.

site/source/doc/api/config.md
```diff
@@ -7,46 +7,108 @@ title: howl.config
...
+Values for configuration variables can be specified at multiple levels, called
+*scopes*, and at multiple *layers* within each scope. These are described below.
 
-The value set for the variable is used unless overridden by a mode or buffer
-specific setting (the `set` command always sets variables globally).
+### Scopes
 
-- *Per mode*
+The *scope* is a path within a hierarchical namespace. An example of a scope is
+`'file/home/user/my_dir'` with represents the file path  `/home/user/my_dir`.
```
kirbyfan64: **which**?

kirbyfan64:
I just have to say, I have absolutely no clue what the hell this is doing, but it looks amazing. :D

lib/howl/buffer.moon
shalabhc: Good point - there's no reason, just my Python habit and I'll fix.

shalabhc:
@kirbyfan64 - the api doc update hopefully can help make sense of this.

shalabhc:
OK, all changes are in now, please take another look.

Since with this the commands dont change at all I think it's ok to leave the manual the same as it is now. It describes the exposed model rather than the underlying model. However, once we change the commands (which I want to do), I'll change the manual too to reflect that.

lib/howl/buffer.moon
nilnor: I still think this seems like it's clobbering things, but perhaps I'm squarely in theoretical county here. However, say I have a buffer open with a config and no file, which I then save to `foo.moon`. The current code would just remove any existing configuration for the `file/foo.moon` and replace that with the current buffer config. That doesn't seem right? In the future I might have a particular config setup for this particular file, which would then be destroyed.

lib/howl/config.moon
nilnor: All right, sounds good.

site/source/doc/api/config.md
```diff
@@ -7,46 +7,108 @@ title: howl.config
...
+ 9. scope='', layer='mode:moonscript'
+10. scope='', layer='default'
+```
+
+
+### API
+
+The primitive API consists of [`get()`](#get) and [`set`](#set) calls which
+accept scope and layer as additional parameters. However, the following code
+snippet illustrates the idiomatic ways of setting variables globally, globally
```
nilnor: "globally, globally"

site/source/doc/api/config.md
```diff
@@ -7,46 +7,108 @@ title: howl.config
...
 
 ```lua
 howl.config.my_var = 'foo'
 howl.mode.by_name('ruby').config.my_var = 'foo'
 howl.app:new_buffer().config.my_var = 'foo'
 ```
 
+Note that internally the values are organized within scopes and layers, but this
+convenient API is available on [buffer] and [mode] objects. [Proxy](#proxy)
+objects, described below are used to build the convinience API.
```
nilnor: One instance of "convinience" left here.

lib/howl/buffer.moon
shalabhc: Yes that's what would happen - I'm not saying it's right or wrong, just that it's closer to the preexisting behavior because the proxy object is attached to the buffer object, afaict.

In terms of what's right, if you did set some config for the buffer before `save-as`, you'd want that to be preserved after the `save-as` operation, so a clobber is reasonable IMO. Some kind of merge is also reasonable, but not clearly defined yet.

site/source/doc/api/config.md
shalabhc: Fixed

site/source/doc/api/config.md
shalabhc: Fixed

shalabhc:
@mfxmfx, @nilnor - I *believe* I've addressed all comments now (but with the new github system, I can't be sure). PTAL.

mfxmfx:
I've been using your branch for the last few days - looks good to me.

shalabhc:
@mfxmfx - thanks for testing!

nilnor:
I still think the clobbering that's apparently going isn't right. Do you have any follow-up to my last comment regarding this? We can merge this for now and leave that for later, but unless I'm missing something it's still a real problem.

shalabhc:
I did have a response to the clobbering. Reproducing your commend and my response below. Strangely the same comment appears in two places above, and only one has my response:

----
nilnor 15 days ago
I still think this seems like it's clobbering things, but perhaps I'm squarely in theoretical county here. However, say I have a buffer open with a config and no file, which I then save to foo.moon. The current code would just remove any existing configuration for the file/foo.moon and replace that with the current buffer config. That doesn't seem right? In the future I might have a particular config setup for this particular file, which would then be destroyed.

shalabhc 15 days ago
Yes that's what would happen - I'm not saying it's right or wrong, just that it's closer to the preexisting behavior because the proxy object is attached to the buffer object, afaict.

In terms of what's right, if you did set some config for the buffer before save-as, you'd want that to be preserved after the save-as operation, so a clobber is reasonable IMO. Some kind of merge is also reasonable, but not clearly defined yet.
---- (end copy)

nilnor:
Ah, thanks for collecting it, the last instance of my comment was and is indeed shown without your reply.  I'm afraid I don't see how the new behaviour mimics the old one though. Previously, with buffer/mode/global there wasn't the notion of file scoped configuration, but now there is. I guess this won't happen much in reality, but it's still very much wrong as I see it. If I have specified a certain configuration for a certain file, I wouldn't expect the entire configuration to be clobbered. It sounds reasonable to clobber any individual configuration variables that have been set for the buffer, but I would expect any other file specific setting to remain.

Anyway, I'll leave the decision of whether to address this in your hands, merge at will!

shalabhc:
I see buffer scoped configs effectively the same as file scoped configs. Here is why I think the new behavior mimics the existing one:

Say you have a file `x.moon` with config values `a=1` and `b=2`. Then you create a new buffer, set `a=3` and no value for `b`. Then you save-as to `x.moon`. So you end up with `x.moon` that only has `a=3` and no value for `b`, effectively clobbering the old config for the file.

nilnor:
It's a matter of perspective I guess, but for me buffer scoped configs aren't the same as file scoped configs. I'm just thinking ahead to when we have directory/project specific file configurations. If I create a new buffer in a project and then save it under a specific name, which is covered by a project specific file configuration setting, I would expect that to be applied. The clobbering breaks this assumption. I see no use case in the current model where my assumptions would be similarly violated. For your example above, I could in the future (as the support arrives) have a file specific configuration specifically for `x.moon`. No such thing exists today, and there's no support for it in the current model. But again, I leave the decision of whether this needs to be addressed now, or at all, to you.

shalabhc:
@nilnor - I see what you mean. I agree it's reasonable to have persistent file configuration that is not clobbered by `save-as`. There may also be a case where it's confusing though. E.g. if I set a config value for a buffer, then `save-as` and my config value now automatically becomes the persistent config value for the file path. So maybe there is the need to explicitly specify values as persistent - and save them in a separate layer? Or maybe it doesn't matter because this is a corner case anyway. In any case, I think we can address this as the features around config evolve further.

kirbyfan64:
So, I think this changed some behavior that I'm curious about:

I noticed that, since I recently upgraded Howl, some of the code in my config file no longer behaves the same way. Here's a snippet:

```moonscript
config.indent = 4

signal.connect 'file-opened', (args) ->
  -- ...
  if args.file\is_below File"/home/ryan/stuff/fluidsynth-g"
    args.buffer.config.indent = 2
```

Now, the indent still stays at 4.

shalabhc:
@kirbyfan64 - is this just a redrawing issue (i.e. indent guides stay at 4 but `tab` works as expected) or config issue (i.e. `tab` doesn't work as expected either)?

kirbyfan64:
Config issue.

Also COME ON GITHUB MOBILE. This is the fourth time in the last week I've hit the wrong button while trying to comment... >:(

ozra:
A parenthesis: - I haven't issued it yet because I wanted to reproduce first, however:
- I've seen indent-guides at 2 spc in buffers - even though I've got 3 spc set at all conf-levels (except for some specific modes - not this one).
- Tab worked as expected.
- Setting the conf to the same value from cmd-line corrected the guides immediately.
- I haven't had 2 spc set for that buffer or any conf level since the start of the editor - which makes it quite peculiar

shalabhc:
@kirbyfan64 - with multiple attempts, I am not able to reproduce this on master. Can you add some logging to ensure your hook is getting executed?

@ozra yes If you can produce the steps for reproduction, please file a bug for the redraw issue.

ozra:
@shalabhc - I'll try and isolate, it seems to be a more general conf-slip - at least it happens with line-wrap settings for me also.

PR #298, opened 2016-12-24
Bind select-all to ctrl_a instead of ctrl_shift_a in default keymap.

mfxmfx:


shalabhc:
+1 because this is the more common shortcut for select all, IME.

But can we also change the mac shortcut to meta_a (see https://support.apple.com/en-us/HT201236)

nilnor:
Let's get the OSX change in as well and then we can merge this.

mfxmfx:
Updated for OSX and rebased on current master.

nilnor:
:+1:

PR #299, opened 2016-12-30
Bind window-toggle-fullscreen to F11 (instead of Alt-F11) and Alt-Enter.

mfxmfx:


mfxmfx:
Minor cleanup and rebased on current master.

kirbyfan64:
Why alt+enter, too? Seems like an obscure fullscreen key combo...

mfxmfx:
Alt-Enter is a common shortcut for us "old timers" (it orginates from command.com and cmd.exe terminals under Windows) - please see the git commit message for more info.

nilnor:
I recognize Alt-Enter as well from way back, but that was a long time ago now and as per the comment I'd say it's not widely recognized today. So, I think just having F11 by default is better - those wanting Alt-Enter can easily bind it themselves.

mfxmfx:
Updated and rebased to current master.

nilnor:
:clinking_glasses:

PR #302, opened 2016-12-30
On startup, emit 'file-opened' a little bit later so that there is an…

mfxmfx:
… editor for the initial buffer.

This is important for an upcoming "improved statusbar" feature.


shalabhc:
Ha - this issue just hit me too and I had to add a check in my file-opened hook. So :+1:  on this change.

But why change the signature (i.e pass the file name instead of file object?). You can always get the name from the object easily in the hook.

mfxmfx:
You are right - this was a misunderstanding from my side.

Fixed and rebased to current master.

nilnor:
:+1:

Issue #147, opened 2015-12-17
Howl can clear out files!


kirbyfan64:
This is quite possibly the most ANNOYING bug EVER!!!!!

Howl can occasionally freeze. Which is fine. However, if you were trying to save when Howl froze, and then you force-close it, _the file will be completely empty upon reopening_. I just lost several days of work because of this.

It's hard to reproduce this, but it's still really bad.

nilnor:
It sounds like there's some special stuff going on here. For one, I very seldom happen upon freezes, and when I do it's because of some kind of code that I'm writing that misbehaves. Does your freezes typically occur when you save? Furthermore, there's nothing special going one when it comes to saving, it's a plain open and write and close. The whole procedure is blocking though. It's sounds like you're saving to a misbehaving file system, are you using a remote mount?

Anyway, it makes sense to improve this generally. I'm thinking we could save to a temporary file in the same directory and then do an atomic rename instead. Also, it would be nice for the remote scenario to switch to a non-blocking save and implement a time-out.

kirbyfan64:
@nilnor Well:

1. Howl will just occasionally *freeze* for me. It only happens once in a while, and it always lasts a second or so. I've never been able to accurately reproduce it; it seems totally random. The problem is when I hit save and then hit close, but in froze in the middle. Also, no remote mount.

2. Yeah, I've seen the temporary file strategy used elsewhere.

shalabhc:
Howl had frozen on me, but it's rare and it hasn't happened recently. I also use a lot of remote folders so I assumed that has something to do with it.

Regarding renaming, one thing to consider is any other apps that have the same file open will not see the changes if a rename is used. Other apps will continue to use the old inode unless they are specifically designed to track renames. I also just verified that both emacs and vi use the same inode.

If the freeze happens after open but before a write call, and then the process exits (of the file is closed) I believe the file will be truncated. It's a shame there's no OS functionality to atomically replace contents. Hmm, maybe we can implement a C function that just does an `open` operation followed by a `write` operation (and maybe also a `flush` or `close`). Perhaps it is much likely to freeze or crash because control does not re-enter the Lua runtime between these calls?

shalabhc:
Other issues with rename might be that hard links may break and symbolic links may get replaced with a file.

nilnor:
[...]

PR #303, opened 2017-01-03
Check the file size to detect if a file is modified on disk.

mfxmfx:
Some tools like to play tricks with the time stamp of files (archivers, and
recently a number of tools that handle "reproducible builds"), so for extra
safety also check the file size.


mfxmfx:
Obsolete - in the meantime if rewritten loading and saving to be safe in case of IO and other erros, so closing this.

kirbyfan64:
> in the meantime if rewritten loading and saving to be safe in case of IO and other erros

#147

PR #304, opened 2017-01-11
Always log non-info to the terminal

kirbyfan64:
Previously, if an error occurred, Howl would always try to show it on the window. The problem is that that  makes it very hard to open bug reports, since you have to try and re-type the error message word-for-word, and it can also make it difficult when one error overshadows another.

### DIFF
--- a/lib/howl/log.moon
+++ b/lib/howl/log.moon
@@ -29,13 +29,18 @@ dispatch = (level, message) ->
   entry = :message, :level
   append entries, entry
   window = _G.howl.app.window
+  showed_message = false
+
   if window and window.visible
     status = window.status
     command_line = window.command_line
     essentials = essentials_of message
     status[level] status, essentials
     command_line\notify essentials, level if command_line.showing
-  elseif not _G.howl.app.args.spec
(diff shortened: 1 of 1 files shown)


nilnor:
FWIW I like it when the console is quiet. We do currently store logged errors, etc., and the idea was that at some point we'll add a message view / log view where one can view all the latest entries. I'm open to doing this in the meantime, but it would be interesting to hear if others has any opinions? If we don't hear anything we can proceed with this.

shalabhc:
I'm fine with this - I usually leave the console dedicated to Howl, so messages showing up are fine.

kirbyfan64:
As a side note: I just realized this also ends up printing when it fails to find a string in a file...

Maybe there could also be a `critical` mode, dedicated to errors within Howl itself? Then only that would get printed out.

nilnor:
@kirbyfan64 I'm open to adding a new `critical` level.

nilnor:
@kirbyfan64 do you want to add the new level first before we merge this?

kirbyfan64:
Superseded by #311.

PR #305, opened 2017-01-11
Add temporary file backups during saving

### LINKED ISSUE #147: Howl can clear out files!

kirbyfan64:
Closes #147. If anything goes super badly, then the backup will never be able to get deleted, so the file stays intact.

I saved the backups into `~/.howl/backups` instead of the cwd to avoid them potentially being an annoyance if they show up in version control, and the PID is used just in case two different instances of Howl are saving two different files with the same name at the same time.

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -205,8 +205,8 @@ class Buffer extends PropertyObject
       local backup
       if @config.backup_files and @file.exists
         file_stem = "#{@file.basename}::#{ffi.C.getpid!}::#{@file.etag}"
-        backup_directory = @config.backup_directory or howl.app.backupdir
-        backup = File(@config.backup_directory) / file_stem
+        backup_directory = File @config.backup_directory or howl.app.settings.backupdir
+        backup = backup_directory / file_stem
         status, err = pcall @file\copy, backup, {'COPY_OVERWRITE', 'COPY_ALL_METADATA'}
         if not status
           log.error "Failed to write backup file #{backup} for #{@file}: #{err}"
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -202,7 +202,7 @@ class Buffer extends PropertyObject
         @append @eol
 
       local backup
-      if config.backup_files and @file.exists
+      if @config.backup_files and @file.exists
         backup = howl.app.settings.backupdir / ("#{@file.basename}::#{ffi.C.getpid!}::#{@file.etag}")
         status, err = pcall @file\copy, backup, {'COPY_OVERWRITE', 'COPY_ALL_METADATA'}
         if not status
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -204,7 +204,7 @@ class Buffer extends PropertyObject
       local backup
       if config.backup_files and @file.exists
         backup = howl.app.settings.backupdir / ("#{@file.basename}::#{ffi.C.getpid!}::#{@file.etag}")
-        status, err = pcall @file\copy, backup, {'OVERWRITE', 'ALL_METADATA'}
+        status, err = pcall @file\copy, backup, {'COPY_OVERWRITE', 'COPY_ALL_METADATA'}
         if not status
           log.error "Failed to write backup file #{backup} for #{@file}: #{err}"
           backup = nil
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -201,10 +201,24 @@ class Buffer extends PropertyObject
       if @config.ensure_newline_at_eof and not @text\match "#{@eol}$"
         @append @eol
 
+      local backup
+      if config.backup_files
+        backup = howl.app.settings.backupdir / ("#{@file.basename}::#{ffi.C.getpid!}")
+        status, err = pcall -> backup.contents = @file.contents
+        if not status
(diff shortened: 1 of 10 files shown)


nilnor:
For something that is intended to be done fairly often (every save), we should make sure that it's as efficient as possible. This causes full round trips (from disk to a Lua string and back) to be done on each save. Instead, we should expose a `File#copy` operation that does this more efficiently (there's a `g_file_copy` that can be used conveniently enough).

kirbyfan64:
@nilnor Done!

lib/howl/buffer.moon
```diff
@@ -357,6 +371,12 @@ with config
     default: true
     type_of: 'boolean'
 
+  .define
+    name: 'backup_files'
+    description: 'Whether or not to make temporary backups of files while saving'
+    default: true
```
shalabhc: I think the default should be `false`, mainly due to privacy concerns. Copying every file saved into another directory shouldn't happen automatically. Consider I may be viewing/editing a sensitive file in a network folder or usb drive - that shouldn't be copied into backupdir.

kirbyfan64: @shalabhc Done.

spec/buffer_spec.moon
```diff
@@ -3,6 +3,9 @@ import File from howl.io
 import with_tmpfile from File
 
 describe 'Buffer', ->
+  before_each ->
```
shalabhc: This is now not needed since default is `false`?

kirbyfan64: @shalabhc Fixed.

lib/howl/io/file.moon
```diff
@@ -201,6 +202,17 @@ class File extends PropertyObject
 
     files, false
 
+  copy: (dest, flags, progress) =>
+    bitflags = 0
+    flags or= {}
+
+    for flag in *flags
```
nilnor: There's actually a helper down in ljglibs that can be used for doing this instead, I'll give some more details on this for ljglibs `File` below. We don't have any use of `progress` right now, so I say we should leave that out for now.

lib/ljglibs/gio/file.moon
```diff
@@ -57,6 +66,17 @@ core.define 'GFile', {
   enumerate_children: (attributes, flags = @QUERY_INFO_NONE) =>
     gc_ptr catch_error C.g_file_enumerate_children, @, attributes, flags, nil
 
+  copy: (dest, flags, cancellable, callback) =>
```
nilnor: Here, for the for parsing the flags I'd like to see [core.parse_flags](https://github.com/howl-editor/howl/blob/master/lib/ljglibs/core.moon#L163) being used. We should actually update `query_info` above to use the same thing, I think this file predates this helper. With this in place we could remove the flags handling from howl's `File`. The code would look like this:

```moonscript
flags = core.parse_flags('G_FILE_', opts.flags)
```

`callback` is too generic a name, `progress_callback` provides an idea of what it is.

lib/ljglibs/gio/file.moon
```diff
@@ -57,6 +66,17 @@ core.define 'GFile', {
...
+    self = @
+    local cb_handle
+
+    handler = (current_bytes, total_bytes) ->
+      callback self, current_bytes, total_bytes
+
+    cb_handle = callbacks.register handler, 'file-progress'
```
nilnor: There's no check for whether "handler" actually is present. In this PR, it actually isn't so there's an ignored error raised for every file save.

lib/ljglibs/gio/file.moon
```diff
@@ -57,6 +66,22 @@ core.define 'GFile', {
   enumerate_children: (attributes, flags = @QUERY_INFO_NONE) =>
     gc_ptr catch_error C.g_file_enumerate_children, @, attributes, flags, nil
 
+  copy: (dest, flags, cancellable, progress_callback) =>
+    self = @
+    local handler, cb_handle, cb_cast, cb_data
+
+    copy_flags = core.parse_flags 'G_FILE_COPY_', flags
```
nilnor: For consistency with flags elsewhere, use 'G_FILE_' as the prefix here.

kirbyfan64:
@nilnor Done.

lib/howl/buffer.moon
```diff
@@ -201,10 +201,24 @@ class Buffer extends PropertyObject
       if @config.ensure_newline_at_eof and not @text\match "#{@eol}$"
         @append @eol
 
+      local backup
+      if config.backup_files and @file.exists
```
nilnor: The option is not defined as being global, but here we only check the top level config? Either it should be global or `@config` should be used. It would be great with a spec exercising this new behaviour. It should be possible for instance to use a file in a temp directory and removed the directory before saving, and then verify that the backup file is present.

kirbyfan64:
@nilnor Done.

kirbyfan64:
*lets try Travis again...*

nilnor:
I merged this manually with a few smaller tweaks, thanks! Can you also update the Changelog to mention the newly added option?

nilnor:
Oh, and I also squashed all of your commits to get a cleaner history, hope you don't mind.

PR #306, opened 2017-01-18
Configuration auto persistence and more

### LINKED ISSUE #13: Config variables set via readline should persist
Feature Request:

When I use the set-\* commands to set a config variable, I would like the setting to be saved permanently, and prefer not to have to edit init.moon for that. Transient settings are seldom required and more often users want to save the new settings.

Allowing manual editing of config variables via ~/howl/system/config.lua is reasonable.

shalabhc:
Adds automatic configuration persistence. Fixes #13 

Config value set via command line or API can now be automatically persisted on clean exit. To enable this `config.save_config_on_exit` must be set to true (default is false). In addition, the `save-config` command can be used at any time to save the current config. The saved config is loaded automatically on the next startup.

Also includes a bugfix and replaces the `config.copy` method with `config.replace` and `config.merge` to make it more explicit.

### DIFF
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -252,6 +252,9 @@ class Application extends PropertyObject
     if force or not @_should_abort_quit!
       @save_session! unless #@args > 1
 
+      unless pcall config.save_config
+        print 'Error saving config'
+
       for _, process in pairs Process.running
         process\send_signal 'KILL'
 
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -252,8 +252,9 @@ class Application extends PropertyObject
     if force or not @_should_abort_quit!
       @save_session! unless #@args > 1
 
-      unless pcall config.save_config
-        print 'Error saving config'
+      if config.save_config_on_exit
+        unless pcall config.save_config
+          print 'Error saving config'
 
       for _, process in pairs Process.running
         process\send_signal 'KILL'
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -250,11 +250,12 @@ class Application extends PropertyObject
 
   quit: (force = false) =>
     if force or not @_should_abort_quit!
-      @save_session! unless #@args > 1
+      unless #@args > 1
+        @save_session!
 
-      if config.save_config_on_exit
-        unless pcall config.save_config
-          print 'Error saving config'
+        if config.save_config_on_exit
+          unless pcall config.save_config
+            print 'Error saving config'
 
       for _, process in pairs Process.running
         process\send_signal 'KILL'
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -308,7 +308,8 @@ class Buffer extends PropertyObject
     scope = @_config_scope
     @_file = file
     config.merge scope, @_config_scope
-
+    config.delete scope
+    @_set_config!
     @title = file and file.basename or 'Untitled'
 
   _set_config: =>
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -307,7 +307,7 @@ class Buffer extends PropertyObject
   _associate_with_file: (file) =>
     scope = @_config_scope
     @_file = file
-    config.copy scope, @_config_scope
+    config.merge scope, @_config_scope
 
     @title = file and file.basename or 'Untitled'
 
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -417,6 +417,13 @@ command.register
   input: (path=nil) -> interact.get_external_command :path
   handler: launch_cmd
 
+command.register
+  name: 'save-config'
+  description: 'Save the current configuration'
+  handler: ->
+    config.save_config!
+    log.info 'Configuration saved'
+
 config.define
   name: 'project_build_command'
   description: 'The command to execute when project-build is run'
--- a/lib/howl/config.moon
+++ b/lib/howl/config.moon
@@ -8,7 +8,7 @@ layer_defs = {'default': {}}
 defs = {}
 watchers = {}
 
-unsaved_scopes = howl.Regex '^buffer/'
+saved_scopes = {''}
 
 predefined_types =
   boolean: {
@@ -134,8 +134,9 @@ save_config = (dir=nil) ->
   return unless scopes
   settings = Settings dir
   scopes_copy = {}
-  for scope, values in pairs scopes
-    continue if unsaved_scopes\match scope
+  for scope, values in *saved_scopes
+    values = scopes[scope]
+    continue unless values
     persisted_values = nil
     if get 'persist_config', scope
       persisted_values = values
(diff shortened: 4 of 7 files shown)


kirbyfan64:
So, I know this seems kind of silly, but could a `set-temp` command also be added? Sometimes it's nice to play around with settings or set them for specific files (e.g. `set indent 2` for MoonScript and HTML) *without* permanently applying them.

mfxmfx:
Automatically transforming command line options into "persistent state" is just a bad idea...

If I want to change my settings I change init.moon - especially that we have such a great scripting language.

It would make more sense to support a "project" configuration system & layer.

shalabhc:
I think using the command line first and then having to do something else to persist your changes should be unnecessary. In most apps, once you apply your settings, they are saved across restarts. In Howl too there is state that persists across restarts - your open files for instance.

However I see there is some resistance to this pattern when applied to the config. So maybe we can have a global setting `config.auto_save_config` - if you turn it off it will not auto save on exit. Of course this will have to be in your init.moon, because it will not auto save :wink:

shalabhc:
OK, I've now turned off auto save by default. Instead I added `config.save_config_on_exit` which can be set to true to get back auto saving behavior. In addition I added a `save-config` command that can be explicitly run to save the current config. Does this address your concerns, @mfxmfx and @kirbyfan64?

shalabhc:
No comment implies no objection?

@nilnor I think this PR is safer now because there's no save-by-default, but an easy switch to enable auto saving - what do you think?

nilnor:
I think this is reasonable now when not enabled by default. Two points though:

- I believe we've talked previously about unifying the various set commands, so that issuing `set foo=bar` would prompt for the scope at which to set the value. This becomes more important when auto-saving is enabled I think. It's very easy otherwise to just open up a file, issue something like `set indent=4` for that particular file and then have it be persisted globally although that wasn't intended. Perhaps we shouldn't auto save at all if opened with arguments?

- The current saving logic is too simplistic. It now dumps everything, which isn't desirable. For instance, it dumps the buffer specific config scopes (e.g. `buffer/12`) which obviously will give weird results when loaded. It also dumps all file specific settings, and for that I'd also much rather see a project/directory specific file option. Even though we don't have one yet, saving files scopes now would clash with that.

shalabhc:
Ah, saving 'buffer/*' is an oversight and I'll fix it. Not saving when opened with arguments is a good idea. I believe the session is not saved in that case as well, and the config should definitely follow that.

shalabhc:
OK, did a bunch of cleanup:

* Empty config scopes are not saved now - so extraneous 'file/..' values should not exist in config.lua anymore
* Reused check for save_session to also protect config saving - so config is not saved for one-off invocations (e.g. `howl path/to/file`)
* Added `config.set_default` - this is to install *unsaved* default values. It is needed for mode values set via the bundle defaults. If mode values are set the usual way, they end up getting persisted, but they should not as they are the default configured by the bundle. This also *might* help with @kirbyfan64's request for a `set-temp` type command.

shalabhc:
@nilnor - so I think this is good for another look. I agree that all this becomes more useful once the `set` commands are updated too. In the latest state, this only saves a small set of explicitly set configs, so I think this is good to check in. I am working on the new `set` command as well, but no real ETA yet.

nilnor:
I still think that saving file configuration will clash when we add the project config, what do you think about that?

shalabhc:
The file config supersedes the directory/project config, but not sure how it is a clash?

nilnor:
I'm just wondering how we'll handle having both. When we add the project config, I'd expect that to be the ultimate config source for files under the project, with any configuration set for the file ending up there. This will work if we just stop loading the file scopes later, but is that what you had in mind?

shalabhc:
Hmm, I see what you mean. I don't expect the semantics of file scopes to change, but just their persistence location. So instead of storing in one global file, some file scopes are stored in a project level config. I think this transition is possible transparently, but I'll think about it some more.

shalabhc:
So after some thinking, I haven't figured out how to best store the file configs in each project. I think this will have to be something along these lines: the 'file/' scope handles persistence by calling a registered handler outside config itself - the handler would find the closest project and persist the configs there, but using relative path only.

In any case, in lieu of the details of the complete solution for 'file/' scopes, I'm going to just remove persistence of file configs for now - so only the global config will be persisted.

shalabhc:
PTAL - global scopes are persisted, this includes mode configs as well.

ozra:
Just saw this, so I'll reference a comment here: https://github.com/howl-editor/howl/issues/317#issuecomment-284408356

I'll re-iterate to keep discussion here:

I consider the moon/lua based config as authority, and I think adding multiple layers of configuration abstraction is bad (TM).

When a user first uses the editor the default-configuration should be copied to user-conf-dir as "init.moon", it also references an initially empty file '".auto-confs.moon"'.

_"cmd-line-/ux-issued conf mods"_ should then be (_if wanted, as above discussion reveal might not be the case of course_) persisted into the '".auto-confs.moon"' as code (through AST-manipulation* or simply source based pattern heuristics). Preferably with a comment with time-stamp and in which file one was when changing the conf (this helps with reviewing what the f**k was going through one's mind when making that setting; not all changes are done in the best state of mind, deadlines on the horizon and shit).

The auto-confs should be write protected to encourage user to think one extra time before (inevitably ;-) ) doing edits directly in it (which is not an "officially intended" thing to do), so that it maintains the structure that can be mutated by the persister-code.

I agree that `set` should target buffer-setting primarily _or_ all `set`-variants should have fully qualified names, none given "precedence": `set-global`, `set-module`, `set-buffer` and _preferably have specific short aliases for (all) commands: 'sg', 'sm', 'sb' - at least I would love that (once again, I'm shaped by command line and vim featureisms (`foo --some-flag` vs `foo -s` and `set softabstops=3` vs `set sts=3`) [I don't particularly love Vim, I've just used it for 20+y side by side with my ui-editors, and it does have specific features that are very good]. I mean that those should be _defined shot names_ not using generic expansion...

shalabhc:
The problem with having one default configuration file dump (as I think @kirbyfan64 mentioned elsewhere) is that it would be not only too big, but may also be incomplete because each bundle may have it's own additional defaults.

Anyway, I think what you propose with auto-conf.moon being auto persisted is similar to what is in this PR - once you set a flag, all config changes are auto persisted into .howl/system/config.moon. It is editable of course, but you'd have to go looking for it first, and hard to edit accidentally.

ozra:
Ah, yes, subsequently added bundles makes a default init kind of lacking.

Feel a bit stupid for the auto-conf suggestion since that that's what this is *face-palm*, haha. Looking forward to have this in the editor :-)

nilnor:
All right, so let's merge this then! A few things more (aren't there always?): A quick note in the Changelog (we've been lazy there but let's start it up again), and I think this would do well to be noted in the documentation chapter about configuration. I'll leave the merge and the decision on when to add these other to you @shalabhc.

shalabhc:
Woohoo! Merging. Will follow up shortly with docs and Changelog update.

PR #311, opened 2017-02-09
Add log.critical and howl.util.safecall (take 2)

kirbyfan64:
Because...Git. Supersedes #304, replaces the failed #310 (apparently I don't know how to push...).

TODO: Add safecall to docs, figure out why the hell the specs are being weird (I'm getting freezes, bizarre test failures, etc.).

### DIFF
--- a/lib/howl/aux.moon
+++ b/lib/howl/aux.moon
@@ -4,6 +4,6 @@ setmetatable {}, {
   __index: (key) =>
     tb = debug.traceback '', 2
     first = tb\match 'stack traceback:\n%s*([^\n]+)'
-    log.error "`howl.aux` is deprecated, please update your code to use `howl.util` instead\n  (#{first})\n"
+    log.critical "`howl.aux` is deprecated, please update your code to use `howl.util` instead\n  (#{first})\n"
     util[key]
 }
--- a/lib/howl/bindings.moon
+++ b/lib/howl/bindings.moon
@@ -205,9 +205,9 @@ export dispatch = (event, source, maps_to_search, ...) ->
     elseif htype == 'table'
       push handler, pop: true
     else
-      _G.log.error "Illegal handler: type #{htype}"
+      _G.log.critical "Illegal handler: type #{htype}"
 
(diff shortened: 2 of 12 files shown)


nilnor:
Adding the `critical` level is fine, but I'm not a fan of the `safecall`. We're actually already doing special stuff for the pcalls, if you look in the `moonscript_support` file, which this new xpcall doesn't. At any rate I rather have the calling code decide whether to log or not explicitly I think.

ozra:
For certain scenarios code should definitely be "safe-run":

- "evalute moonscript/lua" command
- ~~if~~ _when_ a console is added - anything executed there
- Otherwise, I'd prefer the idea of being able to "turn it on / off" _per bundle_, for unstable bundles one might love, or for bundles one is hacking actively on.
- and then also an option whether to "ignore error" or "open the console" (much like in web-browsers, to cater to different "kinds" of users needs) - perhaps also per bundle.

nilnor:
@ozra everything is already "safe-run" in a way already, so this PR changes nothing in that regard. When we get a console/error view later on we can do more interesting stuff with that, such as the 'open in console' behaviour you mention. For this PR though, just adding the critical level would be enough.

kirbyfan64:
@nilnor The idea for `safecall` basically revolves around printing the stack traces, that way it's evident where the actual error is. Maybe I'm alone here, but sometimes I get irritated when I get something about a nil value in a helper function that's used in 12 different places...

Would you be more open to the idea if it worked with the modified `pcall`?

nilnor:
The existing error rewriting already tries to get the stack traces. It's currently geared towards rewriting moonscript , so perhaps there are some cases where it fails to get the stack trace. However, the `log` also shows just the essentials of the message, so it's more likely that that is the reason why you're not seeing anything. In short, I still don't see the need for `safecall`, instead have a look at what's there and see how that can be improved instead. Just adding a message view for viewing the full traces might be the best way forward.

kirbyfan64:
This has *finally* been superseded by #441.

PR #323, opened 2017-03-08
Add custom flair type.

ozra:
I couldn't get the cursor to look like I want in a theme, so I added this simple solution.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,12 @@
 
 ## Unreleased (in master)
 
+- Added `custom_draw` flair type (`highlight.CUSTOM`).
+
+- Modified Makefile to: only recompile bytecode when needed, optimize machine
+code for local architecture (Linux only in lack of test machines), forcefully
+copy `howl` binary on make install, to not have to close instances to rebuild.
+
 - Added command line help which is invoked by pressing `f1` while any
 interactive command is running. This displays a popup containing information
 about the command.
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,10 +4,6 @@
 
 - Added `custom_draw` flair type (`highlight.CUSTOM`).
 
-- Modified Makefile to: only recompile bytecode when needed, optimize machine
-code for local architecture (Linux only in lack of test machines), forcefully
-copy `howl` binary on make install, to not have to close instances to rebuild.
-
 - Added command line help which is invoked by pressing `f1` while any
 interactive command is running. This displays a popup containing information
 about the command.
--- a/lib/aullar/flair.moon
+++ b/lib/aullar/flair.moon
@@ -35,8 +35,21 @@ set_line_type_from_flair = (cr, flair) ->
       cr.dash = {6, 3}
 
 draw_ops = {
-  rectangle: (flair, x, y, width, height, cr) ->
+  custom: (flair, x, y, width, height, cr) ->
+    if flair.custom_draw and
+         pcall flair.custom_draw, flair, x, y, width, height, cr
+      return
+    else
+      -- Draw an intensely red "shit has gone bad"-cross instead
+      cr\set_source_rgba 1, 0, 0, 1
+      cr.line_width = 2
+      cr\move_to x, y
+      cr\line_to x + width, y + height
+      cr\move_to x, y + height
+      cr\line_to x + width, y
+      cr\stroke!
 
+  rectangle: (flair, x, y, width, height, cr) ->
     if flair.background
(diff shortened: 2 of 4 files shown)


kirbyfan64:
> Makefile: speed up rebuild.

#124

Also, I think your Makefile changes broke clean builds? https://travis-ci.org/howl-editor/howl/builds/208806948

ozra:
Ah, I removed the dependency on howl it self, which is needed to compile, thanks for catching that!

ozra:
Hmm, shouldn't it rebuild again? Well, there should be _one_ thing in a PR - my bad to begin with. I'll revert the Makefile in this PR.

lib/aullar/flair.moon
```diff
@@ -35,8 +35,21 @@ set_line_type_from_flair = (cr, flair) ->
       cr.dash = {6, 3}
 
 draw_ops = {
-  rectangle: (flair, x, y, width, height, cr) ->
+  custom: (flair, x, y, width, height, cr) ->
+    if flair.custom_draw and
+         pcall flair.custom_draw, flair, x, y, width, height, cr
+      return
+    else
+      -- Draw an intensely red "shit has gone bad"-cross instead
```
kirbyfan64: Is the wording here really that necessary? :O (Just use *crud* or something...)

shalabhc: Or just do nothing?

Note that in another failure case - if custom_draw fails - nothing is drawn. (The pcall return value is ignored above), I think in either case logging an error and doing nothing is reasonable.

ozra: Ah, sorry, I've been reminded on many occasions my "private language" is less than appealing to many. I'll change it to more factual wording.

ozra: @shalabhc - In the case of the flair being for cursor, it is nice that you still see where you're editing. Also I figure there's a much greater chance it's not skimped over, resulting in an editor with tons of "invisibly failing features" eating cycles (see that happening in many plugins in many editors else where). It seems human mentality is "see no evil - there is no evil", haha.

@shalabhc  Perhaps the code formatting was badly chosen: there's: `and` on line 39...

Changelog.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased (in master)
 
+- Added `custom_draw` flair type (`highlight.CUSTOM`).
+
+- Modified Makefile to: only recompile bytecode when needed, optimize machine
```
nilnor: Unrelated to this PR.

lib/aullar/flair.moon
```diff
@@ -35,8 +35,22 @@ set_line_type_from_flair = (cr, flair) ->
       cr.dash = {6, 3}
 
 draw_ops = {
-  rectangle: (flair, x, y, width, height, cr) ->
+  custom: (flair, x, y, width, height, cr) ->
+    custom_fn = flair.custom_draw
+    if custom_fn and pcall custom_fn, flair, x, y, width, height, cr
```
nilnor: You're throwing away the actual error message here which would make it hard to understand why it's failing. For a failed invocation the error message should be visible somehow - either by printing it to stderr or by failing the draw completely and let the default handling take care of it.

ozra: Should be all good now.
I think it's worth it to catch the error here, otherwise the whole editor draw is wiped if it's an early drawn flair.
I imagine a custom-draw is something a theme-maker "improvises forward", so there's likely a few errors in the progress.

lib/aullar/flair.moon
```diff
@@ -35,8 +35,27 @@ set_line_type_from_flair = (cr, flair) ->
...
 
+    if flair.custom_draw
+      ok, err = pcall flair.custom_draw, flair, x, y, width, height, cr
+    else
+      ok ,err = false, "'custom_draw' function missing"
+
+    if not ok
+      -- Something went wrong in custom_draw, so draw an intensely red "missing"
+      -- cross, for visual attention and to maintain functioning of the editor.
+      log.error err
```
nilnor: We'll have to content ourselves with `io.stderr:write` here, as log is only a global on the Howl level, which is above aullar in the dependency chain.

ozra: Alright, still acquainting myself with the code landscape. It's in.

nilnor:
Thanks! I merged this locally and squashed everything in one commit in order to avoid clutter - I hope that's OK with you.

ozra:
All good!

PR #332, opened 2017-03-10
Change default for line-padding to 0

### LINKED ISSUE #326: mis-feature: line-padding default value - should be 0!
Changing the way fonts are rendered from _their_ default - by default - is generally _not desirable_ to a user.
It took me a while to discover the setting and understand why my preferred fonts made me slightly dizzy even though I changed and changed between them.

Personally the only deviation in this regard _I've ever used_ is "{padding-top: **-1**, padding-bottom: **-1**}" for _specific fonts_, in apps which supports it.

The setting should naturally be 0 by default!

ozra:
If there are no objections. Fix #326.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,8 @@
 ## Unreleased (in master)
 
 - **breaking** - Default for `line-padding` setting has been changed to `0`. If
-you've relied on it: set it to its' previous value `1` in your "init.moon".
+you've relied on it: set it explicitly to its' previous value `1` in your Howl
+configuration.
 
 - Added `config.save_config_on_exit` variable to automatically save global
 configuration to `~/.howl/system/config.lua`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ## Unreleased (in master)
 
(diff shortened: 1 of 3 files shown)


nilnor:
There are actually two different default values set - one in aullar and one at the howl level. As it is this PR doesn't actually change the default at all. Have a look in `editor.moon` for the relevant config definition.

ozra:
Alright, way too quick PR that one. I'll fix it proper.

ozra:
Hopefully as it should now :)

Changelog.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased (in master)
 
+- **breaking** - Default for `line-padding` setting has been changed to `0`. If
+you've relied on it: set it to its' previous value `1` in your "init.moon".
```
nilnor: One final nitpick: Initialization can be written in either Lua or Moonscript, so instead of referring to 'init.moon' I suggest 'Howl configuration' or 'initialization configuration', etc.

ozra: Right! That was a potential religious war entry XD!

nilnor:
:+1:

PR #333, opened 2017-03-11
Fix #327. `editor-smart-tab` aligns to tab-stops

### LINKED ISSUE #327: bug: `editor-smart-tab` has no smarts after non-ws
- At beginning of line it works fine
- _After_ non whitespace, it doesn't align to the `indent` setting like I'd expect it to (have never seen an editor or word-processor _not_ align tabs to tabulator positions) but rather seems to just insert `indent * spaces`

ozra:
I have some question marks and suggestions.

### DIFF
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -401,16 +401,31 @@ class Editor extends PropertyObject
       return
 
     conf = @config_at_cursor
-    if conf.tab_indents and @current_context.prefix.is_blank
-      cur_line = @current_line
-      next_indent = cur_line.indentation + conf.indent
-      next_indent -= (next_indent % conf.indent)
-      cur_line.indentation = next_indent
-      @cursor.column = next_indent + 1
-    else if conf.use_tabs
+
+    if conf.use_tabs
       @view\insert '\t'
-    else
+      if conf.tab_indents
+        @cursor.column = @current_line\ufind('%S', @cursor.column)
+      return
+
+    if not conf.tab_indents
       @view\insert string.rep(' ', conf.indent)
+      return
+
+    -- Do space-based tab-stops alignment
+    cur_line = @current_line
+    cur_col = @cursor.column
+    tabstop_wd = conf.indent
+    next_non_ws = cur_line\ufind('%S', cur_col) or #cur_line + 1
+    delta = tabstop_wd - (next_non_ws - 1) % tabstop_wd
+    next_tabstop = next_non_ws + delta
+
+    if @current_context.prefix.is_blank
+      cur_line.indentation = next_tabstop - 1
+    else
+      @view\insert string.rep ' ', delta
+
+    @cursor.column = next_tabstop
 
   smart_back_tab: =>
     if not @selection.empty
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Fix #327: `editor-smart-tab` to do aligned tab-stops in all positions.
+
 - Respect use_tabs option when commenting
 
 - Ensure scrolling works correctly for Gtk+ 3.4
--- a/spec/ui/editor_spec.moon
+++ b/spec/ui/editor_spec.moon
@@ -420,13 +420,6 @@ describe 'Editor', ->
         editor\smart_tab!
         assert.equal buffer.text, 'h\tƏllo'
 
-      it 'inserts spaces to move to the next tab if use_tabs is false', ->
-        config.use_tabs = false
-        buffer.text = 'hƏllo'
-        cursor.pos = 1
-        editor\smart_tab!
-        assert.equal string.rep(' ', config.indent) .. 'hƏllo', buffer.text
-
       it 'inserts a tab to move to the next tab stop if use_tabs is true', ->
         config.use_tabs = true
         config.tab_width = config.indent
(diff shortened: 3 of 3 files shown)


ozra:
Hmm, did that squash thing, I realize now that that might not be suitable during commenting.

Anyway the "empty-line indents in to outer space" is resolved, so it should be all good now!

lib/howl/ui/editor.moon
```diff
@@ -367,16 +367,34 @@ class Editor extends PropertyObject
...
-      cur_line.indentation = next_indent
-      @cursor.column = next_indent + 1
-    else if conf.use_tabs
+
+    if conf.use_tabs
       @view\insert '\t'
-    else
+      return
+
+    -- !NOTE!
```
nilnor: `tab_indents = false` is most noticeable when `use_tabs` is set to true I think. Consider:

```moonscript
foo ->
  bar!
```
With the cursor at the start of the second line and use tabs set to true, try toggling the option back and forth and see the difference.

WRT `indent`, I think that `indent` is widely used when referring to coding standards, whereas tabstop-width does not ring a bell. I don't think that sounds like a good change personally.

lib/howl/ui/editor.moon
```diff
@@ -367,16 +367,34 @@ class Editor extends PropertyObject
...
     conf = @config_at_cursor
-    if conf.tab_indents and @current_context.prefix.is_blank
-      cur_line = @current_line
-      next_indent = cur_line.indentation + conf.indent
-      next_indent -= (next_indent % conf.indent)
-      cur_line.indentation = next_indent
-      @cursor.column = next_indent + 1
-    else if conf.use_tabs
+
+    if conf.use_tabs
```
nilnor: When `tab_indents` is true and where in whitespace we should not insert any tab, so this would break the current and desired behaviour.

lib/howl/ui/editor.moon
```diff
@@ -367,16 +367,34 @@ class Editor extends PropertyObject
...
+    cur_col = @cursor.column
+    tabstop_wd = conf.indent
+    next_non_ws = cur_line\ufind('%S', cur_col) or #cur_line + 1
+    delta = tabstop_wd - (next_non_ws - 1) % tabstop_wd
+    next_tabstop = next_non_ws + delta
+
+    if @current_context.prefix.is_blank
+      cur_line.indentation = next_tabstop - 1
+    else
+      @view\insert string.rep ' ', delta
```
nilnor: This does not account for the `use_tabs` setting.

ozra: Ah, right, that setting is only checked at 371, and has already returned before this, blindly inserting. Will restructure to cover all the cases as commented.

lib/howl/ui/editor.moon
ozra: I'll try the settings out with master - since the post-content indents didn't behave as expected I wasn't sure which parts of the behaviour to consider correct.

_indent_: Fair enough, it is after all primarily a coders editor.

lib/howl/ui/editor.moon
ozra: Alright

ozra:
I'll:
- check the behaviour of all `in indent part of line` scenarios with different settings
- check the specs
- write new / change specs to cover all the cases

ozra:
Revisited. I just ran the specs and they passed. So I wrote some new, doing what I think is the reasonable thing for the cases I didn't know. And fixed func accordingly.

Holler if I still misunderstood it (when trying the main branch, toggling the options, tabs and spaces started mixing in an unexpected way when having use_tabs on, couldn't figure out any logic in it, so...)

nilnor:
@ozra Could you clean up the branch? Now it contains a lot of other irrelevant changes from master which makes it impossible to review properly.

ozra:
Sure, ehmm, my git-knowledge is rather "shallow", do you have a quick pointer for best approach all things considered?

nilnor:
It's very messy now, so I would suggest two possible routes:

- Take some time to learn about Git interactive rebasing (well worth your time) and rebase, keeping only your own commits
- Start a new branch based on master, and git cherry-pick only your own commits from the old branch. After that you can either just open a new PR, or if you feel up to it change the old branch to point to the new and force push that

Generally speaking, never merge into a PR branch - either rebase on top of master or just keep it as is unless there's something you actually need from master.

ozra:
Thanks for the advice! When my work-pressure is down I'll pick up those skills and get a clean PR up!

ozra:
Hope that did the trick!

lib/howl/ui/editor.moon
nilnor: This comment has not been addressed, and the behaviour with this PR remains broken. A recap of what this function should do: 

If we're in leading whitespace and `tab_indents` is true, we should indent and nothing else. Whether tabs are used in that case or not is already handled when setting a buffer line's indentation.

If the above is not true, then we should insert a tab, or a suitable number spaces. The current implementation of this is very straightforward: one tab, or <space> * `indent`

In an attempt to better align with tab widths it would then be desirable to make that last part smarter, by optionally inserting fewer spaces to make it align with a virtual tab stop. That seems like it should only require modifications for the last `else` clause in the current implementation, unless one also wants to skip tabs even when `use_tabs` is true in order to align correctly.

As it stands, I think the current implementation is correct and rather clear given the above requirements, but could be smarter for that last case. The PR now however makes it incorrect, and more unclear as to what the behaviour should be. I suggest just looking at modifying the last `else` clause and see if that is enough.

ozra: Alright, I'll revisit after Jul and hopefully get it almost completely right the next commit XD

ozra:
So, I haven't managed to get time off from biz-code, and I'm just using this the way I want by "overriding" the default from my init — so I'll simply close this, as no-one else seems to be much concerned about the non-alignment.

PR #342, opened 2017-03-28
Add `force` to cp howl command in Makefile install

ozra:
If Howl is running, the install won't complete the copy. I oftenly keep
several instances of Howl running for days (weeks) on end, and I might
not want to close down a certain session just to recompile. I think it's
the right thing (TM).


ozra:
Please merge this quickly if there are no objections. :-)
It's such a time waster to mod / de-mod the Makefile for every branch switch / rebase / push, yada yada B-).

nilnor:
Well I don't feel strongly against it, so unless someone else objects we can merge this soon.

shalabhc:
No objections from me.

ozra:
:+1:

PR #344, opened 2017-03-29
Config update shouldn't be applied to every editor

### LINKED ISSUE #343: set-for-buffer line_number affects all visible views
In master:

1. Start howl with `howl --no-profile`
2. Open two different Untitled buffers in two views
3. Ensure line numbers are visible in both views (the default)
4. In one buffer, execute 'set-for-buffer line_numbers=false'

Both views clear the line numbers, only one view should.

The same issue exists for `indent` and other variables as well.

shalabhc:
Instead just refresh the config settings for each editor.

Fixes #343

### DIFF
--- a/lib/howl/ui/editor.moon
+++ b/lib/howl/ui/editor.moon
@@ -30,15 +30,12 @@ indicator_placements =
   bottom_left: true
   bottom_right: true
 
-apply_variable = (option, value) ->
-  for e in *editors!
-    e.view.config[option] = value
-
 apply_global_variable = (name, value) ->
   aullar_config[name] = value
 
-apply_property = (name, value) ->
-  e[name] = value for e in *editors!
+apply_variable = ->
+  for e in *editors!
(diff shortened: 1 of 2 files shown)


nilnor:
It seems slightly heavy handed to always invoke `_set_config_settings` for each and every change. How about we keep the existing property handling, and for the other variables we set the view's config from the corresponding buffer's config value instead?

shalabhc:
Not sure I got what you mean. Are you saying we update all views for every change, but only update the config that changed? That seems reasonable.

nilnor:
We probably mean the same thing I think. I mean that we don't update all possible config values as currently in this PR, but only process the actual config value changes. We can check the config for the current buffer even, and only update the aullar config if the buffer's config value for the parameter equals the new updated value, which means we only update the aullar config if the value has actually changed for the current buffer.

shalabhc:
Withdrawn, will redo later.

PR #345, opened 2017-03-29
Fix config evaluation for layer defaults

shalabhc:
Config layer defaults should take precedence over global values.

### DIFF
--- a/lib/howl/config.moon
+++ b/lib/howl/config.moon
@@ -182,34 +182,51 @@ set_default = (name, value, layer='default') ->
 
   layer_defs[layer].defaults[name] = value
 
+  broadcast name, value, true
+
+get_default = (name, layer='default') ->
+  if layer_defs[layer].defaults
+    return layer_defs[layer].defaults[name]
+
 parent = (scope) ->
   pos = scope\rfind('/')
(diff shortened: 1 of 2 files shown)


shalabhc:
I think this bugfix is good. - merging.

ozra:
New stuff. Is good stuff. Trying it out when I wake up tomorrow :)

shalabhc:
@ozra - it doesn't fix your reported issue completely, but only one aspect. Specifically, `mode.configure` should now work (but switching away and back to a buffer may still be needed).

ozra:
Alright @shalabhc. I hope I can figure out a way to shift my "active spare time" some, soon, to get more acquainted with the code. I'm already (_too_) head-in into a couple of  long term OSS-projs that means a lot to me—and as I've determined: society—aside from the "must-dos" of the farm, family and life, but _finding this editor is such a good thing (TM)_ that I simply must (_want_) to relocate time to it (one's primary tool of getting shit done _is_ worth investing in), so I _hope_ to be more helpful in debugging / giving better feedback and/or PR'ing when I've figured out how to juggle the hours.

**Thank you guys for having made this!!!** Keep it up!

Issue #235, opened 2016-06-30
Clipboard weirdness


kirbyfan64:
This is somewhat odd and hard to reproduce accurately, but, sometimes, if I open up Howl, then copy something from, say, Chrome, and try to paste it into Howl, it won't paste. Again, it's super hard to reproduce correctly, but I'm trying to see if I can at least get a modest test case.

nilnor:
I recognize this fwiw, but I don't have an easy solution to it unfortunately. The problem occurs when you have more than one instance open as far as I can tell. I therefore speculate that it's something to do  with dbus/X11/Gtk (as we register the application on dbus via GApplication). So I don't think it's necessarily a bug in Howl per se, but I could of course be wrong - perhaps there's something we're doing wrong in the app init, etc., which has this effect. Anyway, perhaps there's a work around to be found, but it happens only now and then as you've said so that makes is harder to experiment. That it happens with multiple instances open is as far as I've narrowed it down until now, but if a reproducible test case could be found that would be great.

PR #346, opened 2017-03-29
Primary selection

### LINKED ISSUE #308: X11 primary selection
Hi, 

is there a way to enable the X11 primary selection, i.e. pasting with middle mouse button ?

nilnor:
This enables the use of the X11 primary selection from within Howl. It allows for pasting from the primary selection using the middle button, and also integrates the Howl selection with the primary selection in the system.

Fixes #308.

### DIFF
--- a/lib/howl/clipboard.moon
+++ b/lib/howl/clipboard.moon
@@ -3,6 +3,8 @@
 
 Atom = require 'ljglibs.gdk.atom'
 GtkClipboard = require 'ljglibs.gtk.clipboard'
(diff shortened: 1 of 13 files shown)


kirbyfan64:
Maybe this would also fix #235? I can never accurately reproduce it, but it still happens all the time, which seems to fit in line with X11 clipboards.

shalabhc:
@kirbyfan64 - hope so too! Going to try this out now to see (I was hit with #235 recently as well.)

shalabhc:
Woohoo - just tried it out and middle click works for me!

PR #348, opened 2017-04-04
A new set command that unifies and replaces all existing set commands

shalabhc:
The new set command has the following form:

```
  set <name>@<scope>[<layer>]=<value>
```

The [<layer>] part is optional.

Examples:
```
  set theme@global=Monokai
  set indent@global[mode:moonscript]=2
  set indent@project[mode:python]=4
  set auto_inspect@buffer=manual
```
Selection lists provide auto-complete for the `name` and `scope` part,
making it generally unnecessary to type out the whole thing.

This is now implemented as a sequence of three interactions, rather
than one complex text entry interaction.

Backspacing backtracks to the previous interaction, allowing you to
edit any part of the command, but different than the current behavior
in that it clears the whole previous selection, rather than deleting
one character. (In that sense it is similar to backspacing into the
parent directory in the file browser).

### DIFF
--- a/lib/howl/interact.moon
+++ b/lib/howl/interact.moon
@@ -21,13 +21,15 @@ sequence = (order, def) ->
 
     state = {}
     pos = 1
+    local current
 
     while true
       def.update state if def.update
 
+      prev = current
       current = order[pos]
       if current
-        result = def[current](state)
+        result = def[current](state, prev)
         return nil unless result
 
         if result.back
--- a/lib/howl/interact.moon
+++ b/lib/howl/interact.moon
@@ -15,9 +15,36 @@ register = (spec) ->
 unregister = (name) ->
   interactions[name] = nil
 
+sequence = (order, def) ->
+    for name in *order
+      error "#{name} not found in def" unless def[name]
+
+    state = {}
+    pos = 1
+
+    while true
+      def.update state if def.update
+
+      current = order[pos]
+      if current
+        result = def[current](state)
+        return nil unless result
+
+        if result.back
+            pos -= 1
+            state[order[pos]] = nil
+        else
+            state[current] = result
+            pos += 1
+        continue
+      else
+        if def.finish
+          return def.finish state
+        return state
+
 get = (name) -> interactions[name]
 
-return setmetatable {:register, :unregister, :get}, {
+return setmetatable {:register, :unregister, :get, :sequence}, {
   __index: (interaction_name) =>
     spec = get(interaction_name)
     return if not spec
--- a/lib/howl/interactions/variable_assignment.moon
+++ b/lib/howl/interactions/variable_assignment.moon
@@ -74,10 +74,13 @@ interact.register
           { header: 'Description', style: 'comment' }
         }
 
-      scope: (state) ->
+      scope: (state, from_state) ->
         def = state.var.selection.def
         if def.scope == 'global'
-          return {selection: {'global', scope: 'global'}}
+          if from_state == 'value'
+            return { back: true }
+          else
+            return { selection: { 'global', scope: 'global' } }
 
         interact.select
           title: def.name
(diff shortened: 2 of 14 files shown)


ozra:
Why not `set indent@mode:moonscript=2` or something terser like that?

shalabhc:
@ozra - that would be fine too. But any scheme would also have to support specifying both mode *and* project. How would that be written?

lib/howl/interactions/variable_assignment.moon
```diff
@@ -36,99 +30,151 @@ option_current_value = (def, options, current_buffer) ->
...
-    escape: =>
-      self.finish!
+  {
+    scope_str(scope_name, layer), to_s(target[def.name]), description,
+    :scope_name, :layer, :target,
+    quick_select: scope_str(scope_name, layer) .. '='
+  }
+
+scope_items = (def, buffer) ->
+  mode_layer = buffer and buffer.mode.config_layer
```
nilnor: Since we have an early return a few lines below, these two lines could be moved below that and be simplified.

lib/howl/interactions/variable_assignment.moon
```diff
@@ -36,99 +30,151 @@ option_current_value = (def, options, current_buffer) ->
...
+
+  append items, scope_item(
+    def, 'global', config, "For all buffers with mode #{mode_name}", mode_layer)
+
+  if buffer.file
+    project = Project.for_file buffer.file
+    if project
+      append items, scope_item(
+        def, 'project', project.config,
+        "For all files under #{project.root.basename}")
```
nilnor: Perhaps use `short_path` here? Makes it's clearer that it's a directory we're talking about.

site/source/doc/manual/configuration.md
```diff
@@ -138,12 +178,8 @@ mode.configure 'c', {
...
 }
 
--- Hook up a signal handler to set it to three for this weird project
-that_project_root = File '/home/nino/code/that_project'
-
-signal.connect 'file-opened', (args) ->
-  if args.file\is_below that_project_root
-    args.buffer.config.indent = 3
+-- Set it to three for this weird project
+config.for_file('/home/nino/code/some_project').indent = 3
```
nilnor: Good change here that illustrates the new power :)

spec/config_spec.moon
```diff
@@ -544,3 +544,50 @@ describe 'config', ->
...
+          .name2 = 'dir-value2'
+        with config.for_file(dir .. 'abc')
+          .name1 = 'file-value1'
+          .name3= 'file-value3'
+        assert.same 'file-value1', config.for_file(dir .. 'abc').name1
+        assert.same 'dir-value2', config.for_file(dir .. 'abc').name2
+        assert.same nil, config.for_file(dir).name33
+        assert.same nil, config.name3
+
+    context 'for_layer', ->
```
nilnor: Incorrectly indented

lib/howl/interact.moon
```diff
@@ -15,9 +15,38 @@ register = (spec) ->
 unregister = (name) ->
   interactions[name] = nil
 
+sequence = (order, def) ->
```
nilnor: Worth adding a spec for?

shalabhc: Done.

spec/config_spec.moon
shalabhc: Fixed now.

lib/howl/interactions/variable_assignment.moon
shalabhc: Done.

lib/howl/interactions/variable_assignment.moon
shalabhc: Done.

site/source/doc/manual/configuration.md
ozra: I actually use 3 space indent for everything not part of others' projects - It's the perfect indent!
But "weird" sounds cool :-)

nilnor:
:+1:

PR #350, opened 2017-05-04
Fix config broadcast handling for editors

shalabhc:
Fixes #344 

(Second attempt at fixing this - this time the code is much smarter about not updating stuff unnecessarily)


nilnor:
:+1:

PR #351, opened 2017-05-12
Allowing running different inspectors for idle and save

nilnor:
The previous solution had only one list of inspectors, and you could
chose when those would run (idle, save or manually). While this was a
simpler (and thus nicer in that sense) solution, it didn't allow for
running lightweight inspections on idle and heavy weight inspections
on save.

With this split there are now two configuration variables for
controlling what inspectors are run, `inspectors_on_idle` and
`inspectors_on_save`. The `auto_inspect` option now allows the
following three alternatives for controlling the running behaviour:

- manual: Run all inspectors when explicitly asked
- idle: Run on-idle inspectors on idle and on-save inspectors on save
- save: Run all inspectors, but only on save

@shalabhc How about this for your heavyweight stuff?

### DIFF
--- a/lib/howl/variables/core_variables.moon
+++ b/lib/howl/variables/core_variables.moon
@@ -34,8 +34,15 @@ config.define
   scope: 'global'
 
 config.define {
-  name: 'inspectors'
-  description: 'List of inspectors to run for a buffer'
+  name: 'inspectors_on_idle'
+  description: 'List of on-idle inspectors to run for a buffer'
+  type_of: 'string_list'
+  default: {}
+}
+
+config.define {
+  name: 'inspectors_on_save'
+  description: 'List of on-save inspectors to run for a buffer'
   type_of: 'string_list'
   default: {}
 }
@@ -45,9 +52,9 @@ config.define {
   description: 'When to automatically inspect code for abberrations'
   default: 'idle'
   options: {
-    { 'manual', 'Only inspect when explicitly asked' }
-    { 'idle', 'Inspect on idle' }
-    { 'save', 'Inspect when saving a buffer' }
+    { 'manual', 'Run all inspectors when explicitly asked' }
+    { 'idle', 'Run on-idle inspectors on idle and on-save inspectors on save' }
+    { 'save', 'Run all inspectors, but only on save' }
   }
 }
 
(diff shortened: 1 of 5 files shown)


lib/howl/variables/core_variables.moon
```diff
@@ -45,9 +52,9 @@ config.define {
   description: 'When to automatically inspect code for abberrations'
   default: 'idle'
   options: {
-    { 'manual', 'Only inspect when explicitly asked' }
-    { 'idle', 'Inspect on idle' }
-    { 'save', 'Inspect when saving a buffer' }
+    { 'manual', 'Run all inspectors when explicitly asked' }
+    { 'idle', 'Run on-idle inspectors on idle and on-save inspectors on save' }
+    { 'save', 'Run all inspectors, but only on save' }
```
shalabhc: I feel these three options are useful but names can be confusing? E.g. 'idle' runs save inspectors too (and vice-versa). Maybe 'off' for manual (*auto*_inspect is off implies manual), 'on' for the default setting and 'save' (can't think of a better name here).

shalabhc:
Looks good, added one comment. But haven't tried it out yet - will test soonish.

nilnor:
@shalabhc Updated with new option names.

shalabhc:
OK I tried inspectors_on_save and it worked as expected.

One thing that was not great in the workflow was that all inspections clear on any edit - say when I fix the first inspection. Since it is an 'on_save' inspector, I can't see the rest of the inspections until I save again and wait for the inspections. This is also a problem for on_idle inspections (though not as much and depends on the inspection delay).

So it would be nice to only clear the inspection that was fixed (this is impossible to know for sure, but clearing inspections on the edited line only might be a good approximation?).

This can be fixed or improved in a later CL too so I as-is this CL is good to go.

nilnor:
@shalabhc great, I'll merge this now then. About the workflow, I'd suggest that the current behaviour is more correct than the one you suggest, although not necessarily the best one for you. There's no way of knowing whether any inspections are still valid after an edit, so keeping inspections necessarily means running the risk of displaying incorrect an stale information. Still, this might be preferable in some cases, but I don't see it being the default, and the display would optimally show in some way that the inspections could be stale.

PR #353, opened 2017-05-19
Allow bundles to exposed vendored lua modules

nilnor:
Allow bundles to expose modules using `provide_module` helper
    
This allows for a bundle to easily vendor an entire lua module. The `provide_module` helper exposes this both internally within the bundle and externally with Howl, so that any files can be `require`d just as if they we're on the default search path.


shalabhc:
This sounds like a great idea!

PR #354, opened 2017-05-22
Vi Enhancements

jlombs:
Fixes a few small issues with the vi-bundle in regards to conforming to expected vi behavior, including t/T motion fixes when no match is found, cc behavior, and cc with visual selection. 

(Sorry about my commit history this time--requires some squashing upon the merge if accepted...can't learn if you don't make mistakes!).

### DIFF
--- a/bundles/vi/base_map.moon
+++ b/bundles/vi/base_map.moon
@@ -31,7 +31,8 @@ forward_till_char = (event, source, translations, editor) ->
     apply editor, (editor) ->
       current_pos = editor.cursor.pos
       editor\forward_to_match event.character
-      if current_pos != editor.cursor.pos then editor.cursor\left!
+      if current_pos != editor.cursor.pos
+        editor.cursor\left!
   else
     return false
 
@@ -40,7 +41,8 @@ back_till_char = (event, source, translations, editor) ->
     apply editor, (editor) ->
       current_pos = editor.cursor.pos
       editor\backward_to_match event.character
-      if current_pos != editor.cursor.pos then editor.cursor\right!
+      if current_pos != editor.cursor.pos
+        editor.cursor\right!
   else
     return false
 
--- a/bundles/vi/spec/vi_spec.moon
+++ b/bundles/vi/spec/vi_spec.moon
@@ -105,6 +105,8 @@ describe 'VI', ->
   it '<t><character> searches forward in the current line to one character before <character>', ->
     press 't', 'n'
     assert.equal 2, cursor.column
+
+    -- No movement when match not found
     press 't', 'q'
     assert.equal 2, cursor.column
 
@@ -112,6 +114,8 @@ describe 'VI', ->
     cursor.column = 5
     press 'T', 'i'
     assert.equal 3, cursor.column
+
+    -- No movement when match not found
     press 'T', 'q'
     assert.equal 3, cursor.column
 
--- a/bundles/vi/spec/vi_spec.moon
+++ b/bundles/vi/spec/vi_spec.moon
@@ -127,8 +127,6 @@ describe 'VI', ->
     assert.equal 'LinƏ 1\n\nAnd third linƏ\n', buffer.text
     assert.equal 'insert', state.mode
 
-
-
     -- empty lines
     buffer.text = '\n\n'
     cursor.pos = 1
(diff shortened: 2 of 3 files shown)


bundles/vi/base_map.moon
```diff
@@ -29,16 +29,18 @@ back_to_char = (event, source, translations, editor) ->
 forward_till_char = (event, source, translations, editor) ->
   if event.character
     apply editor, (editor) ->
+      current_pos = editor.cursor.pos
       editor\forward_to_match event.character
-      editor.cursor\left!
+      if current_pos != editor.cursor.pos then editor.cursor\left!
```
nilnor: Style nitpick: We don't typically do `then` like that, so either move the action clause below on it's own line or go with a statement modifier, e.g.:

`editor.cursor\left! unless current_pos == editor.cursor.pos`

The first alternative would probably look better here. Ditto for the change below.

bundles/vi/spec/vi_spec.moon
```diff
@@ -105,17 +105,35 @@ describe 'VI', ->
   it '<t><character> searches forward in the current line to one character before <character>', ->
     press 't', 'n'
     assert.equal 2, cursor.column
+    press 't', 'q'
```
nilnor: I realize that this testes the cases where the cursor shouldn't move when no match is found, but that's not clear here just by reading this. Either a short comment here, or alternatively one could expand it into a block, e.g. 

```moonscript
describe '<t><character>', ->
  it 'searches forward[..]', ->
  it 'does nothing if <character> is not found', ->
```

jlombs:
Thanks for the review @nilnor. I've changed the style for the logic, and added comments on the spec to clarify instead of doing a separate block, since that's what was used in the spec for `dd` when checking the primary behavior and the behavior on empty lines

jlombs:
@nilnor Not sure why the build seems to have failed here, since it passed at the last commit and I only changed comments and style based on your review, which again passed specs locally. Should I trigger a rebuild, or is there something else at work here? Sorry, first time using Travis.

kirbyfan64:
Uhh...yeah, that spec's been weird in the past. I'll try closing and re-opening the PR.

kirbyfan64:
...to trigger a rebuild (uuugggh GitHub on mobile).

jlombs:
@kirbyfan64 Thanks. It's weird: under the Travis "more options--requests" tab both your recent trigger and my original one (`#1736/7`) have "build created successfully" tags (if I'm reading that right), but the individual requests show failures on their respective detailed views...

shalabhc:
So I clicked 'restart build' on travis and looks like this run was successful?

jlombs:
@shalabhc Welp, that's unnerving, but better a pass than a fail I guess. Thanks for running the build again

shalabhc:
@jlombs  - the test that failed earlier and passed on rerun is a memory usage checker and somewhat flaky because memory allocations patterns are not reproducible.

jlombs:
@shalabhc I see. For the future, is there a way to differentiate a genuine failure of the memory usage test from a false failure? Example: using the scale of the memory diff? Similarly, can a passing build have a minor memory leak that goes undetected? Just curious

shalabhc:
@jlombs - memory usage testing is tricky. If the threshold is too big, material memory increases can go undetected. If the threshold is too small, we get false alarms (which is what happened here). Memory allocation is not deterministic so multiple runs of the test from the same code produce different usage metrics. @nilnor added the memory test and can add more context.

Maybe we should run the memory test three times and use the best (i.e. lowest) memory metric to determine if the test fails. What do you think @nilnor?

nilnor:
It's as @shalabhc says, the memory usage tracking is frail, and the limit needs to be not too big nor too small. The tests themselves are currently run multiple times and averaged, but even so there can be rather big differences between runs, which I contribute in no small part to memory fragmentation. The recent update of LuaJIT also brought about changes in memory management it seems, since we've hit a lot of these after that upgrade. I'm pondering whether to remove these specs altogether due to their flakyness, but then memory issues due to reference issues would go undetected (and they're hard to track down starting from scratch).

Anyway, I merged this PR manually, I rebased to get rid of the unwanted merges in the branch, but kept the individual commits.

PR #356, opened 2017-05-28
Refresh editor config when buffer mode changes.

### LINKED ISSUE #347: buffer-mode change doesn't update editor properties
0. Run 'howl --no-profile' on master
1. Open a new Untitled buffer
2. buffer-mode moonscript
3. set-for-mode line_numbers=false
4. Line numbers are hidden
5. buffer-mode lua
6. Run 'set line_numbers' - shows effective value is true for buffer

Expected line numbers to be shown again but they are not.

shalabhc:
- Fixes #347
- Also don't set the config value unless different from existing value


nilnor:
:+1:

PR #360, opened 2017-06-12
Changes to allow the build in the OpenBSD Operating System.

oficial:
Hi folks. 
I try to build from source in the OpenBSD operatiing system, and need to make these two little changes.
This is basically the same rules for building on FreeBSD.

virtualbox_openbsd - snapshot_12_06_2017_17_44_18

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -25,6 +25,9 @@ LIBS = -lm -ldl ${GTK_LIBS}
 ifeq ($(UNAME_S),FreeBSD)
     LIBS = -lm ${GTK_LIBS}
 endif
+ifeq ($(UNAME_S),OpenBSD)
+    LIBS = -lm ${GTK_LIBS}
+endif
 ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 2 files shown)


kirbyfan64:
I wonder if we should just be checking for any OS ending in `BSD`...

oficial:
I just use OpenBSD and FreeBSD... i can try to install NetBSD on a VM tomorow to see if this works too.

oficial:
Hi, guys. 
I install NetBSD in a VM but it freeze two times during boot, and i have to kill the VirtualBox process.
So i ask in a C/C++ group and it was like i am talking some kind of alien language :)
After that i ask a guy who i know to work in the NetBSD kernel. After a couple of days he respond that in NetBSD it probably works the same as in Free/Open. 
So i think that it is OK to use something like *BSD. 
Other desktop oriented BSDs, like PCBSD, will probably work too.

shalabhc:
Thanks for the update @oficial. I would hope uname on PCBSD just reports as 'FreeBSD', but I may be wrong.

In any case, I say lets merge this first and fix the *BSD issue later.

nilnor:
Agreed, thanks @oficial!

PR #361, opened 2017-06-15
Add a mypy inspector for Python code

kirbyfan64:
@shalabhc Please feel free to try and break this! :D

Note that if the source file contains a `from .module import whatever`, then mypy will crash; it's fixed in [this PR](https://github.com/python/mypy/pull/3554).


shalabhc:
Yay!

But why isn't this is separate `bundle/mypy`?

kirbyfan64:
@shalabhc Partly because I didn't even think about it, partly because it seemed really odd to me to have one bundle (Python) depend on another, so closely related one (mypy).

shalabhc:
Well the Python bundle shouldn't depend on it. If you remove the default inspector, that should be it?

Given mypy is a separate project (and far from even a defacto Python checker), it should be in a separate bundle, I think.

kirbyfan64:
@shalabhc Fair point. I just changed it.

shalabhc:
BTW, is this supposed to work for unsaved files? This might be better as 'inspectors_on_save' (just a readme update)

kirbyfan64:
@shalabhc It works on unsaved files, though I did update the README to mention `inspectors_on_save`.

nilnor:
I'm thinking that this shouldn't necessarily go into core. We haven't really had an official guideline for this, but the bundles so far can basically be divided up into what I'd call "canonical" language bundles and generic cross language bundles offering functionality aimed at general editing or usage. I think it makes sense to keep it that way and have additional, more specific, support bundles be maintained outside of Howl. 

Eventually in some future I can see us ending up with a repository type of bundle manager, at which point nearly all bundles could potentially be maintained outside of the core, but until then I think we should adopt some scope policy for core bundles.

kirbyfan64:
I *do* understand that, but one thing worth nothing is that many Python editors are now adopted support for type annotations, and it's very quickly becoming a major tool for a lot of Python programs.

Of course, if you still prefer it outside, I'll leave it as such. ;)

shalabhc:
Yeah lets leave it out for now. It's possible someone makes a language-server-protocol for this stuff too in which case that would be another way to integrate it. In any case, we can revisit this once mypy gets a little more established as the de facto standard?

kirbyfan64:
Fair enough.

PR #364, opened 2017-06-25
Add a bundle for Dart

shalabhc:


kirbyfan64:
OH MY GOSH I've been wanting to do this for ages but kept getting distracted.

FWIW is would be neat if at some point support for the Dart analysis server were added.

shalabhc:
Yes, the analysis stuff would be great to add later. Also it currently doesn't highlight the $ formatting within strings - so that's another thing to be done in the future.

nilnor:
:+1:

Issue #349, opened 2017-05-02
New Release


Immortalin:
0.4.2?

shalabhc:
The master branch is pretty stable and that's what most developers use.

But I agree we should probably do another release soon - might even be 0.5 - there are some major unreleased updates.

nilnor:
For the record, for me I want to get new inspections in a better shape first: Update to split between idle and save inspectors, add support for Luacheck for Lua, and add documentation. It's in progress now.

ozra:
I just fetched HEAD and rebuilt just minutes ago, and `./howl --version` still says 0.4.1, has the ver been bumped in the source code, or have I fucked up completely somehow? (last commit seen in `git log` is db3d814507e8ffcd92bee60fd10998969a5b6cfc)

kirbyfan64:
No, it's not you: #366

shalabhc:
So I wonder if running `howl --version` on trunk should say something like '0.6-devel'?

ozra:
It could be neat to get both the version tag and last commit hash automatically at compile time, and perhaps commit-count since last tag - elliminating manual version edits minimizes confusion :-)

nilnor:
I didn't do anything fancy with regards to tags, etc., but added a sanity check for the release script. Also, I released 0.5.1 now which should hopefully have a correct version. In master the version is now 0.6-dev.

PR #366, opened 2017-07-04
Update init.lua for Howl 0.5.0

kirbyfan64:
https://github.com/howl-editor/howl/issues/349#issuecomment-312930234


kirbyfan64:
TBH we should be deriving versions from Git tags or something else anyway, but this works for now!

PR #367, opened 2017-07-07
Allow completions to be tables in addition to strings

shalabhc:
When a table is specified, the value of the 'completion' key is
inserted on accepting the completion. Tables are automatically
rendered as styled table and accept markup.


nilnor:
:+1:

PR #368, opened 2017-07-11
Don't reuse table from property object base class

### LINKED ISSUE #363: Overriding a property in a child class also overrides it in the base class
e.g. this spec fails:

```moonscript
class Parent extends PropertyObject
  @property foo:
    get: => 'parent'

class SubClass extends Parent
  @property foo:
    get: => 'sub'

assert.equal Parent!.foo, 'parent'
assert.equal SubClass!.foo, 'sub'
```

Found while investigating https://github.com/kirbyfan64/aisu/issues/2.

shalabhc:
Fixes #363 

Spec copied from what @kirbyfan64 reported

### DIFF
--- a/lib/howl/util/moon/property_object.moon
+++ b/lib/howl/util/moon/property_object.moon
@@ -2,7 +2,11 @@
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 class_properties = (cls) ->
-  cls.__base.__properties = {} if not cls.__base.__properties
+  if cls.__base.__properties
+    if rawget(cls.__base, '__properties') != cls.__base.__properties
+      cls.__base.__properties = moon.copy cls.__base.__properties
+  else
+    cls.__base.__properties = {}
   cls.__base.__properties
 
 meta_methods = (cls) ->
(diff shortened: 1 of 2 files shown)


nilnor:
Since overriding a property wasn't originally considered, I'm wondering about other parts of this (with this change) supported functionality. For instance, defining a property in a super class and then overriding only a getter? Does this work? In short, if we're going to support this we should make sure to cover the cases, otherwise an alternative would be to just explicitly not support it.

kirbyfan64:
Nice! FWIW I tried to fix this, but then I realized I had no clue how the hell the property object code worked...

TBH I'd say things like overriding a read-wrire property with a read-only one should probably just be an error; that's the type of stuff that would normally be caught e.g. by a type checker.

shalabhc:
@kirbyfan64  what about overriding the `get` of a read-write property but inheriting the `set`? 

Maybe there's not good use case for that (is the read logic changes, presumably the write logic should change too). So we could have a rule that the override property must have exactly the same of `get|set` methods that it is overriding. Does that address your concern @nilnor?

Also, should we allow overrides by `@property` or require explicit `@property_override`?

nilnor:
I think any kind of explicitly defined behaviour is OK with me - if we want to support overriding I just want all kinds of potential usages to be covered and decided upon.

shalabhc:
Closing an obsolete PR. I think this is still workable if the concerns in the discussion are addressed - essentially whatever is supported is explicitly defined behavior.

PR #372, opened 2017-07-24
Add editor-newline-above and editor-newline-below commands

mrshankly:
I find these commands quite useful and seeing that they exist in pretty much every text editor I thought that these might be good additions.

The diff for keymap.moon is big because I had to change the alignment by one space. The new lines are [27](https://github.com/howl-editor/howl/pull/372/files#diff-b520b991e39bc22b2803ed93b56b316fR27), [28](https://github.com/howl-editor/howl/pull/372/files#diff-b520b991e39bc22b2803ed93b56b316fR28), [150](https://github.com/howl-editor/howl/pull/372/files#diff-b520b991e39bc22b2803ed93b56b316fR150) and [151](https://github.com/howl-editor/howl/pull/372/files#diff-b520b991e39bc22b2803ed93b56b316fR151).


shalabhc:
Great - thanks! Is re-positioning the cursor intentional?

mrshankly:
Do you mean placing the cursor on the newly created line? If so, yes it's intentional. That's the behaviour I'm used to seeing in other text editors at least.

shalabhc:
Yes that's what I mean. OK - this change looks fine!

nilnor:
The commands look fine to me! I'm not a fan of the new key bindings though, perhaps mostly because I envision `ctrl_enter` as being used for future navigation support. For that reason I don't want to give that away :) Other thoughts?

mrshankly:
I picked the key bindings used by other editors in order to keep some familiarity. I don't know, I think having the key bindings be some `modifier` combination + `enter` makes sense since the commands behaviour is similar to the behaviour of just pressing `enter`.

There is always the option of adding the commands without a default key binding and let the user choose one in the configuration file.

nilnor:
@shalabhc @kirbyfan64 what do you think?

kirbyfan64:
What about using Alt-O and Alt-Shift-O? It's analogous to Vim's O and o commands.

shalabhc:
I don't have any strong opinion regarding shortcut keys for this. @mrshankly can you tell us which text editors use the proposed shortcuts though? I'd like to see how universal it is.

mrshankly:
Sublime Text, Atom and Visual Studio Code offer these on Ctrl+Enter and Ctrl+Shift+Enter, and like @kirbyfan64 said, Vim with o/O. Not sure if emacs has something analogous.

shalabhc:
If ST, Atom and VSCode use these shortcuts, I'm fine with having them as the default.

nilnor:
All right, merging. Thanks @mrshankly!

PR #374, opened 2017-08-10
Better EOL handling

### LINKED ISSUE #369: Inspection results not shown correctly
Hi,
when I use `ctrl_shift_e` to inspect a lua script, the offending lines in the results buffer get mixed (if there are offending lines).
First line is fine, second is empy, third line is the line which is supposed to be the second line, and so on. I tried latest git.
Attached screenshots:

First line is fine.

2017-07-12-100842_683x736_scrot

Second line shows nothing.

2017-07-12-100859_682x742_scrot

Third line shows, what should be in second line.

2017-07-12-100930_678x740_scrot

nilnor:
Automatically set EOL after loading a file, and fix buffer lines's .chunk to account for EOL according to @shalabhc's pointer.

Fixes #369

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -50,6 +50,13 @@ class Buffer extends PropertyObject
       if file.exists
         @text = file.contents
         @sync_etag = file.etag
+
+        -- update EOL from loaded file if possible
+        first_line = @_buffer\get_line 1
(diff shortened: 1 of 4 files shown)


shalabhc:
Looks good!

(And I'm OK with the fact that this won't work with mixed EOL files.)

Issue #379, opened 2017-09-18
Odd bug with Python f-strings


kirbyfan64:
TL;DR: when using f-strings, the end of the string will sometimes randomly not be picked up. I think. I actually have no freaking clue.

Also, this is *really* hard to get a small test case for.

Basically, I open up [this file](https://github.com/kirbyfan64/sawanobot/blob/master/sawanobot.py) and scroll to line 184. It looks like this:

![howl_029](https://user-images.githubusercontent.com/1690697/30563787-9dce9a72-9c88-11e7-8b72-1c43ac4bcd0b.png)

HOWEVER, if I go to line 164, remove the `f`, and add it back, then the syntax highlighting goes back to normal.

This is probably something I did wrong when I added f-strings support to Python, but I'm not sure what. Same thing has happened when I added interpolation support to the Dart lexer...

PR #380, opened 2017-09-18
Dart bundle enhancements

kirbyfan64:
~~Hey look, another PR that I'm totally not going to abandon for like 27 months!~~

This:

- Makes syntax highlighting more consistent. `true` and `false` are now `special` just like `null`. I also tweaked a few other keywords.
- Adds some more operators, for consistency.
- Fixes the `identifer` typo that's in like *every single bundle*.
- Adds support for Dart's string interpolation.
- Fixes type lexing for generic types and function definitions to also count built-in types.
- Fixes the `named_param` lexer to avoid highlighting function argument *definitions* (since now `this.x: 123` and `this.x = 123` are the same thing, but the former was halfway-highlighted as a keyword argument).

The *DON'T MERGE* is because the interpolation support here has the same bug as Python's (#379), but I figured I might as well get some review first ~~because I'm so slow~~.


kirbyfan64:
Also, I swear I'm copying #381 almost exactly, but I still get the weird issues with interpolations never closing...

nilnor:
@kirbyfan64 if you have an example of the interpolation issues I can have a look

kirbyfan64:
@nilnor [This file](https://raw.githubusercontent.com/kirbyfan64/vuedart/master/lib/transformers/vue_transformer.dart). If you scroll to the bottom by either holding the down arrow key or using a scroll wheel/two-finger scrolling, you'll see it not highlighted correctly. *However*, if you instead use the scrollbar or the page down key, it looks correct. :thinking:

kirbyfan64:
FWIW I was trying to get analysis server support worked into this PR, but I'm not going to be able to finish that for a bit, so I removed those commits. Therefore, this should be pretty much done other than that weird format string issue.

nilnor:
The issue with the multiline lexing is that it does not follow an important part of the lexing protocol, namely that multiline lexing constructs such as multiline strings, here docs, etc., must be lexed so that any contained EOL's are styled with something else other than 'whitespace'. This is not the case with the Dart lexer currently, as it currently leaves contained EOLs within interpolations as unstyled:

```dart
    rewriter.edit(unit.end, unit.end, '''
@initMethod
void vuedart_INTERNAL_init() {
${components.map((comp) =>
      "  VueComponentBase.register(${comp.name.name}.constructor);").join('\n')}
}
    '''); 
```

`${components.map((comp) =>` ends up with an unstyled EOL here, which breaks things. You can check these things by placing the cursor at EOLs and using the `describe-style` command.

kirbyfan64:
Ohhhhhh...FWIW is that documented anywhere? :thinking: 

Anyway, I made a few more final touches and also fixed nested generics lexing (and made it a tad bit less yellow). *Now* this should be ready to merge!

nilnor:
@kirbyfan64 I don't think it is unfortunately. Ideally there would be a developer document that deals with lexing, but time is as ever short.

nilnor:
:+1:

PR #396, opened 2017-11-13
Smooth scrolling support

kirbyfan64:
Just installed Arch Linux, only to discover my precious two-finger scrolling wasn't working...turns out, Gtk+ and Wayland mixed don't support two-finger scrolling without smooth scrolling.


nilnor:
Good work!

PR #407, opened 2018-01-03
Automatically remove an empty buffer when another one is added

kirbyfan64:
In short, when you have no buffers open, an Untitled empty buffer will be added. Then, when you proceed to open another file...the untitled buffer still stays open. This will now automatically close it.

### DIFF
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -343,10 +343,13 @@ class Application extends PropertyObject
       count += 1
 
   _append_buffer: (buffer) =>
-    if config.autoclose_single_buffer and #@_buffers == 1 and not @_buffers[1].file and not @_buffers[1].modified
-      @_buffers[1] = buffer
-    else
-      append @_buffers, buffer
+    if config.autoclose_single_buffer and #@_buffers == 1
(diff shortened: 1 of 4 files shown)


lib/howl/application.moon
```diff
@@ -313,6 +313,12 @@ class Application extends PropertyObject
     while count < max_count and C.g_main_context_iteration(ctx, false) != 0
       count += 1
 
+  _append_buffer: (buffer) =>
+    if config.autoclose_single_buffer and #@_buffers == 1 and not @_buffers[1].file and not @_buffers[1].modified
```
nilnor: This can be stricter. While unlikely, it's possible to have an unmodified buffer with no file that would still be interesting. I suggest requiring a couple of different criteria for the buffer:

- No associated file
- Not modified
- Empty
- Optionally we could also check that there haven't been any revisions added

kirbyfan64:
@nilnor Done.

nilnor:
Sorry, could have mentioned this the first time had I thought about it, but a spec would be nice here (and shouldn't be hard to write).

PR #408, opened 2018-01-06
Add additional Scheme source extensions

stutonk:
The following extensions are included:
	.sld -- R7RS Library Source
	.sls -- R6RS Library Source
	.ss  -- 'Scheme Source' as used in some implementations (e.g. Chez)

### DIFF
--- a/bundles/basic_modes/mode_definitions.moon
+++ b/bundles/basic_modes/mode_definitions.moon
@@ -312,7 +312,7 @@ common_auto_pairs = {
     parent: 'curly_mode'
 
   scheme:
-    extensions: { 'sch', 'scm' }
+    extensions: { 'sch', 'scm', 'sld', 'sls', 'ss' }
     comment_syntax: ';'
     auto_pairs: {
       '(': ')'


nilnor:
While I wouldn't mind adding these extra extensions, I'm interested to hear whether we could instead use the newer `lisp` mode instead? That's a newer, custom, Howl mode for lisp languages that I've written some time ago, which should have better indentation support, etc. Do you think you could try that out for scheme and see whether that works as well or better? Simply uncomment the entire scheme definitions block from the basic modes and add the extensions to the lisp mode instead (bundles/lisp).

stutonk:
Sure; I'll give it a shot and let you know how it goes.

stutonk:
After using it for a few days I don't notice any problems with editing Scheme code in Lisp mode. Indentation is MUCH better than having to fiddle manually. If it's not 100% accurate to the canonical indentation algorithm, it definitely appears to be close enough. The lexer it uses seems to get all the major stuff for Scheme as well. I'd recommend your proposed change over the one I proposed here.

nilnor:
Thanks for the feedback! I've now switched things over in master.

Indentation for lisp is hard (more likely impossible) to get right for all languages. I wrote this mode with Clojure in mind originally, and as I recall the canonical indentation there was dependent not only on syntax but differed depending on what exactly the first form element was. Anyway, to make this better it would be possible to define a custom `scheme` mode that derives from the lisp mode. Indentation support could then be tailored exactly towards scheme.

PR #409, opened 2018-01-10
Basic support for ES6 template literals.

JSONVoorhees:
Prevents syntax highlighting from breaking when using ES6 template literals.


nilnor:
Thanks! Together with this, could you also update the `example.js` file in the bundle, to include various examples of the ES6 template literals? The idea is to have the example files include examples of all covered syntax.

JSONVoorhees:
No problem, documented. I included usage of the string ${interpolation} operator, even though that does not have syntax highlighting support yet.

nilnor:
:+1:

PR #417, opened 2018-03-06
Add missing javascript keywords introduced in ES2015+

Delapouite:
Hi.

Here's a refreshed list of js keywords with recent additions: `async`, `await`, `class`…


nilnor:
Thanks! Could you also update the `misc/example.js` file with example of these key words?

Delapouite:
Sure. I've updated the pull request

nilnor:
:+1:

PR #420, opened 2018-03-14
Go bundle upgrade

rokf:
Added two inspectors:
- golint
- gotoolvet

Added a `go-doc` command. It uses `gogetdoc` to fetch documentation.

- https://github.com/golang/lint
- https://github.com/zmb3/gogetdoc

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased (in master)
 
+- Added two inspectors and a command for documentation to the Go mode
+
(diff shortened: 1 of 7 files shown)


nilnor:
I merged your changes, but then also took the time to do some modifications, such as 

- removed the `go-doc` command and incorporated that into the show-doc-at-cursor` command
- check for the existence of needed commands 
- make the path to gogetdoc configurable

It would be great if you could try everything out and make sure it still works the way it should!

rokf:
Still works, thanks! :+1:

PR #427, opened 2018-04-11
project-switch-buffer command

shalabhc:
Adds new command `project-switch-buffer` which shows the list of open buffers within the current process only and allows you to select one to open.

### DIFF
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -6,6 +6,21 @@ import ActionBuffer, ProcessBuffer, BufferPopup, StyledText from howl.ui
 import Process from howl.io
 serpent = require 'serpent'
 
+get_project_root = ->
+  buffer = app.editor and app.editor.buffer
+  file = buffer.file or buffer.directory
+  error "No file associated with the current view" unless file
+  project = Project.get_for_file file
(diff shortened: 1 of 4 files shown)


lib/howl/commands/app_commands.moon
```diff
@@ -6,6 +6,21 @@ import ActionBuffer, ProcessBuffer, BufferPopup, StyledText from howl.ui
...
+  file = buffer.file or buffer.directory
+  error "No file associated with the current view" unless file
+  project = Project.get_for_file file
+  error "No project associated with #{file}" unless project
+  return project.root
+
+belongs_to_project = (buffer, project_root) ->
+  file = buffer.file or buffer.directory
+  return false unless file
+  project = Project.for_file file
```
nilnor: Are you aware of `File.is_below(dir)`? Would shorten this somewhat.

shalabhc: Yes I did think about it, then decided to depend on `Project.for_file`. I thought what if the logic for determining which files belong in a project changes, e.g. excludes certain files? We want to ensure that both files belong to the same project.

shalabhc:
See comment about using `Project.for_file`. I can still switch to using `File.is_below` if you want.

PR #435, opened 2018-04-24
Fix scrollbar getting stuck right before scrolling up to line 1

kirbyfan64:


kirbyfan64:
@tastyminerals Can you try this and see if it fixes the issue?

tastyminerals:
Works fine now. Thank you.

PR #441, opened 2018-05-21
Rework logging to be signal-based and add a log buffer

kirbyfan64:
This does two key things:

- It reworks logging to be based on signals and subscriptions. Anyone can now subscribe to log output. This can be handy for things like bundles to be able to listen to log events and send them to custom receivers (e.g. journald). Or, of course, it can be useful so we don't have to keep on making `log.dispatch` a huge function that sends events to a bunch of different places with a bunch of preconditions (a problem I ran into while adding the log buffer).
- Adds a log buffer that can be opened via `open-journal`. I named it the journal because *log* tends be used as a verb, which wouldn't fit with this noun-based usage.

I actually found two minor but interesting bugs in Howl while looking at the journal buffer, which I'll be fixing soon. :man_shrugging:

### DIFF
--- a/lib/howl/application.moon
+++ b/lib/howl/application.moon
@@ -102,6 +102,21 @@ class Application extends PropertyObject
       on_error: _G.log.error
     }
 
+    signal.connect 'log-entry-appended', (entry) ->
+      level = entry.level
+      message = entry.message
+      return if level == 'traceback'
+
+      if @window and @window.visible
+        essentials = entry.essentials
+
+        status = @window.status
+        command_line = @window.command_line
+        status[level] status, essentials
+        command_line\notify essentials, level if command_line.showing
+      elseif not @args.spec
+        print message
+
     super!
 
   @property idle: get: =>
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -2,7 +2,7 @@
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 import app, activities, breadcrumbs, Buffer, command, config, bindings, bundle, interact, signal, mode, Project from howl
-import ActionBuffer, ProcessBuffer, BufferPopup, StyledText from howl.ui
+import ActionBuffer, JournalBuffer, ProcessBuffer, BufferPopup, StyledText from howl.ui
 import Process from howl.io
 serpent = require 'serpent'
 
@@ -371,6 +371,13 @@ command.register
   handler: (loc) ->
     return unless loc
     breadcrumbs.location = loc[1]
+
+command.register
+  name: 'open-journal'
+  description: 'Opens the Howl log journal'
+  handler: ->
+    app\add_buffer JournalBuffer!
+    app.editor.cursor\eof!
 
 -----------------------------------------------------------------------
 -- Howl eval commands
--- a/lib/howl/log.moon
+++ b/lib/howl/log.moon
@@ -3,7 +3,7 @@
 
 _G = _G
 import table from _G
-import config from howl
+import config, signal from howl
 
 append = table.insert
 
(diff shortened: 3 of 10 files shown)


kirbyfan64:
Ah right, I also fixed a bug in the MoonScript support library that was causing tracebacks to never actually be rewritten.

kirbyfan64:
@nilnor Everything you mentioned was fixed...*except* for the subscription functions. There's actually a bit of a rationale for that:

- `log.subscribe` also replays any passed log entries. Coincidentally enough, this is what made it easy to have `JournalBuffer` constructed whenever wanted; when it subscribes to logging events on construction, any previous ones are also replayed. (Of course, they can be ignored by returning if `entry.replay` is `true`.)
- Being to able to register/unregister the append and trim functions at once makes it a bit easier to use, instead of having to store both separately and unregister separately.

nilnor:
@kirbyfan64 Yes I noticed the replay functionality, but that to me is not a good reason to introduce some new publish/subscribe thing. The entries are already available from the log module (`log.entries`) and can easily be fetched from there when constructing the log buffer.

kirbyfan64:
@nilnor Ok, ok, you win... It's all ready now!

I also added a little check to try and make sure an infinite recursion doesn't occur if a signal call fails (which has happened to me before).

lib/howl/application.moon
```diff
@@ -74,6 +74,22 @@ class Application extends PropertyObject
       on_error: _G.log.error
     }
 
+    signal.connect 'log-entry-appended', (entry) ->
+      level = entry.level
+      message = entry.message
+      return if level == 'traceback'
+
+      if @window and @window.visible
+        return if entry.replay
```
nilnor: Left-over

lib/howl/commands/app_commands.moon
```diff
@@ -326,6 +326,14 @@ command.register
     else
       log.info "No next location recorded"
 
+command.register
+  name: 'open-journal'
+  description: 'Opens the Howl log journal'
+  handler: ->
+    breadcrumbs.drop!
```
nilnor: `drop` not needed here as `add_buffer` takes care of it.

lib/howl/log.moon
```diff
@@ -13,9 +13,29 @@ config.define
...
+  description: 'Called when the log is trimmed to the given size'
+  parameters:
+    size: 'The new number of entries in the log'
+
 log = {}
 setfenv 1, log
 
+safe_emit = (name, options) ->
+  parent = _G.debug.getinfo 4
+  if parent and parent.name == 'safe_emit' and parent.source\match 'signal.moon$'
```
nilnor: This seems cumbersome and frail - I suggest just using a local state variable defined above instead, e.g. `emitting`.

lib/howl/ui/journal_buffer.moon
```diff
@@ -0,0 +1,51 @@
...
+
+class JournalBuffer extends ActionBuffer
+  new: =>
+    super!
+
+    @read_only = true
+    @title = 'Howl Journal'
+    @entry_sizes = {}
+
+    signal.connect 'log-entry-appended', (entry) -> @append_entry entry
```
nilnor: There's a garbage collecting issue here, as the signal handlers cause the journal buffers to be pegged. I.e. closed journal buffers are never really removed, and continue to silently process signals invisibly. I'd suggest keeping track of open journal buffers in this file using a weak table, and have on set of signal handlers dispatching to all available buffers.

kirbyfan64: Would it be easier to just save the journal buffer and re-use it when opened again, or just flat-out completely destroy it and disconnect the signals? I can't really imagine a use case where someone would need more than one active at once...

nilnor: Yes, these options would works as well (but if the buffer were to be re-used then signal would need to be disconnected between views as well). As long as there's no penalty when not viewing the journal buffer. I agree that having more than one active doesn't make sense.

kirbyfan64:
@nilnor All fixed and rebased onto master w/ merge conflicts fixed.

nilnor:
All right, looks good. There are some issues with breadcrumbs and the journal buffer, but I can have a look at that in master.

PR #443, opened 2018-05-22
Add an examples.rs file and fix a few bugs in the Rust lexer

### LINKED ISSUE #442: Character literals are recognized as lifetime specifiers by the Rust bundle
The Rust bundle incorrectly highlights character literals as lifetime specifiers, as is shown in line 4 below:

2018-05-22-152552_224x96_scrot

Inverting the precedence of the lexer rules that match the two tokens should be enough to resolve this issue.

Leonardo2718:
This series of commits adds an `example.rs` file and fixes the following in the Rust lexer:

- lifetime specifiers are given lower precedence than character literals
- `type` keyword is given lower precedence than `type` keyword
- rules for capturing numeric literals are re-implemented to better conform to Rust syntax rules
- support for Unicode character literals is added

Closes #442 

2018-05-22-155454_226x104_scrot

### DIFF
--- a/bundles/rust/misc/example.rs
+++ b/bundles/rust/misc/example.rs
@@ -1,0 +1,111 @@
+#![allow(unused)]
+
+extern crate
+
+// an old fashion single line comment
+
+/*
+a multi-line
+comment
+*/
+
+/*
+a
+/*
+nested
+*/
+multiline
+comment
+*/
+
+as
+const
+crate
+extern
+for
+impl
+in
+move
+mut
+ref
+return
+Self
+static
+super
+trait
+unsafe
+where
+while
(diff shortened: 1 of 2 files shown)


Leonardo2718:
Heh, that's likely the reason the issue got missed in the first place, so adding an examples.rs file is probably a good idea :grin: .

I'll see about creating (or at least starting) one.

nilnor:
That would be awesome @Leonardo2718! It slipped by at the time of the first merge, but we really need to get one going.

shalabhc:
Rerunning CI fixed it.

nilnor:
Great, thanks!

PR #450, opened 2018-06-08
Make sure editor-move-{lines-down},{text-right} work properly at EOF

### LINKED ISSUE #424: 'editor-move-lines-up' doesn't move cursor with line if moving last line
When selecting the last line of a file, and using the `editor-move-lines-up` to move the line upwards a level, the cursor does not move with the line and instead produces a new line at the end of the file.

deepin-screen-recorder_select area_20180318155322

kirbyfan64:
Fixes #424, as well as another issue I found with `editor-move-text-right` at the second-to-last character.

### DIFF
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -195,7 +195,8 @@ command.register
     buffer\as_one_undo ->
       editor\with_selection_preserved ->
         buffer.lines\delete nr, nr
-        buffer.lines\insert last, text
+        editor\with_position_restored ->
+          buffer.lines\insert last, text
 
 command.register
   name: 'editor-move-lines-down'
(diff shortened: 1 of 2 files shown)


nilnor:
Nice, thanks! However, whenever bugs are found and it's at all possible them I think we need to add specs for the incorrect behaviour. This seems like a very good candidate for this.

kirbyfan64:
Are there any specs for commands right now though? I went to add some, but none of them seemed to be tested... (Unless I missed something obvious, which I *have* done before!)

nilnor:
True, it's a command. No, you're right, there aren't any specs for commands yet, but I think we should start adding some since commands can be quite complex in their own right. You want to have a stab? In this case it should be pretty straightforward (add a edit_commands_spec file, test the command via the command module).

kirbyfan64:
@nilnor Done!

nilnor:
Great! :+1:

PR #454, opened 2018-07-05
Add some background image enhancements

kirbyfan64:
You can now set an alpha value for a background image. In addition, if you specify a color *and* an image, the color will be used as a background for the image, meaning that you can set an alpha value for the image and have it blended with the color.


nilnor:
:+1:

PR #456, opened 2018-07-20
Flatpak!

kirbyfan64:
There are some niceties that come along with this:

- New user wants to try out new changes? They can just use `flatpak-builder` to build and run a Flatpak without having to install dependencies and such.
- This would let us put Howl on Flathub, which would make installation rather nice and *incredibly* easy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.flatpak-builder/
 src/deps/*/*
 src/*.o
 src/howl
--- a/io.howl.Editor.yaml
+++ b/io.howl.Editor.yaml
@@ -1,0 +1,69 @@
+app-id: io.howl.Editor
+runtime: org.gnome.Platform
+runtime-version: 3.28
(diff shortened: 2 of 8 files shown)


MarcusE1W:
Flatpack has just been submitted as [1.0](https://flatpak.org/press/2018-08-20-flatpak-1.0/)
 
Would it be a good timing to complete this PR ? :smiley:

kirbyfan64:
Just as an FYI I've been using this as my main Howl installation for a while now (especially after switching to Silverblue), and, as far as I can tell, it works perfectly in real-world use.

nilnor:
Looks good! Two minor things:

- I'd rather not the yaml in the root, is possible to move that to `src` instead?
- The `sys.info.flatpack` variable I think could be renamed to `sys.info.is_flatback` to better reflect its boolean nature

kirbyfan64:
@nilnor Done!

PR #464, opened 2018-08-20
[WIP] @feature: "cargo check"-inspector for the rust bundle

bugabinga:


bugabinga:
add "work in progress" label, because I have encountered a bug that I will fix.
@nilnor Is this feature of interest to you?

bugabinga:
I still try to fix this from time to time, but regularly hit a wall.

There is a bug blocking this:
- inspections get appended to the popup list on every `buffer-inspect`

Hopefully you will find some time to whip up some docs about how bundles are supposed to get built ;).

zesterer:
Has these been any progress on this? Seems like a very useful feature. I've not used Moon before, but I'm happy to try my hand at fixing the remaining issue if explained a little further.

PR #467, opened 2018-09-17
Fix running external commands from Flatpak Howl

kirbyfan64:
Basically, an app inside a Flatpak doesn't have full access to the host environment, meaning just trying to read `PATH` won't work.

This leaves out `lib/howl/interactions/external_command.moon`'s autocomplete capabilities, which I'm still trying to figure out how to cleanly fix...

### DIFF
--- a/lib/howl/sys.moon
+++ b/lib/howl/sys.moon
@@ -2,7 +2,9 @@
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 glib = require 'ljglibs.glib'
-{:File} = howl.io
+{:File, :Process} = howl.io
+
+is_flatpak = File('/.flatpak-info').exists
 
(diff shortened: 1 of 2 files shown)


lib/howl/sys.moon
```diff
@@ -20,22 +22,27 @@ env = setmetatable {}, {
 find_executable = (name) ->
   return File(name).exists if File.is_absolute(name)
 
-  path = env['PATH']
+  if is_flatpak
+    stdout, stderr = Process.execute {'which', name}
```
nilnor: `execute` returns `stdout`, `stderr` and the process object. Use the latter and `.successful` instead of relying on the `stderr` to be available.

kirbyfan64:
@nilnor Done.

PR #473, opened 2018-10-18
Allow disabling titlebar hiding when maximized

### LINKED ISSUE #419: Provide a way of re-enabling the titlebar
Personally, I don't use a tiling window manager. However, I remember the keybindings I need to manipulate a window; but many users do not. The option to re-enable the title bar would be useful for new or inexperienced users. This setting would preferably not be hidden deep in the settings (although it's useful to provide a hook for the init script to use). Perhaps a tiny icon in the corner of the window?

kirbyfan64:
Closes #419.

### DIFF
--- a/lib/howl/ui/window.moon
+++ b/lib/howl/ui/window.moon
@@ -8,7 +8,7 @@ gobject_signal = require 'ljglibs.gobject.signal'
 Background = require 'ljglibs.aux.background'
 {:PropertyObject} = howl.util.moon
 {:Activity, :CommandLine, :Status, :theme} = howl.ui
-{:signal} = howl
+{:config, :signal} = howl
 
 append = table.insert
 ffi_cast = ffi.cast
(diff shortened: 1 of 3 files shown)


lib/howl/ui/window.moon
```diff
@@ -24,6 +24,8 @@ placements = {
   below: 'POS_BOTTOM'
 }
 
+GTK_SUPPORTS_HIDDEN_TITLEBAR = Gtk.get_major_version! >= 3 and Gtk.get_minor_version! >= 4
```
nilnor: I guess that whether [gtk_check_version](https://developer.gnome.org/gtk3/stable/gtk3-Feature-Test-Macros.html#gtk-check-version) would be more appropriate here, as this won't work with say Gtk 4.0. You want to implement that?

kirbyfan64:
@nilnor Done!

PR #485, opened 2018-12-12
Update the C lexer to parse raw strings

shalabhc:


shalabhc:
Yes, do you know how to do that in Howl lexers?

kirbyfan64:
Don't have time to look into it more ATM (probably will next week), but it would be like [it's done for Lua](https://github.com/howl-editor/howl/blob/master/bundles/lua/lua_lexer.moon#L14). Something like:

```moonscript
c('special', 'R') * c 'string', sequence {
  P'"'  -- opening quote
  Cg((P 1 - '(')^0, 'tag')  -- all characters until opening paren (technically can't be backslash or space either, but might as well be more liberal in the lexer)
  P'('  -- left paren
  scan_to(P')' * match_back'tag' * '"')^-1  -- include anything until )tag"
}
```

(Side note: the `R` should probably be *special* like it is for Python and Nim.)

shalabhc:
Thanks - `match_back` is what I was looking for.

bundles/c/misc/example.c
```diff
@@ -1,6 +1,6 @@
 /*
 ** Syntax examples taken mostly from http://en.wikipedia.org/wiki/C_syntax
-** My library.
+match_back('delim') c('string', ')"')** My library.
```
nilnor: Inadvertent modification it seems

PR #489, opened 2019-01-05
JavaScript bundle updates

kirbyfan64:
Yeah, I'm cutting it a bit close to release...but IMO this is really quite an improvement. Changes:

- Proper template literal lexing.
- Class name lexing.
- Lex `this.xyz` as a member.
- Added more regex operators.
- Add shebangs and some more extensions.


bundles/javascript/init.moon
```diff
@@ -1,6 +1,8 @@
 mode_reg =
   name: 'javascript'
-  extensions: { 'js', 'jsfl' }
+  extensions: { 'js', 'jsfl', 'jsm', 'es' }
+  -- shebangs: '[/ ]node|nodejs|gjs$'
```
nilnor: Left-over

Issue #495, opened 2019-02-25
Command line option --reuse doesn't work with relative file paths


dgaw:
The ``--reuse`` command line option doesn't work correctly with relative file paths in cases where the current working directory has changed since Howl was launched.

Steps to reproduce:
```
cd howl
howl README.md
```
... and then in another terminal:
```
cd howl/src
howl --reuse main.c
```

The last command creates a new empty buffer for ``main.c`` instead of opening the file correctly.

I'll send a PR with a fix for this.

dgaw:
Thanks for merging @nilnor!

PR #496, opened 2019-02-25
Fix for --reuse with relative file paths

dgaw:
A fix for #495

### DIFF
--- a/lib/ljglibs/cdefs/gio.moon
+++ b/lib/ljglibs/cdefs/gio.moon
@@ -280,6 +280,8 @@ ffi.cdef [[
     GApplicationCommandLine *cmdline,
     int *argc);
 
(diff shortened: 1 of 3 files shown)


lib/ljglibs/cdefs/gio.moon
```diff
@@ -280,6 +280,8 @@ ffi.cdef [[
     GApplicationCommandLine *cmdline,
     int *argc);
 
+  gchar * g_application_command_line_get_cwd (GApplicationCommandLine *cmdline);
```
nilnor: Missing a `const` here.

dgaw: Thanks for reviewing. I've just added the missing const.

nilnor:
👍

PR #502, opened 2019-04-18
The Appdata files now go to /usr/share/metainfo

tim77:
From
 `/usr/share/appdata/%{id}.appdata.xml`
to
 `/usr/share/metainfo/%{id}.appdata.xml`

https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps


refi64:
I'm :+1: to the first commit, but AFAIK the second not using the direct paths is intentional; Howl isn't solely targeting Linux, and these example scripts are just for illustration anyway.

nilnor:
I'm with @refi64 here, first commit is fine, second one should be dropped.

tim77:
Ok, the second one is not really problem since it possible to fix this in .spec file and i already fixed this for Fedora build. :)

nilnor:
@tim77 sorry for the delay - I cherry-picked the first commit into master now, thanks!

PR #509, opened 2019-06-07
[WIp] Automated nightly Howl Flatpak builds

refi64:
Basically, it is what it says. You can try out the repo:

```bash
# Add the remote
flatpak remote-add --user --no-gpg-verify howl-nightly https://storage.googleapis.com/howl-dl/repo
# List the refs inside
flatpak remote-ls howl-nightly
# Install Howl nightly
flatpak install howl-nightly io.howl.Editor
```

### Why is this [WIP]?

There are four more things that probably need to be done:

- Right now the cloud bucket bucket name is just `howl-dl`. If we want to use a shorter URL, e.g. off the howl.io domain (something like `dl.howl.io`), it needs to be changed here to match.
- Dropping `--no-gpg-verify` is generally a good thing, but first the commits need to be GPG-signed. In general using a new, separate key from anyone's normal one is preferred, but I wasn't sure if @nilnor would prefer it point to e.g. his email or mine or something else, so I didn't create one yet.
- Related to the above, we can create a `.flatpakrepo` file that a user can double-click to automatically add the remote, but it depends on GPG signatures being enabled, so see the above point.
- Make artifacts public (see below).

In addition to the scheduled "nightly" builds (I haven't actually enabled the trigger yet), Azure Pipelines will build a Flatpak bundle for every PR or commit. So, if someone wants to try something out, they can just download the Flatpak bundle and install it to give that PR / commit / whatever a spin. You can see the current bundles [here](https://dev.azure.com/howl-editor/howl/_build), just select a build, then go to Artifacts -> Flatpak bundle -> download howl.flatpak. These artifacts need to be published in a slightly different way to make them public but I haven't done that yet.

So yeah, initial feedback welcome. :grin:

**EDIT:** Also the commit history is a mess, please squash when merging.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,19 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- master
+
+pool:
+  vmImage: 'Ubuntu-16.04'
+
+steps:
+- script: echo Hello, world!
+  displayName: 'Run a one-line script'
+
+- script: |
+    echo Add other tasks to build, test, and deploy your project.
+    echo See https://aka.ms/yaml
+  displayName: 'Run a multi-line script'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,21 +1,17 @@
-# Starter pipeline
-# Start with a minimal pipeline that you can customize to build and deploy your code.
-# Add steps that build, run tests, deploy, and more:
-# https://aka.ms/yaml
-
 trigger:
 - master
 
 pool:
   vmImage: 'Ubuntu-16.04'
 
-container: 'fedora:29'
+container: 'registry.gitlab.gnome.org/gnome/gnome-runtime-images/gnome:3.30'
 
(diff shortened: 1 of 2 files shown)


nilnor:
Thanks for this, and apologies that it sat here for so long without a verdict.

Master's build and CI setup has since changed in ways that make this unmergeable:

- Travis was replaced by **GitHub Actions** (`.github/workflows/ci.yml`), and CI now runs on master. Adding a second, third-party CI system (Azure Pipelines) isn't something I want to take on.
- The build moved to **Gtk 4** and LuaJIT was bumped from `2.1.0-beta3` to `luajit-2.1-2025-04-09` (now hosted on `download.howl.io`). `src/io.howl.Editor.yaml` was updated accordingly, so the manifest changes here conflict.
- The open questions in the WIP list (bucket naming, GPG signing keys, `.flatpakrepo`, artifact visibility) were never resolved and depend on infrastructure decisions that have since moved on.

Closing as obsolete. The Flatpak manifest itself is still maintained in-tree; if nightly Flatpak builds are ever revived it would be as a GitHub Actions job against current master. Also worth flagging: the README now states that master is unreleased and narrows contribution channels, with no promises about reviewing or merging PRs — so I don't want to leave this hanging open indefinitely as if it were queued.

PR #514, opened 2019-08-03
Add an INI mode

refi64:
May I present to you, the ultimate nightmare lexer.

Basically, we've got four very, very different syntaxes to support, but they have enough common ground that leaving them in fully different modes results in a ton of ugly duplication.

I was originally going to play around with an okay-ish editorconfig bundle, but then I realized the file didn't look pretty to edit and got a bit sidetracked...


nilnor:
I think it might actually be better to accept some duplication over having this many conditionals inside the lexer. Whatever definitions can be reused could placed in a shared lexer module. Having everything in one lexer makes it very hard to reason about.

refi64:
Hmm I think I might have a nice way to avoid *too* much reuse, will start toying with it soon.

PR #515, opened 2019-08-03
Rust syntax highlighting tweaks

refi64:
Been using for a bit now, ironically enough I ended up not using Rust for the project I added this for, but hey it's syntax highlighting improvements so ¯\\\_(ツ)\_/¯.


nilnor:
Add some corresponding examples in the example file as well, then it's all good.

refi64:
So I just realized that the stuff I changed *already* had examples in example.rs, but also had examples for stuff two minor tweaks I didn't fix, so I just did that too. :sweat_smile:

PR #524, opened 2019-08-25
Make File.with_tmpfile() return the callback result.

Janfel:
`File.with_tmpfile()` now returns the return values of the passed callback, if successful, instead of an unhelpful `nil`.

### DIFF
--- a/lib/howl/io/file.moon
+++ b/lib/howl/io/file.moon
@@ -39,9 +39,12 @@ class File extends PropertyObject
 
   with_tmpfile: (f) ->
     file = File.tmpfile!
-    status, err = pcall f, file
+    result = { pcall f, file }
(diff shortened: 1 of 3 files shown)


nilnor:
Seems like a good change as intended. The logic is actually inverted though, which probably explains all the spec errors from travis since this now raises an error for every successful call. You can read a little bit about how to run the specs [here](https://howl.io/doc/manual/dev-howl.html)

nilnor:
Oh, and if we could add a short spec for the new behaviour as well then :+1: .

Janfel:
I feel so stupid right now... thanks for writing all those specs.
I have fixed it and added a spec for the new behaviour.

nilnor:
No worries. Looks good now, thanks!

PR #536, opened 2019-11-03
ctrl + [ support for exiting insert mode

drudgesentinel:
Added support for 'ctrl + [' to exit insert mode in the vi bundle.

Just duplicated the 'escape' portion of the insert_map.moon and swapped in the 'ctrl + ['

### DIFF
--- a/bundles/vi/insert_map.moon
+++ b/bundles/vi/insert_map.moon
@@ -33,6 +33,12 @@ insert_map = {
     state.change_mode editor, 'command'
     editor.cursor.column = math.max 1, editor.cursor.column - 1
 
+  'ctrl_[': (editor) ->
+    state.insert_edit = get_edit editor
+    insert_pos = nil
+    state.change_mode editor, 'command'
+    editor.cursor.column = math.max 1, editor.cursor.column - 1
+
   ctrl_i: (editor) -> editor\shift_right!
   ctrl_d: (editor) -> editor\shift_left!
 }
--- a/bundles/vi/insert_map.moon
+++ b/bundles/vi/insert_map.moon
@@ -33,8 +33,8 @@ insert_map = {
       blink_interval: -> config.cursor_blink_interval
   }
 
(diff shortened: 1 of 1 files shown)


bundles/vi/insert_map.moon
```diff
@@ -33,6 +33,12 @@ insert_map = {
     state.change_mode editor, 'command'
     editor.cursor.column = math.max 1, editor.cursor.column - 1
 
+  'ctrl_[': (editor) ->
+    state.insert_edit = get_edit editor
```
shalabhc: Can you please extract the code into a `exit_insert_mode` function at the top next to `get_edit`?

ofperfection: I've refactored the code per your request. Let me know if there's anything else I can do!

bundles/vi/insert_map.moon
```diff
@@ -27,22 +33,11 @@ insert_map = {
...
-    state.change_mode editor, 'command'
-    editor.cursor.column = math.max 1, editor.cursor.column - 1
-
-  'ctrl_[': (editor) ->
-    state.insert_edit = get_edit editor
-    insert_pos = nil
-    state.change_mode editor, 'command'
-    editor.cursor.column = math.max 1, editor.cursor.column - 1
-
+  escape: (editor) -> exit_insert_mode editor
```
shalabhc: I think this can just be written as

`escape: exit_insert_mode`

Same for next line.

Otherwise looks good to me!

ofperfection: Alright, I've done so.

ofperfection: I've found that the ctrl + [ doesn't work when writing python code at least, there's probably something overriding the key combo for that syntax, I'll investigate further at some point.

shalabhc: Do you still want to merge this then?

PR #540, opened 2019-11-14
Don't close menu_popup when just a modifier key is pressed.

shalabhc:
Also add a list of modifiers to bindings.moon.

### DIFF
--- a/lib/howl/ui/menu_popup.moon
+++ b/lib/howl/ui/menu_popup.moon
@@ -2,7 +2,7 @@
 -- License: MIT (see LICENSE.md at the top-level directory of the distribution)
 
 {:List, :ListWidget, :Popup} = howl.ui
-{:config} = howl
+{:bindings, :config} = howl
 
 class MenuPopup extends Popup
   new: (@items, @callback) =>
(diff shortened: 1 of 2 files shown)


lib/howl/ui/menu_popup.moon
```diff
@@ -46,6 +46,9 @@ class MenuPopup extends Popup
     page_up: => @list\prev_page!
 
     on_unhandled: (event, source, translations, self) ->
+      -- if a bare modifier such as just 'ctrl', don't close popup
+      return if bindings.is_modifier translations
+      -- any other not character such as home or escape closes the popup
```
refi64: nit: would `non-character` be better here?

shalabhc: `escape` is a non-character but that should close the popup (also `home`, `end`). Also we already have `.character` which is kind of the inverse of non-character. We really want to identify the modifiers like ctrl, shift, alt only.

shalabhc:
@Janfel does this fix your iso shift modifier issue?

Janfel:
Yes, it works perfectly with this fix.

PR #544, opened 2019-11-16
feat: add '%' key for vi (goto_matching_brace)

Delapouite:
Hi

This PR adds a basic support for the `%` in vi. (note: the behavior is slightly different, but I think it's already quite handy and good enough for a first iteration).

As the 3 lines logic was declared inline in the `cursor-goto-brace` command registration, I had to move this bit of code to a proper function `goto_matching_brace` on the cursor class.

I've completed the spec / api doc accordingly.

### DIFF
--- a/bundles/vi/base_map.moon
+++ b/bundles/vi/base_map.moon
@@ -146,6 +146,9 @@ map = {
 
     '}': (editor) -> apply editor, (editor) ->
       editor.cursor\para_down!
+
+    '%': (editor) -> apply editor, (editor) ->
+      editor.cursor\goto_matching_brace!
    }
 
   on_unhandled: (event, source, translations) ->
--- a/lib/howl/commands/edit_commands.moon
+++ b/lib/howl/commands/edit_commands.moon
@@ -199,9 +199,7 @@ command.register
   name: 'cursor-goto-brace'
   description: 'Go to the brace matching the current brace, if any'
   handler: ->
-    cursor = app.editor.cursor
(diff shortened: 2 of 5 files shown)


nilnor:
Merged — with apologies for the six year wait. Master sat untouched for years while development happened on the `next` branch, and this fell through the gap. It shouldn't have: of the PRs still open from that era, this is the only one that arrived with both a spec and an API doc update, and it went in essentially as written.

Your commit is preserved with you as author. Two small changes on merge, noted below.

> [!NOTE]
> **The verification below was produced by AI, run and reviewed by me.**

---

Cherry-picked onto current master. Rename detection resolved the one stale path automatically — `site/source/doc/api/ui/cursor.md` became `cursor.html.md` when the site moved to a current middleman, and the hunk applied there untouched.

Amendments:

- `goto_matching_brace: () =>` → `goto_matching_brace: =>`. The empty parens were the only occurrence of that form in `lib/howl/`; `cursor.moon` alone has ten no-arg methods written without them.
- The API doc hunk landed in `cursor.html.md` per the above.

Checks against master:

```
bin/howl-lint lib/howl/ui/cursor.moon   clean
spec/ui/cursor_spec.moon                80 passed, 0 failed
bundles/vi/spec/vi_spec.moon            49 passed, 0 failed
```

The divergence from vi's `%` that you flagged in the description — jumping only when the cursor is already on a brace, rather than finding the nearest enclosing pair on the line — is recorded in the commit message rather than papered over, so it stays visible to whoever extends it next.

PR #552, opened 2020-02-23
Fix for buffer search matches being obscured

### LINKED ISSUE #551: Obscured search match highlight
When searching with `ctrl_f` (`buffer-search-forward`) and navigating between matches using the up/down arrow keys the last match in the file sometimes gets obscured.

I've attached a screenshot to illustrate this.

image

I have a fix for this issue and will send a PR shortly.

dgaw:
Fixes #551


shalabhc:
Thanks for fixing!

PR #555, opened 2020-03-05
Adding Zig (0.5.0+378bf1c3b) support.

AequoreaVictoria:


shalabhc:
I haven't used Zig but would be nice to have. Any opinions, @refi64?

Delapouite:
Including zig support will be a good move.
This language still has a small community. That's also the case for the Howl community.
I believe open source development is about helping each other, by offering better visibility and spreading the love. So it's a win win situation.

AequoreaVictoria:
Pushed a final update to add decent support for Zig's arbitrary precision integers, where types could range from 'i0' to 'i65535', as well as for unsigned. Also expanded highlighting for constants, types and libraries to cover CamelCase, CAPS and Snake_Case.

At this point I consider the lexer finished, as I've used it to write a moderately sized codebase in Zig without issue other than the two cases I've just addressed.

bundles/zig/zig_lexer.moon
```diff
@@ -0,0 +1,138 @@
...
+  important = upper^1 * (upper + lower + digit + '_')^0
+  important_name = c 'constant', important
+
+  -- Identifiers
+  identifier = c 'identifier', ident
+
+  -- Operators
+  operator = c 'operator', S'+=%-*/<>&|^~?!:;,.()[]{}'
+
+  P {
```
nilnor: Just a FYI: since there's no actual grammar being used (the grammar table contains only the one 'all' entry) it would be sufficient to return just the `any {` block directly and not use `P` at all.

nilnor:
Looks good to me, solid PR :+1:

bundles/zig/zig_lexer.moon
AequoreaVictoria: Updated as suggested. Thanks!

bundles/zig/zig_lexer.moon
```diff
@@ -0,0 +1,133 @@
...
+    'enum', 'errdefer', 'error', 'export', 'extern', 'fn', 'for', 'if',
+    'inline', 'noalias', 'orelse', 'or', 'packed', 'promise', 'pub', 'resume',
+    'return', 'linksection', 'struct', 'suspend', 'switch', 'test',
+    'threadlocal', 'try', 'union', 'unreachable', 'usingnamespace', 'var',
+    'volatile', 'while'
+  }
+
+  -- Special words
+  special = c 'special', word {
+      'true', 'false', 'null', 'undefined', '@addWithOverflow', '@alignCast',
```
refi64: I'm not intimately familiar with Zig, but if all these start with an `@` sign, then I'm guessing that signifies some sort of compiler-level built-in. Could it then just match any word after an `@` as special? Might be also worth handling a plain `@` then too so it looks prettier as you type the code.

(As a side note, builtin *functions* are usually lexed using the `function` color, and *compiler directives* are usually `operator` [I think...]. I'm not sure where these fall exactly, but it may be worth taking a look at those two color options as well.)

AequoreaVictoria: `@` is supposed to be used for builtin functions. The language documentation's own highlighting has it distinct as its own color and its been tokenized as 'builtin'. I'm open to any suggestion how else it should be tagged, but I do think they should remain distinctly colored the way they are in the docs. 

As to your question about matching any word, I wouldn't do that in the general case as it quickly offers feedback right now when you've got the builtin correct. I think I would miss not having that as a feature. I constructed that list with a shell one-liner scraping the docs page thanks to the CSS tags including the token type. I don't imagine they'd have to be updated frequently, either.

However, one thing unsupported but related is the `@""` syntax for treating the quoted string as a string token, rather than a keyword or random bytes, with  `foo.@“bar”` being identical to foo.bar. This allows for you to do things like `@"😁"` as in `pub fn @"😁"(self: @This()) void { }` or `export fn @"A function name that is a complete sentence."() void {}`. There's no rule for matching that as anything other a bare `@` followed by a string. That seemed fine enough.

refi64: Ah okay, so in that case you should probably use the `function` color for these, which is what builtins generally use in other languages (see Python for an example). It's hard to compare syntax highlighting across editors / libraries / sites, because they all tend to separate the different colors differently.

AequoreaVictoria: I personally will not be doing that on my end, but if that's what you wish to do in order to merge the PR, that's fine.

refi64:
Sorry for the delay, but again thanks!

PR #557, opened 2020-03-07
Add a basic mode for Cabal files (Haskell)

### LINKED ISSUE #556: A basic mode for Cabal files (Haskell)
I've created a basic mode to provide syntax highlighting for Cabal build files. It's useful to anyone writing Haskell in Howl.

I'll send a PR shortly.

dgaw:
Closes #556

### DIFF
--- a/bundles/basic_modes/lexers/cabal.lua
+++ b/bundles/basic_modes/lexers/cabal.lua
@@ -1,0 +1,53 @@
+-- LPeg lexer for Cabal files (Haskell)
+
+local l = lexer
+local token, word_match = l.token, l.word_match
+local P, S = lpeg.P, lpeg.S
+
+local M = {_NAME = 'cabal'}
+
+-- Whitespace
+local ws = token(l.WHITESPACE, l.space^1)
+
+-- Comments
+local comment = token(l.COMMENT, '--' * l.nonnewline_esc^0)
+
+-- Strings
+local string = token(l.STRING, l.delimited_range('"', '\\'))
+
+-- Chars
+local char = token(l.STRING, l.delimited_range("'", "\\", false, false, '\n'))
+
+-- Numbers
+local number = token(l.NUMBER, l.float + l.integer)
+
+-- Sections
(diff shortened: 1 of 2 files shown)


refi64:
Thanks for the contribution! However, basic_modes are generally "legacy" lexers (pulled from LPeg I think?). All the new ones are Howl-native bundles instead.

dgaw:
Thanks for reviewing @refi64!

If this is a matter of packaging then I can create a new bundle called "haskell" and move the haskell mode along with my cabal mode from "basic_modes" into this bundle.

What do you think?

refi64:
I mean you *could*, but they have slightly different syntaxes, even though they're both built on top of LPeg, so it would have to be more of a port. For the moment, I'd suggest just moving the Cabal one over to make things easier.

dgaw:
I use Haskell often so I don't mind spending a little time porting the Cabal mode, as well as the Haskell mode to the new syntax. 

I think it would be a logical move to create a "haskell" bundle with those two modes in it since all Haskell projects use Cabal files. Also this would give us a container for future development of Haskell-related functionality.

@refi64 Thanks for reviewing my PR. It took me a while to get back to this, sorry! Do you see any problems with this?

shalabhc:
@dgaw, given the code is small, do you mind porting it to the 'new style' modes in its own bundle? E.g. `bundles/cabal`.

For a simple example you could look at bundles/haml.

dgaw:
I don't mind porting the `cabal` mode in its own bundle.

However, what do you think about creating a `haskell` bundle I [mentioned earlier](https://github.com/howl-editor/howl/pull/557#issuecomment-612810470)?

shalabhc:
Oh ya - a haskell bundle sounds even better.

If you do, please add an examples folder like some bundles have with a file containing haskell code. Doesn't have to work, just to see the syntax highlighting.

dgaw:
Sure. I'll add an examples folder when I get around to implementing this.

dgaw:
Superseded by #584

Issue #558, opened 2020-03-08
Search in directory


dgaw:
Perhaps I'm missing something but it seems there's no command in Howl to search the contents of a specific directory.

There is `project-file-search` and `project-file-search-list`, which are very useful but often I find myself wanting to narrow down the results to a specific directory within a project.

E.g. Searching for "command" with `project-file-search-list` in the howl project generates a lot of results, including specs and docs. I'm looking for core code so I'd like to narrow down my search to only the "lib" folder. At the moment I use `buffer-grep` within the results buffer but it feels odd and doesn't allow me to switch between the narrowed-down results and an edit buffer.

Is there a way to do this with stock Howl or would it require some custom code?

shalabhc:
Yeah I don't think there's a command to search a specific directory. I usually run `exec` (`ctrl_shift_r`), cd to the directory and then run `ag`. The result buffer lines are linked to the search result - you can press enter with your cursor on a line and it jumps to the buffer.

Still, it seems like having a folder-search command would be useful. Possibly it could reuse parts of the `project-file-search-list` command.

dgaw:
Thanks for the suggestion @shalabhc ! 

`exec`ing ag is already better than what I've been doing. I think I'll put together a `folder-search` command though. 

I wonder if I should make it a separately distributed bundle or perhaps put in `app_commands.moon` and PR it? I feel like it would be useful to a lot of Howl users and should be part of the editor.

What do you think?

shalabhc:
I think this should be a built-in command, similar to the project search commands. Please reuse the underlying code for those commands. I'm happy to review a PR.

dgaw:
Thanks for the offer to review @shalabhc! I've just added a [PR](https://github.com/howl-editor/howl/pull/559) for this.

dgaw:
Done in #559

PR #559, opened 2020-03-22
Folder search

dgaw:
An attempt to provide the functionality discussed in #558.

### DIFF
--- a/lib/howl/commands/app_commands.moon
+++ b/lib/howl/commands/app_commands.moon
@@ -647,10 +647,7 @@ do_search = (directory, search, whole_word) ->
     return matches
 
   matches = file_search.sort matches, directory, search, app.editor.current_context
-
-  project = Project.for_file directory
-  conf = if project then project.config else config.for_file(directory)
-  display_as = conf.file_search_hit_display
+  display_as = config.for_file(directory).file_search_hit_display
 
   status = "Loaded 0 out of #{#matches} locations.."
   cancel = false
@@ -670,34 +667,29 @@ do_search = (directory, search, whole_word) ->
 
   locations, searcher
 
-ask_for_search_directory = (search_term) ->
+ask_for_search_directory = (search_query) ->
   buffer = app.editor.buffer
   file = buffer.file or buffer.directory
   project = Project.for_file(file) if file
+
+  opts =
+    title: if search_query
+      "(Please specify a directory to search for '#{search_query}'): "
+    else
+      "(Please specify a directory to search): "
+    prompt: "Directory: "
 
   start_dir = if project
     project.root
   else
     get_buffer_dir(buffer)
 
-  if not start_dir
-    start_dir = howl.io.File(howl.sys.env.HOME)
+  if start_dir
+    opts.path = start_dir.path
 
-  title = if search_term
-    "(Please specify a directory to search for '#{search_term}'): "
-  else
-    "(Please specify a directory to search): "
+  interact.select_directory opts
 
-  directory = interact.select_directory
-    title: title
-    prompt: "Directory: "
-    path: start_dir.path
-
-  directory
-
--- Common input handler for search commands displaying results
--- in a pop-up panel. E.g. project-file-search and folder-search
-search_input = (opts, get_directory) ->
+get_search_query = (opts) ->
   editor = app.editor
 
   search = nil
@@ -708,7 +700,7 @@ search_input = (opts, get_directory) ->
 
   unless search or app.window.command_panel.is_active
     if editor.selection.empty
-      search = app.editor.current_context.word.text
+      search = editor.current_context.word.text
       whole_word = true unless search.is_empty
     else
       search = editor.selection.text
@@ -716,11 +708,18 @@ search_input = (opts, get_directory) ->
   if not search or search.is_empty
     search = interact.read_text prompt: opts.prompt
 
+  search, whole_word
+
+-- Common input handler for search commands displaying results
+-- in a pop-up panel. E.g. project-file-search and folder-search
+search_input = (opts) ->
+  search, whole_word = get_search_query opts
+
   if not search or search.is_empty
     log.warn "No search query specified"
     return
 
-  directory = get_directory search
+  directory = opts.directory or ask_for_search_directory search
   if not directory
     log.warn "No directory selected"
     return
@@ -734,30 +733,14 @@ search_input = (opts, get_directory) ->
 
 -- Common input handler for search commands displaying results
 -- in a buffer. E.g. project-file-search-list and folder-search-list
-search_input_list = (opts, get_directory) ->
-  editor = app.editor
-
-  search = nil
-  whole_word = false
-
-  if opts.text and not opts.text.is_empty
-    search = opts.text
-
-  unless app.window.command_panel.is_active
-    if editor.selection.empty
-      search = app.editor.current_context.word.text
-      whole_word = true unless search.is_empty
-    else
-      search = editor.selection.text
-
-  if not search or search.is_empty
-    search = interact.read_text prompt: opts.prompt
+search_input_list = (opts) ->
+  search, whole_word = get_search_query opts
 
   if not search or search.is_empty
     log.warn "No search query specified"
     return
 
-  directory = get_directory search
+  directory = opts.directory or ask_for_search_directory search
   if not directory
     log.warn "No directory selected"
     return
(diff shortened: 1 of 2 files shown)


lib/howl/commands/app_commands.moon
```diff
@@ -639,16 +639,19 @@ file_search_hit_to_location = (match, search, display_as) ->
...
   unless #matches > 0
     log.error "No matches found for '#{search}'"
     return matches
 
-  matches = file_search.sort matches, project.root, search, app.editor.current_context
-  display_as = project.config.file_search_hit_display
+  matches = file_search.sort matches, directory, search, app.editor.current_context
+
+  project = Project.for_file directory
+  conf = if project then project.config else config.for_file(directory)
```
shalabhc: You don't need to determine the project config. The `config.for_file(directory)` is sufficient and will fallback to the project, or global config automatically.

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +668,112 @@ do_search = (search, whole_word) ->
...
-  name: 'project-file-search',
-  description: 'Searches files in the the current project'
-  input: (opts) ->
-    editor = app.editor
-    search = nil
-    whole_word = false
+ask_for_search_directory = (search_term) ->
+  start_dir = get_buffer_dir(app.editor.buffer)
+  if not start_dir
+    start_dir = howl.io.File(howl.sys.env.HOME)
```
shalabhc: Not required. If `select_directory` is not given a `path:`, it will fall back to the home directory anyway.

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +668,112 @@ do_search = (search, whole_word) ->
...
+  directory = interact.select_directory
+    title: title
+    prompt: "Directory: "
+    path: start_dir.path
+
+  directory
+
+-- Common input handler for search commands displaying results
+-- in a pop-up panel. E.g. project-file-search and folder-search
+search_input = (opts, get_directory) ->
```
shalabhc: Instead of passing a function `get_directory`, can we just pass `opts.directory`? If nil, this can invoke `ask_for_search_directory` but if present, it just gets used? Then the caller from the project commands can just pass the directory instead of a function.

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +668,112 @@ do_search = (search, whole_word) ->
...
-      search = interact.read_text prompt: opts.prompt
+  unless search or app.window.command_panel.is_active
+    if editor.selection.empty
+      search = app.editor.current_context.word.text
+      whole_word = true unless search.is_empty
+    else
+      search = editor.selection.text
+
+  if not search or search.is_empty
+    search = interact.read_text prompt: opts.prompt
```
shalabhc: Some of this code is duplicated between this and the `search_input_list` function. Can we merge or refactor these?

dgaw:
Thanks for reviewing @shalabhc! I've implemented all your requested changes in  
391112c.

dgaw:
ask_for_search_directory does always ask for directory. The project is only used to determine the start directory for the interaction. That's because I want the folder-search commands to open their directory selector at the root of the project (if present).

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +665,107 @@ do_search = (search, whole_word) ->
...
+  if not directory
+    log.warn "No directory selected"
+    return
+
+  locations, searcher = do_search directory, search, whole_word
+
+  if #locations > 0
+    interact.select_location
+      title: "#{#locations} matches for '#{search}' in #{directory.short_path} (using #{searcher.name} searcher)"
+      items: locations
```
shalabhc: Not sure how the original code behaves, but would be nice to have an `else` case here that just shows "No matches" or something like that.

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +665,107 @@ do_search = (search, whole_word) ->
...
+  locations, searcher
 
-command.register
-  name: 'project-file-search',
-  description: 'Searches files in the the current project'
-  input: (opts) ->
-    editor = app.editor
-    search = nil
-    whole_word = false
+ask_for_search_directory = (search_query) ->
```
shalabhc: Can we rename to `get_search_directory` just for consistency with `get_search_query`?

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +665,107 @@ do_search = (search, whole_word) ->
...
+  file = buffer.file or buffer.directory
+  project = Project.for_file(file) if file
 
-    if opts.text and not opts.text.is_empty
-      search = opts.text
+  opts =
+    title: if search_query
+      "(Please specify a directory to search for '#{search_query}'): "
+    else
+      "(Please specify a directory to search): "
```
shalabhc: Looks like we ensure `search_query` exists and is non-empty before calling this function - so this `else` is unnecessary?

lib/howl/commands/app_commands.moon
```diff
@@ -665,39 +665,107 @@ do_search = (search, whole_word) ->
...
+    prompt: "Directory: "
 
-    unless search or app.window.command_panel.is_active
-      if editor.selection.empty
-        search = app.editor.current_context.word.text
-        whole_word = true unless search.is_empty
-      else
-        search = editor.selection.text
+  start_dir = if project
+    project.root
```
shalabhc: Follow up on top level [comment](https://github.com/howl-editor/howl/pull/559#issuecomment-603656120):

I think we should always use get_buffer_dir() as the start_dir for any search. Project dir searches already have their own commands. Yes, they wont let you pick the directory, but when picking a dir for folder-search, I'd prefer to start with the current dir and go up the tree if necessary, which is easy by just pressing `backspace`. WDYT?

lib/howl/commands/app_commands.moon
dgaw: I haven't changed anything here compared to the original version. do_search takes care of the "No matches found" message.

lib/howl/commands/app_commands.moon
dgaw: Done

lib/howl/commands/app_commands.moon
dgaw: Done

lib/howl/commands/app_commands.moon
dgaw: This is the behaviour that I'd started with but soon noticed that most of the time I navigated back to the project root anyway. I usually search 2-3 folders in the project root.

I've added a configuration variable to control this: folder_search_start_dir, which defaults to the directory of the current buffer.

shalabhc:
Thanks, I'm going to play around a bit with this before merging.

dgaw:
Sure. Thanks!

dgaw:
@shalabhc I've been using the `folder-search` command for a while and realised that the UI may be a bit confusing for new users.

When you run the command you are greeted by the following prompt:
`:folder-search`

It's not clear whether the prompt asks for a search string or the name of the folder to search.

I see two possible solutions:
1. Easy - change the title of the prompt to read "Search string:" or something to that effect. 
2. Display the folder selector before asking for a search string.

In my opinion solution 1 is trivial and should be done to begin with. 
Solution 2 would be nice to have, since selecting a folder first feels more natural to me.

What do you think?

shalabhc:
I think solution 2 is better. The default folder is hopefully the desireable one most often, so just pressing 'enter' again is easy in most cases. If we do 1 now and later switch to 2, it could confuse people who have built muscle memory.

shalabhc:
I also think there should always be a title visible. E.g. it should say 'Folder search in:' initially when you're selecting the folder and later say 'Folder search in: /selected/folder'. Almost all commands I can think of provide a title bar. What do you think?

dgaw:
I agree. I'll implement solution 2 including your title suggestion and update this PR.

dgaw:
@shalabhc I've implemented solution 2. I used slightly different wording for the title and prompt. 

Can you please have a look?

shalabhc:
OK this works great. Thank you for your patience!

One last thing - do you really need `folder_search_start_dir`? If folks wanted to search at the project level, they could just use the  project-search commands?

dgaw:
No problem! I've had plenty of time to test this feature myself. :)

Re `folder_search_start_dir`. I thought I'd need it but it turned out I can live without it. I'll remove it.

dgaw:
OK, I've removed `folder_search_start_dir`. I hope it's ok to merge now.

shalabhc:
Merged. Thanks @dgaw!

PR #564, opened 2020-04-28
doc: fix one typo and one incorrectly docummented default binding

Wolf480pl:
Just a couple mistakes I noticed while reading the docs.


nilnor:
Thanks!

PR #571, opened 2020-06-06
Allow for using Lua files in bundles

ghost:


TmpFinger3:
For context, here is the bundle that made me propose these changes - https://repo.or.cz/howl-parkour.git

I had a whole bunch of already written Lua code, and wrote an `init.moon` to make a Howl bundle out of it.  
In order to `require` the Lua files, I had to resort to changing `package.path` from withing the bundle, and to hardcoding the name of the plugin directory.

Issue #371, opened 2017-07-20
Weird scrollbar behaviour


mrshankly:
When using the scroll wheel on my mouse the text scrolls as expected, but the scrollbar doesn't move. If instead I use the arrow keys to navigate everything works as expected.

Clicking and dragging on the scrollbar doesn't work well either, the text moves but the scrollbar doesn't.

I've made a recording of the behaviour: [http://i.imgur.com/jzZmASA.gifv](http://i.imgur.com/jzZmASA.gifv)

Using `git bisect` the first bad commit is 9e9a1209a373aa2794c45bc3fc5421d78500abd3.

### System
I tested on two machines with the same results:
* Linux 4.11.11 on machine 1 and Linux 4.11.9 on machine 2
* GTK+ 3.22.16 on both machines
* The code was compiled with GCC 6.3.0 on machine 1 and GCC 7.1.1 on machine 2

If you need more information or want me to test something else, please do say so. Thanks!

kirbyfan64:
I can actually reproduce this under Kubuntu 17.04, though it seems to have worked on previous versions.

kirbyfan64:
Also, it seems to break under Windows, too.

nilnor:
Thanks a lot for the bisect! In general, Gtk scrollbars are generally a hassle right now, with some issues that are outside of our control (e.g. see [this issue](https://bugzilla.gnome.org/show_bug.cgi?id=765410)). The change pointed at by the bisect was to work around some other scrollbar issue, but perhaps there's some other kind of work-around that can be done. I'm in vacation mode and won't be able to look at it before a few weeks though.

nilnor:
I've revisited the Gtk scrollbar issue and have now added a workaround for the issue. I'll let it simmer in master for a while before closing this issue though to see that it doesn't cause any other problems.

PR #583, opened 2020-09-11
Reinstate the scrollbar sync workaround

### LINKED ISSUE #567: Scrollbar freezes after opening a popup panel
The position of the scrollbar stops being updated after opening a popup panel (e.g. `alt+x` or `ctrl+f`).

The text itself scrolls correctly but the scrollbar is frozen. After closing the popup panel the scrollbar remains broken until I click on the editor area.

[This screen capture](https://vimeo.com/424580649) demonstrates the issue.

dgaw:
@nilnor This PR brings back your scrollbar sync workaround.

It fixes #567 and generally seems to make the scrollbars behave more predictably.

It was originally introduced in e83e98ee5e9e90286636bb753c85ddd887ba268c
as a fix for issue #371.

Then it was removed in f7c84b6d3220bb13f136154a68d67e70415a2348

### DIFF
--- a/lib/aullar/view.moon
+++ b/lib/aullar/view.moon
@@ -4,6 +4,7 @@
 ffi = require 'ffi'
 bit = require 'bit'
 ffi_cast = ffi.cast
+callbacks = require 'ljglibs.callbacks'
 
 Gdk = require 'ljglibs.gdk'
 Gtk = require 'ljglibs.gtk'
@@ -21,6 +22,8 @@ config = require 'aullar.config'
 {:define_class} = require 'aullar.util'
 {:parse_key_event} = require 'ljglibs.util'
 {:max, :min, :floor} = math
+cast_arg = callbacks.cast_arg
+C = ffi.C
 append = table.insert
 
 jit.off true, true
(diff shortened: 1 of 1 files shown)


nilnor:
Thanks for digging this out, and sorry for the long silence.

This workaround was specific to Gtk+ 3 — it deferred `_sync_scrollbars` out of the `size_allocate` handler via a low-priority idle to work around https://bugzilla.gnome.org/show_bug.cgi?id=765410. Master has since been ported to Gtk 4, and that whole code path in `lib/aullar/view.moon` was rewritten: the widget hierarchy, the size-allocate handling and the widget lifecycle (`View#release` replaced `destroy`) all changed, and the Gtk 3 bug being worked around no longer applies. The patch no longer applies and reinstating it as-is would be re-adding a workaround for a bug that isn't there.

Closing for that reason. #567 is a separate matter — the Gtk-4 port fixed a number of rendering/scrolling issues but also introduced new ones, so if you still see misbehaving scrollbars on current master, please do report it there (or in a new issue) against the Gtk-4 code and we'll look at it fresh.

PR #610, opened 2023-03-29
Little update on README.md

DriNeo:
I was stucked on the compilation because my recently installed Archlinux didn't include pkg-config.


nilnor:
thanks!

PR #613, opened 2024-04-26
Added `rename` and `delete` commands

zesterer:
It's very common for me to accidentally create a file with the wrong name, be it due to a typo or just changing my mind about what the file should be called or where it should be. Previously, this meant:

1) Closing the file in Howl so that I don't accidentally end up re-saving it as its old name
2) Switching to my terminal / file explorer
3) Moving + renaming the file
4) Switching back to Howl, opening it via its new path

With the addition of this `rename` command, it's as simple as just typing out the new name/path I wish to give it, and Howl handles the rest.

For completeness, I've also added a `delete` command that deletes the underlying file and then disassociates the buffer with that file.

### DIFF
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -248,6 +248,15 @@ class Buffer extends PropertyObject
     @_associate_with_file file
     @save!
 
+  rename: (file) =>
+    old_file = @file
+    @_associate_with_file file
+    @save!
+    if @file != old_file
+      status, err = pcall old_file\delete
+      if not status
+          log.warn "Failed to delete old alias #{old_file} for #{@file}: #{err}"
+
   as_one_undo: (f) => @_buffer\as_one_undo f
 
   undo: =>
--- a/lib/howl/buffer.moon
+++ b/lib/howl/buffer.moon
@@ -248,7 +248,7 @@ class Buffer extends PropertyObject
     @_associate_with_file file
     @save!
 
-  rename: (file) =>
+  rename_file: (file) =>
     old_file = @file
     @_associate_with_file file
     @save!
@@ -256,6 +256,13 @@ class Buffer extends PropertyObject
       status, err = pcall old_file\delete
       if not status
           log.warn "Failed to delete old alias #{old_file} for #{@file}: #{err}"
+
+  delete_file: (file) =>
+    if @file
+      status, err = pcall @file\delete
+      if not status
+          log.warn "Failed to delete #{@file}: #{err}"
+    @_associate_with_file nil
 
   as_one_undo: (f) => @_buffer\as_one_undo f
 
--- a/lib/howl/commands/file_commands.moon
+++ b/lib/howl/commands/file_commands.moon
@@ -173,10 +173,31 @@ command.register
       command.run 'save-as'
       return
 
-    buffer\rename file
+    buffer\rename_file file
     buffer.mode = mode.for_file file
     log.info ("%s: %d lines, %d bytes written")\format buffer.file.basename,
       #buffer.lines, #buffer
+
+command.register
+  name: 'delete',
+  description: 'Delete the file associated with the current buffer'
+  handler: ->
+    buffer = app.editor.buffer
+    if not buffer.file
+      log.info "Not deleting; buffer not associated with a file"
+      return
+
+    if buffer.modified_on_disk
+      overwrite = interact.yes_or_no
+        prompt: "Buffer '#{buffer}' has changed on disk, delete anyway? "
+        default: false
+      unless overwrite
+        log.info "Not deleting; buffer not saved"
+        return
+
+    old_name = buffer.file.basename
+    buffer\delete_file!
+    log.info ("%s has been deleted")\format old_name
 
 command.register
   name: 'buffer-close',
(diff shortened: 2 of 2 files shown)


zesterer:
Argh, crumbs, I've accidentally included a tonne of my personal changes here. Let me revise the history.

zesterer:
As a point of interest, what's the maintenance status of Howl? I gather that there's been very little activity for a while. I'm interested in doing some low-level contribution. At least for the time being, it's probably going to remain my editor of choice.

nilnor:
Hi! Maintenance status: I've been swamped with other things and hasn't much time to spend on Howl. I still use it myself for daily coding and all things, and as things have started to deteriorate with gtk-3 I'm "now" working on a Gtk-4 port. That's what I use now and will use going forward - it's in the [gtk-4 branch](https://github.com/howl-editor/howl/tree/gtk-4) but things are still a bit rough there with only the steinom theme ported.

As for the PR:  

I'd skip the added methods for `buffer` and instead do it solely in the commands. That already has some checks for modification, etc., which require user interaction. Better to keep ` buffer` as simple as possible, and use the available `file` property as is and keep the rest outside.

DriNeo:
> Hi! Maintenance status: I've been swamped with other things and hasn't much time to spend on Howl.

Hi, why not a new release ? There are some new nice things since 2019.

nilnor:
Thanks for this, and apologies for the long silence — master sat untouched for years while work happened on the `next` branch, which has now been merged. I'm going back through the open PRs.

> [!NOTE]
> **The review below was produced by AI, run and reviewed by me.** The mergeability, lint and compile checks were actually executed against current master; the judgement calls are mine to defend, so push back on anything that looks wrong.

Good news first: **this one survived the merge intact.** Master has since had a backwards-incompatible command/interaction refactor and a full Gtk-3 → Gtk-4 port, which invalidated most older PRs, but neither touched your insertion points. `Buffer#save_as` and the `save-as` command are unchanged, and command handlers taking a single argument is exactly what the refactor requires. I checked:

- `git apply --check` — clean
- `bin/howl-lint lib/howl/buffer.moon lib/howl/commands/file_commands.moon` — clean
- `bin/howl-moonc lib/howl/buffer.moon` — compiles

Style-wise it reads like the code around it, which I appreciate: `pcall old_file\delete` + `log.warn` mirrors the backup-deletion idiom already in `save`, and `rename` follows `save-as` closely.

There are a few things I'd want addressed before merging.

### Blocking

**1. `delete` never actually confirms the deletion.**

It only prompts when `buffer.modified_on_disk`, and that prompt is a copy-paste from `save`:

```moon
prompt: "Buffer '#{buffer}' has changed on disk, delete anyway? "
...
log.info "Not deleting; buffer not saved"
```

For the common case — an unmodified file — it deletes with no confirmation at all. That's irreversible, and it's inconsistent with `save-as`, which confirms before merely *overwriting*. Deleting should always ask, and the "not saved" wording needs fixing regardless.

**2. Command names are unqualified.**

Everything else in `file_commands.moon` is namespaced by the object it acts on: `save-as`, `buffer-close`, `project-open`, `vc-diff-file`. A bare `delete` in the command palette, sitting next to the editing commands, reads like "delete text" rather than "delete the file on disk". I'd prefer `file-rename` / `file-delete`, or `buffer-rename` / `buffer-delete` if you'd rather emphasise the buffer side.

**3. `rename`'s no-file path is incoherent.**

The `input:` function prompts for and returns a file, and then the handler does:

```moon
if not buffer.file
  command.run 'save-as'
  return
```

which discards the file the user just picked and prompts them for a second one. Either guard in `input:` and bail early, or drop the `save-as` delegation and let `rename_file` handle an unassociated buffer.

### Specs

There aren't any, and here there's a clear precedent: `spec/buffer_spec.moon` has a `describe 'save_as(file)'` block covering file-absent, file-present, and the association afterwards. `rename_file` and `delete_file` are public `Buffer` methods of the same shape and want a matching block.

One case worth pinning down explicitly: renaming a file to its own path. That's currently safe because `File.__eq` compares `tostring` (`lib/howl/io/file.moon:345`), so `@file != old_file` is false and nothing gets deleted — but that's load-bearing behaviour sitting on an `__eq` implementation detail, and it should have a test so it stays true.

The commands themselves need no tests; there's no `spec/commands/` directory in the codebase at all.

### Docs

Same precedent as the specs: `save_as` is documented in `site/source/doc/api/buffer.html.md`, and the `save-as` command in `site/source/doc/manual/files.html.md`. Two new public methods and two new commands should show up in both.

### Nits

- The `log.warn` lines under `if not status` are indented 10 spaces rather than 8, in both methods.
- `delete_file: (file) =>` takes a `file` parameter it never uses.

---

One caveat on expectations: the README now states that master is unreleased and narrows contribution channels, with no promises about review or merge timelines. I don't want to imply a schedule I can't keep. But the feature itself is one I'd like to have, and the implementation is close — if you're still interested in picking it up, the blocking points above are what stand between it and a merge. If you'd rather not, say so and I'll close it; the analysis here is enough for someone else to carry it forward.

nilnor:
Since it has been quite a while, I'm closing this. Please re-open if it's still of interest.

zesterer:
No problem. I've since ended up writing [my own editor](https://git.jsbarretto.com/zesterer/zte) that integrates a lot of ideas from Howl, so no longer mind too much about this issue. Good to see Howl development is active again.

nilnor:
Hey, nice! I can always respect a fellow editor builder, best of luck with it :+1:

PR #617, opened 2026-03-10
feat(bundles/howl-themes/catppuccin): init

ninetailedtori:
I added the Catppuccin colourways to the Howl Editor!

License is MIT, but OK with most licenses that preserve credit. Copyright owned by author [Toria (@ninetailedtori)] and [Catppuccin Org.](https://catppuccin.com/).


ninetailedtori:
I need some assistance though, to finish this!

- the unfocused cursor, which comes up when you leave the window, or technically also the cursor used when searching, how do we theme this?

- the text decorations, how do we theme those? they're rather jarringly not using the correct colour for that, and it's rather annoying. This is for all error, info and warning, and actually ANY element that has text components.

Here's an image to explain the questions above, file used to test the editor in use is dracula.moon:
image

ninetailedtori:
image

Also: the borders of popup are rounded, whereas the box itself is non-rounded. I can't seem to assign a type to this to fix it, is this an upstream issue or otherwise?

nilnor:
Thanks for this, and apologies for the very long silence.

Master has moved a lot since this was opened. The most relevant change here is the Gtk-4 port, which landed on master together with a switch to **100% CSS themes**. `howl.ui.theme` no longer accepts MoonScript theme definitions at all — all bundled themes were converted (`bundles/howl-themes/*/*.css`) and `bundles/howl-themes/init.moon` now registers `.css` files. The four `.moon` files in this PR can't be loaded by master.

So I'm closing this as superseded by the theme system rewrite rather than as rejected — the colourways themselves are still welcome. If you're still interested, a fresh PR adding `catppuccin_{latte,frappe,macchiato,mocha}.css` (modelled on e.g. `bundles/howl-themes/dracula/dracula.css`) would apply cleanly, and the `catppuccin.tera` generator could be retargeted at CSS output.

Note that master is unreleased and Gtk-4 based; the last release (`0.6`) and anything before `0ca4ffa2` is still Gtk-3 / MoonScript themes.

